PR #53, opened 2018-03-27
Define an enum for Action to clean up constants in the code

mikrostew:
From a pairing session with @dherman today, just cleaning up the code and making things nicer.

### DIFF
--- a/crates/notion-core/src/style.rs
+++ b/crates/notion-core/src/style.rs
@@ -1,6 +1,6 @@
 //! The view layer of Notion, with utilities for styling command-line output.
 
-use std::fmt::Display;
+use std::fmt::{self, Display, Formatter};
 
 use console::style;
 use indicatif::{ProgressBar, ProgressStyle};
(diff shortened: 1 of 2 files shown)


crates/notion-core/src/style.rs
```diff
@@ -31,13 +31,31 @@ pub fn display_unknown_error() {
...
+impl Display for Action {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), fmt::Error> {
+        let s = match self {
+            &Action::Installing => "Installing"
+        };
+        f.write_str(s)
+    }
+}
+
 /// Constructs a command-line progress bar with the specified "action" string
```
dherman: We should update this comment (I can do it in a follow-up).

crates/notion-core/src/style.rs
```diff
@@ -47,7 +65,8 @@ pub fn progress_bar(action: &str, details: &str, len: u64) -> ProgressBar {
 
     let bar = ProgressBar::new(len);
 
-    bar.set_message(&format!("{: >12} {}", style(action).green().bold(), details));
+    // Action::MAX_WIDTH = 12
```
dherman: I think there is a way to pass in an argument for the width after all, something like:

```rust
bar.set_message(&format!("{: >width$} {}", style(action.to_string()).green().bold(), details, width = Action::MAX_WIDTH));
```

I'll do this in a followup commit.

dherman: (I found this info at:

https://www.reddit.com/r/rust/comments/32wdnj/dynamically_create_a_format_string/

and

https://doc.rust-lang.org/std/fmt/

)

dherman:
Thanks @mikrostew !

PR #54, opened 2018-03-29
Follow-up with cleanup after #53

mikrostew:
Clean up the comments and progress bar message format based on comments for https://github.com/notion-cli/notion/pull/53.

Compiled and tested install locally.


dherman:
awesome!

PR #55, opened 2018-03-29
Progress bar width is already calculated using `term_size` (closes #35)

### LINKED ISSUE #35: progress bar should be sized based on terminal size
`style::progress_bar` currently has a hard-coded size of 40, but should be sized dynamically based on the terminal size (which can be computed via the `term_size` crate).

mikrostew:


dherman:
lol

PR #57, opened 2018-03-30
Add spinner to indicate fetching the public node registry

### LINKED ISSUE #33: visible indication while fetching the public node listing
I believe fetching the public node listing is causing a pause with no visible indication to the user. Test this hypothesis, and if it's true, we should:
- show an activity and/or progress indicator while fetching it
- possibly cache it

I'm not sure if the caching semantics works, since it could mean missing new versions. So we should think that through.

mikrostew:
This adds a progress spinner to give an indication of what is happening when we are fetching the public node registry. Otherwise there is a visible pause with no feedback about what is going on.

This closes #33

### DIFF
--- a/crates/notion-core/src/catalog.rs
+++ b/crates/notion-core/src/catalog.rs
@@ -20,6 +20,7 @@ use installer::Installed;
 use installer::node::Installer;
 use serial;
 use config::{Config, NodeConfig};
+use style::progress_spinner;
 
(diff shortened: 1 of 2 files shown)


dherman:
Nice! At some point we will probably want to unify all our animations and status messages for consistent style but we can't generalize until we have more instances, so for now I think we can just land small changes like this and then iterate later.

PR #62, opened 2018-04-26
much cleaner trait and type structure for the node-archive crate

dherman:
kills the janky `define_source_trait!` macro and instead makes the top-level API the trait


mikrostew:
Nice! Looks like this partially resolves https://github.com/notion-cli/notion/issues/32?

dherman:
@mikrostew I think it was already implemented and the issue is just obsolete? But I will read the code more carefully and report back.

dherman:
Yeah it turns out both the tarball and zip versions have logic for copying the contents into the cache file when they fetch remotely (tarball streams the data with a tee reader, zip fetches synchronously to a file so it just copies). So that issue is bogus.

PR #63, opened 2018-04-27
Simplify setup - automatically figure out artifact paths

mikrostew:
Make the build and install process less painful by automatically figuring out the paths to the build artifacts.

### DIFF
--- a/support/unix/build.sh
+++ b/support/unix/build.sh
@@ -1,18 +1,23 @@
 #!/usr/bin/env bash
 
+script_dir="$(dirname "$0")"
+
 usage() {
-  cat <<'END_USAGE'
+  cat <<END_USAGE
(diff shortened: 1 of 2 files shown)


support/unix/build.sh
```diff
@@ -4,19 +4,16 @@ usage() {
...
 
 The output file is saved as ./install.sh.
 END_USAGE
 }
 
-if [ "$#" -ne 4 ]; then
-  usage >&2
-  exit 1
+target_dir='release'
+if [ -n "$1" ]; then
```
dherman: Can you also add a test that if `$#` is not 0 or 1, and if `$1` is neither `release` nor `debug` nor empty, that it should call `usage()` and `exit 1`?

mikrostew: Good call I'll fix that

PR #77, opened 2018-05-17
Initial events implementation

mikrostew:
This implements the events described in https://github.com/notion-cli/rfcs/pull/16.

For now this includes a bash script that echos the data that it receives (and prints it nicely if you have [jq](https://stedolan.github.io/jq/) installed).

In the normal case, start and end events are produced:

```
$ ./target/debug/notion current
local: v6.11.1 (active)
```

```javascript
[
  {
    "timestamp": 1526597696022,
    "name": "notion",
    "event": "start"
  },
  {
    "timestamp": 1526597696082,
    "name": "current",
    "event": "start"
  },
  {
    "timestamp": 1526597696083,
    "name": "current",
    "event": "end"
  },
  {
    "timestamp": 1526597696083,
    "name": "notion",
    "event": "end",
    "exit_code": 0
  }
]
```

For errors some of the environment is included:

```javascript
[
  {
    "timestamp": 1526594862711,
    "name": "node",
    "event": "start"
  },
  {
    "timestamp": 1526595015053,
    "name": "tool",
    "event": "error",
    "exit_code": 13,
    "error": "No Node version found for = 12.11.3",
    "env": {
      "argv": "./target/debug/node --version",
      "exec_path": "/Users/mistewar/fpe/notion/notion/./target/debug/node",
      "path": "/Users/mistewar/.gem/ruby/2.3.0/bin:/Users/mistewar/.rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/bin:/Users/mistewar/.rubies/ruby-2.3.0/bin:/usr/local/sbin:/Users/mistewar/.yarn/bin:/Users/mistewar/dotfiles/scripts:/Users/mistewar/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/linkedin/bin:/Library/TeX/texbin:/opt/X11/bin:/export/content/linkedin/bin",
      "platform": "Mac OS",
      "platform_version": "10.13.4"
    }
  }
]
```

There are still a few small TODOs left, and some operations that should generate events (specifically downloading and installing node). But I wanted to get this up and start getting feedback while I knock down the rest of the things.


crates/notion-core/src/session.rs
```diff
@@ -92,4 +126,30 @@ impl Session {
...
+        self.event_log.add_event_error(activity_kind, error)
+    }
+
+    // send the events from this session to the monitor
+    pub fn send_events(&mut self) {
+        let command = self.events_command();
+        self.event_log.send_events(command)
+    }
+
+    pub fn events_command(&self) -> Option<String> {
```
dherman: I prefer to avoid `match` expressions for types like Option. I like `.map`/`.and_then` chaining, but it's a little subtle with `.as_ref()`:
```rust
self.project
    .as_ref()
    .and_then(|project| project.manifest().events_plugin.as_ref())
    .map(|plugin| plugin.to_string())
```
Or you can do two nested `if let`s:
```rust
if let Some(ref project) = self.project {
    if let Some(ref p) = project.manifest().events_plugin {
        return Some(p.to_string());
    }
}
None
```

dherman:
Awesome PR! :D Just a few suggested changes here and there, but one change that's a little more fundamental to the structure of the event types.

mikrostew:
Updated this to remove the optional fields in `Event`, which simplifies things a little. Now events look like so:

```
$ ./target/debug/node --version
error: No Node version found for = 16.11.1

[
  {
    "timestamp": 1528144998359,
    "name": "node",
    "event": "start"
  },
  {
    "timestamp": 1528144998660,
    "name": "tool",
    "event": {
      "error": {
        "exit_code": 100,
        "error": "No Node version found for = 16.11.1",
        "env": {
          "argv": "./target/debug/node --version",
          "exec_path": "/Users/mistewar/fpe/notion/notion/./target/debug/node",
          "path": "/usr/local/sbin:/Users/mistewar/.yarn/bin:/Users/mistewar/.cargo/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin",
          "platform": "Mac OS",
          "platform_version": "10.13.4"
        }
      }
    }
  }
]
```

and

```
$ ./target/debug/notion current
local: v6.11.1 (active)

[
  {
    "timestamp": 1528145053600,
    "name": "notion",
    "event": "start"
  },
  {
    "timestamp": 1528145053689,
    "name": "current",
    "event": "start"
  },
  {
    "timestamp": 1528145053690,
    "name": "current",
    "event": {
      "end": {
        "exit_code": 0
      }
    }
  },
  {
    "timestamp": 1528145053690,
    "name": "notion",
    "event": {
      "end": {
        "exit_code": 0
      }
    }
  }
]
```

crates/notion-core/src/tool.rs
```diff
@@ -33,19 +41,30 @@ pub trait Tool: Sized {
...
     /// Delegates the current process to this tool.
-    fn exec(self) -> ! {
+    fn exec(self, mut session: Session) -> ! {
         let mut command = self.command();
-        let status = command.status();
+        let status = command.status().unknown();
         match status {
             Ok(status) if status.success() => {
+                session.add_event_end(ActivityKind::Tool, Some(0));
+                session.send_events();
```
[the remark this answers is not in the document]
mikrostew: done

long-runner
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
mikrostew: I think so, to keep things clean - and it's a unix-only script at the moment

crates/notion-core/src/event.rs
```diff
@@ -0,0 +1,162 @@
...
+use session::ActivityKind;
+
+// the Event data that is serialized to JSON and sent the plugin
+#[derive(Serialize)]
+pub struct Event {
+    timestamp: u64,
+    name: String,
+    event: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    exit_code: Option<i32>,
```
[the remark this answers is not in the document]
mikrostew: Played around with this and got it to work!

crates/notion-core/src/event.rs
```diff
@@ -0,0 +1,162 @@
...
+        };
+        f.write_str(s)
+    }
+}
+
+impl EventKind {
+    pub fn into_event(
+        self,
+        activity_kind: ActivityKind,
+        exit_code: Option<i32>,
```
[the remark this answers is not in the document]
mikrostew: done

crates/notion-core/src/event.rs
```diff
@@ -0,0 +1,162 @@
...
+            Some(exit_code),
+            Some(error),
+        )
+    }
+
+    fn add_event(
+        &mut self,
+        event_kind: EventKind,
+        activity_kind: ActivityKind,
+        exit_code: Option<i32>,
```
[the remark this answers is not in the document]
mikrostew: done

src/notion.rs
```diff
@@ -164,21 +165,24 @@ See 'notion help <command>' for more information on a specific command.
...
+            CommandName::Help => Help::go(self, session),
+            CommandName::Version => Version::go(self, session),
         }
     }
 }
 
 /// The entry point for the `notion` CLI.
 pub fn main() {
-    let exit_code = match Notion::go() {
+    let mut session = Session::new().unwrap();
```
dherman: You don't want to `unwrap()` here, since it will turn the error into a panic, which will crash with an unhelpful Rust error message. Since you can't even generate events without a `Session`, you probably just need a different `match` for the `Session::new()` and the `Notion::go(...)` lines. But you can pull out the common error-handling logic of the two of them into a helper:

```rust
fn handle_errors<F: NotionFail>(err: &F) {
    if err.is_user_friendly() {
        display_error(&err);
    } else {
        display_unknown_error(&err);
    }

    if let Some(ref usage) = err.usage() {
        eprintln!();
        eprintln!("{}", usage);
    }
}

pub fn main() {
    let mut session = match Session::new() {
        Ok(session) => session,
        Err(err) => {
            handle_errors(&err);
            exit(1);
        }
    };

    session.add_event_start(ActivityKind::Notion);

    let exit_code = match Notion::go(&mut session) {
        ...
        Err(err) => {
            handle_errors(&err);
            err.exit_code()
        }
    };
    ...
}

File: crates/notion-core/src/monitor.rs [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+    /// send event to the monitor process
+    // if plugin command is not configured, this is a no-op
+    pub fn send_events(&mut self, events: &Vec<Event>) -> () {
+        if let Some(ref mut child_process) = self.monitor_process {
+            let p_stdin = child_process.stdin.as_mut().unwrap();
+
+            let json = serde_json::to_string(&events);
+            match json {
+                Ok(data) => {
+                    write!(p_stdin, "{}", data).expect("Writing data to plugin failed!");
```
dherman: I think we will want to tighten this `.expect()` up a bit but I've just added it to my [list of warnings](https://paper.dropbox.com/doc/Notion-Warning-Scenarios-vkSds1q3W6LT3C2qVW22M) and we can come back to it later. So don't worry about it for now.

crates/notion-core/src/monitor.rs
```diff
@@ -0,0 +1,84 @@
...
+                }
+                Err(error) => {
+                    println!("There was a problem serializing the JSON data: {:?}", error);
```
dherman: We'll want to tighten this up a bit too, but for now do `eprintln!` instead of `println!`.

crates/notion-core/src/event.rs
```diff
@@ -0,0 +1,131 @@
...
+    argv: String,
+    exec_path: String,
+    path: String,
+    platform: String,
+    platform_version: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "lowercase")]
+enum EventKind {
```
dherman: Beautiful!

crates/notion-core/src/serial/manifest.rs
```diff
@@ -26,6 +26,7 @@ pub struct Manifest {
 pub struct NotionManifest {
     pub node: String,
     pub yarn: Option<String>,
+    pub events_plugin: Option<String>,
```
dherman: This shouldn't live in the manifest, but in the Notion config file. We can fix this up in a follow-up PR though.

crates/notion-core/src/event.rs
mikrostew: thanks! 😄

src/notion.rs
mikrostew: There's a similar issue in the Tool code that I cleaned up like this as well

crates/notion-core/src/monitor.rs
mikrostew: ok, I added a note about this

crates/notion-core/src/monitor.rs
mikrostew: done!

crates/notion-core/src/serial/manifest.rs
mikrostew: ok, added a note about this too

PR #78, opened 2018-05-22
script to run rustfmt with a fixed toolchain version

dherman:
…since rustfmt is currently not stable and changes style policies from one Rust version to another


support/rustfmt.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+IFS='%'
+install_rust_output=$(rustup install ${TOOLCHAIN_VERSION} 2>&1)
+if [ "$?" -ne 0 ]; then
+  echo $install_rust_output 1>&2
+  exit 1
+fi
+
+IFS='%'
```
mikrostew: You only need to set `IFS='%'` once, but this is not bad for being explicit

PR #79, opened 2018-05-31
Cache Node index.json response

### LINKED ISSUE #58: index of all node versions should be cached.
We should respect the cache headers provided by index.json provided by node.

```sh
curl -I https://nodejs.org/download/release/index.json

last-modified: Thu, 05 Apr 2018 05:37:38 GMT
etag: "5ac5b622-1bf31"
access-control-allow-origin: *
expires: Fri, 13 Apr 2018 05:44:24 GMT <--- I think we care about this one.
cache-control: public, max-age=14400
```

In general, they are about a week. But basically always expire before the next Sunday.

Important Scenarios:

1. When the file has expired, and `https://nodejs.org/download/release/index.json` is accessible, we should download the new artifact, and replace it with the exist.
2. When the file hasn't expired, use the cached version.
3. to support concurrent accesses of this file, we should be sure to update the cached file atomically.

benblank:
Caches https://nodejs.org/dist/index.json when fetching it and uses the cached copy until the time specified by the accompanying `Expires` header has passed.

This resolves #58.


crates/notion-core/src/catalog.rs
```diff
@@ -159,16 +200,46 @@ impl NodeCatalog {
...
+                }
+
+                cached.persist(path::node_index_file()?).unknown()?;
+
+                let expiry: NamedTempFile = NamedTempFile::new().unknown()?;
+
+                // Block to borrow expiry for expiry_file.
+                {
+                    let mut expiry_file: &File = expiry.as_file();
+                    let header: &reqwest::header::Raw = response.headers().get_raw("Expires").unwrap();
```
benblank: @dherman — You briefly mentioned not liking the use of `.unwrap()` here, but we never got a chance to discuss alternatives.

benblank:
I also renamed `read_cached` to `read_cached_opt`, as that seemed more consistent.

For cache expiry, I added both support for the Cache-Control header's max-age property as well as a final fallback of four hours, which is what the max-age has been set to in all the responses I've seen from node.org.

Issue #73, opened 2018-05-02
yarn support


dherman:
This is a meta-issue for tracking all the work required to have yarn support.

- [x] #61 download matching version of yarn
- [x] #67 yarn shim

**Edit:** Removing the nightly functionality for the first milestone.

**Edit:** Also removing the `latest` functionality for the first milestone.

Issue #67, opened 2018-05-02
yarn shim


dherman:
Implement the yarn shim executable.

Issue #61, opened 2018-04-25
download matching version of yarn


mikrostew:
Notion should download the version of yarn that matches the semver specified in config.

Specific versions of yarn can be downloaded from  `https://yarnpkg.com/downloads/MAJOR.MINOR.PATCH/yarn-vMAJOR.MINOR.PATCH.tar.gz`

I haven't been able to find anything on https://yarnpkg.com/ that gives info on yarn release versions. The closest thing is a link to https://github.com/yarnpkg/yarn/releases.

PR #80, opened 2018-06-07
Adding Yarn support to notion

monshi:
Enable yarn support as a new tool that follows the same model as node.
Implements yarn version `catalog`, yarn `installer` and `shim`.

Since yarnpkg.com doesn't provide an API for released versions and some limitation with GitHub API to fetch versions and download, we decided to use `notion-cli/yarn-releases` git repo till we figure out a better solution.
Released versions: `https://github.com/notion-cli/yarn-releases/raw/master/index.json`
Download tarball URL: `https://github.com/notion-cli/yarn-releases/raw/master/dist/yarn-v{$version}.tar.gz`

This PR is for the following issues in MVP milestone: #73 #61 #67

### DIFF
--- a/crates/notion-core/src/catalog.rs
+++ b/crates/notion-core/src/catalog.rs
@@ -29,7 +29,7 @@ use serial;
 use serial::touch;
 use style::progress_spinner;
 
-// TODO: Should we move these constants to a config file or even maybe a user config?
+// ISSUE (#86): Move public repository URLs to config file
 /// URL of the index of available Node versions on the public Node server.
 const PUBLIC_NODE_VERSION_INDEX: &'static str = "https://nodejs.org/dist/index.json";
 /// URL of the index of available Yarn versions on the public git repository.
@@ -59,7 +59,7 @@ impl LazyCatalog {
     }
 }
 
-pub struct Collection<I:Install> {
+pub struct Collection<I: Install> {
     /// The currently activated Node version, if any.
     pub activated: Option<Version>,
 
@@ -147,7 +147,7 @@ impl Catalog {
         Ok(())
     }
 
-    // TODO: refactor activate, install and uninstall methods
+    // ISSUE (#87) Abstract Catalog's activate, install and uninstall methods
     // And potentially share code between node and yarn
     /// Activates a Yarn version matching the specified semantic versioning requirements.
     pub fn activate_yarn(&mut self, matching: &VersionReq, config: &Config) -> Fallible<()> {
@@ -228,14 +228,14 @@ impl NotionFail for NoYarnVersionFoundError {
     }
 }
 
-impl<I:Install> Collection<I> {
+impl<I: Install> Collection<I> {
     /// Tests whether this Collection contains the specified Tool version.
     pub fn contains(&self, version: &Version) -> bool {
         self.versions.contains(version)
     }
 }
 
-pub trait Resolve<I:Install> {
+pub trait Resolve<I: Install> {
     /// Resolves the specified semantic versioning requirements from a remote distributor.
     fn resolve_remote(&self, matching: &VersionReq, config: Option<&ToolConfig<I>>) -> Fallible<I> {
         match config {
--- a/crates/notion-core/src/catalog.rs
+++ b/crates/notion-core/src/catalog.rs
@@ -8,6 +8,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::string::ToString;
 use std::time::{Duration, SystemTime};
+use std::marker::PhantomData;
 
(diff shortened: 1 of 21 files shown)


crates/notion-core/src/installer/mod.rs
```diff
@@ -27,3 +31,22 @@ impl Installed {
         }
     }
 }
+
+/// TODO: figure out how to get rid of `Sized`
+pub trait Install:Sized {
```
dherman: Space after the colon

crates/notion-core/src/installer/mod.rs
```diff
@@ -27,3 +31,22 @@ impl Installed {
         }
     }
 }
+
+/// TODO: figure out how to get rid of `Sized`
```
dherman: Just delete this comment; I'll see if there's a way to get rid of `Sized` but it's not a big deal.

crates/notion-core/src/plugin.rs
```diff
@@ -29,14 +29,14 @@ pub struct InvalidCommandError {
     command: String,
 }
 
-impl Resolve {
-    /// Performs resolution of a Node version based on the given semantic
+impl ResolvePlugin {
+    /// Performs resolution of a Tool version based on the given semantic
     /// versioning requirements.
-    pub fn resolve(&self, _matching: &VersionReq) -> Fallible<Installer> {
+    pub fn resolve<I:Install>(&self, _matching: &VersionReq) -> Fallible<I> {
```
dherman: Space after the colon

crates/notion-core/src/catalog.rs
```diff
@@ -136,6 +146,56 @@ impl Catalog {
 
         Ok(())
     }
+
+    // TODO: refactor activate, install and uninstall methods
```
dherman: This would be good, agreed -- but in the interest of merging this PR, why don't you file it as an issue and then turn this into an `// ISSUE(#nn): etc` comment.

crates/notion-core/src/catalog.rs
```diff
@@ -153,74 +213,49 @@ impl NotionFail for NoNodeVersionFoundError {
...
         }
     }
 
-    /// Resolves the specified semantic versioning requirements from the public distributor (`https://nodejs.org`).
-    fn resolve_public(&self, matching: &VersionReq) -> Fallible<Installer> {
+    /// Resolves the specified semantic versioning requirements from the public distributor (e.g. `https://nodejs.org`).
+    fn resolve_public(&self, matching: &VersionReq) -> Fallible<I>;
+
+    /// Resolves the specified semantic versioning requirements from the local catalog.
+    fn resolve_local(&self, req: &VersionReq) -> Option<Version>;
```
dherman: Isn't the behavior of `resolve_local` identical for all implementations? If so, you should be able to hoist the definition here and avoid the duplication between the `YarnCollection` and `NodeCollection` implementations.

crates/notion-core/src/catalog.rs
```diff
@@ -17,18 +18,22 @@ use serde_json;
...
+use installer::node::NodeInstaller;
+use installer::yarn::YarnInstaller;
 use notion_fail::{Fallible, NotionError, NotionFail, ResultExt};
 use path::{self, user_catalog_file};
 use semver::{Version, VersionReq};
 use serial;
 use serial::touch;
 use style::progress_spinner;
 
+// TODO: Should we move these constants to a config file or even maybe a user config?
```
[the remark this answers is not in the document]
monshi: Done https://github.com/notion-cli/notion/issues/86

crates/notion-core/src/catalog.rs
```diff
@@ -54,18 +59,23 @@ impl LazyCatalog {
...
 }
 
-/// The catalog of tool versions available locally.
-pub struct Catalog {
-    pub node: NodeCatalog,
-}
-
-/// The catalog of Node versions available locally.
-pub struct NodeCatalog {
+pub struct Collection<I:Install> {
```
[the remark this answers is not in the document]
monshi: Thanks 👍  Fixed all instances

crates/notion-core/src/catalog.rs
monshi: Done https://github.com/notion-cli/notion/issues/87

crates/notion-core/src/catalog.rs
monshi: Good point, and makes total sense, but as we spoke offline I couldn't simply move the method to the `Resole` trait since `resolve_local` has reference to `self.versions` which is a field on the `Collection` struct. I'd love to get your help and learn how can I abstract that.

PR #81, opened 2018-06-14
Make 'release' the default build target

evansolomon:
Currently running the build script errors when `release` is not set, even though it claims to be the default. This sets `target_dir` to `release` when it is not specified.


dherman:
Seems good, thanks! :D

PR #88, opened 2018-06-29
Move `Resolve::resolve_local` to an inherent method of `Collection`.

dherman:


dherman:
> Do you think we can move other resolve methods to Collection as well?

I don't think so; I think everything else depends on the `I` type variable.

PR #97, opened 2018-07-19
`notion deactivate`

### LINKED ISSUE #72: `notion deactivate`
While the draft [system node RFC](https://github.com/notion-cli/rfcs/pull/15) tried to move us away from recommending installing Notion over-top of a system Node, @caridy was rightly concerned this is too much friction for adoption, especially for casual/curious adopters.

At least for experimentation, we should have workflows for switching back to system Node and getting Notion out of the way.

**Update:** Now that we've worked out the `notion use` strategy (see PR #92), we can implement `notion deactivate` basically the same way.

### LINKED ISSUE #93: abstract the shell backend for generating postscripts
`notion use` needs to generate a tiny postscript in the right shell language for the current console. So instead of hardcoding to bash-style `export NOTION_NODE_VERSION={}` syntax, there should be an abstracted backend for generating this postscript in every supported shell.

dherman:
Implements `notion deactivate`.

Closes #72.
Closes #93.


crates/notion-core/src/shell/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+mod bash;
+
+pub(crate) use self::bash::Bash;
+
+pub enum Postscript {
+    Path(String),
+    ToolVersion { tool: String, version: Version }
+}
+
+/// Thrown when the shell name specified in the Notion environment is not supported.
```
mikrostew: Copy/paste from UnrecognizedShellError?

Maybe "Thrown when Postscript file is not found"?

crates/notion-core/src/shell/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+    fn save_postscript(&self, postscript: &Postscript) -> Fallible<()> {
+        let mut file = File::create(self.postscript_path()).unknown()?;
+        file.write_all(self.compile_postscript(postscript).as_bytes()).unknown()?;
+        Ok(())
+    }
+}
+
+pub struct CurrentShell(Box<dyn Shell>);
+
+/// Thrown when the shell name specified in the Notion environment is not supported.
```
mikrostew: Same - maybe "Thrown when the shell name could not be detected"?

crates/notion-core/src/shell/mod.rs
dherman: Good catch!

PR #100, opened 2018-07-20
Implement `notion shim` command

### LINKED ISSUE #96: `notion shim`
Command to explicitly manage the set of shims.

mikrostew:
Implements `notion shim`

### List shims

Running `notion shim` lists the shims that exist in `~/.notion/bin`:

```sh
$ notion shim
yarn
eslint
node
npx
npm
```

Running `notion shim --verbose` lists those shims and where they will resolve to:

```sh
# in a project where Notion is configured
$ notion shim -v
yarn -> /Users/mistewar/.notion/versions/yarn/1.2.0/bin/yarn
eslint -> /Users/mistewar/src/notion/crates/notion-core/fixtures/basic/node_modules/.bin/eslint
node -> /Users/mistewar/.notion/versions/node/6.11.1/bin/node
npx -> [shim not implemented!]
npm -> /Users/mistewar/.notion/versions/node/6.11.1/bin/npm

# when Notion is not configured
$ notion shim -v
yarn -> [system]
eslint -> [system]
node -> [system]
npx -> [shim not implemented!]
npm -> [system]

# when node or yarn has not been installed locally
$ notion shim -v
yarn -> [will install version = 1.3]
eslint -> /Users/mistewar/src/notion/crates/notion-core/fixtures/basic/node_modules/.bin/eslint
node -> [will install version = 6.11.3]
npx -> [shim not implemented!]
npm -> [will install version = 6.11.3]
```

### Create shims

Running `notion shim <name>` creates a 3rd-party executable shim in the notion shim dir (`~/.notion/bin/` on unix), pointing to `.notion/launchbin`:

```sh
$ notion shim foo
$ ls -og ~/.notion/bin/
total 22504
lrwxr-xr-x  1        33 Jul 23 13:11 foo -> /Users/mistewar/.notion/launchbin
lrwxr-xr-x  1        33 Jul 20 14:20 eslint -> /Users/mistewar/.notion/launchbin
-rwxr-xr-x  1   5946580 Jul 20 11:57 node
lrwxr-xr-x  1        36 Jul 20 11:57 npm -> /Users/mistewar/.notion/launchscript
lrwxr-xr-x  1        36 Jul 20 11:57 npx -> /Users/mistewar/.notion/launchscript
-rwxr-xr-x  1   5572172 Jul 20 11:57 yarn

# creating a shim that already exists shows an error
$ notion shim foo
error: shim `foo` already exists
```

### Delete shims

Running `notion shim <name> --delete` deletes a 3rd-party executable shim:

```sh
$ ls -go ~/.notion/bin/
total 22504
lrwxr-xr-x  1        33 Jul 20 14:20 eslint -> /Users/mistewar/.notion/launchbin
lrwxr-xr-x  1        33 Jul 23 15:12 foo -> /Users/mistewar/.notion/launchbin
-rwxr-xr-x  1   5946580 Jul 20 11:57 node
lrwxr-xr-x  1        36 Jul 20 11:57 npm -> /Users/mistewar/.notion/launchscript
lrwxr-xr-x  1        36 Jul 20 11:57 npx -> /Users/mistewar/.notion/launchscript
-rwxr-xr-x  1   5572172 Jul 20 11:57 yarn

$ notion shim foo --delete
$ ls -go ~/.notion/bin/
total 22504
lrwxr-xr-x  1        33 Jul 20 14:20 eslint -> /Users/mistewar/.notion/launchbin
-rwxr-xr-x  1   5946580 Jul 20 11:57 node
lrwxr-xr-x  1        36 Jul 20 11:57 npm -> /Users/mistewar/.notion/launchscript
lrwxr-xr-x  1        36 Jul 20 11:57 npx -> /Users/mistewar/.notion/launchscript
-rwxr-xr-x  1   5572172 Jul 20 11:57 yarn

# errors if it doesn't exist, or you try to delete the non-3rd-party shims
$ notion shim foo --delete
error: shim `foo` does not exist
$ notion shim npm --delete
error: cannot delete `npm`, not a 3rd-party executable
```

Closes #96.


src/command/shim.rs
```diff
@@ -0,0 +1,225 @@
...
+            Shim::List(verbose) => list(session, verbose),
+            Shim::Create(shim_name, verbose) => create(session, shim_name, verbose),
+            Shim::Delete(shim_name, verbose) => delete(session, shim_name, verbose),
+        };
+        session.add_event_end(ActivityKind::Shim, 0);
+        result
+    }
+}
+
+fn list(session: &mut Session, verbose: bool) -> Fallible<bool> {
```
dherman: This function is hard to read. Can you break out some of the nested loops into helper functions?

mikrostew: Done!

src/command/shim.rs
```diff
@@ -100,38 +108,41 @@ Options:
     }
 }
 
-fn list(session: &mut Session, verbose: bool) -> Fallible<bool> {
+fn list(session: &Session, verbose: bool) -> Fallible<bool> {
```
dherman: This is an improvement! But there are still a few things we can do to make it simpler. First a few general points:

- You should generally avoid inspecting error cases, and just propagate them.
- You should generally prefer `let foo = bar()?;` to `bar().map(|foo|, { ... })`. This avoids rightward drift, avoids confusing iteration `map` versus result `map`, and lets you code closer to a style that's like exceptions. As an analogy in JS, this is why `let foo = await bar();` is nicer than `bar.then((foo) => { ... })`.
- It's easiest to just convert foreign error types like `io::Error` into notion's error types as early as possible, and the `?` syntax makes that lightweight. So instead of returning `Result<Foo, io::Result>` from helper functions, just return `Fallible<Foo>`.

As for the `Command` trait, don't throw away errors and return `Ok(false)` -- you want to let the particular error propagate out.

Here's how we could write this function to be simpler and more readable/maintainable:

```rust
fn list(session: &Session, verbose: bool) -> Fallible<()> {
    let shim_dir = path:shim_dir()?;
    let files = fs::read_dir(shim_dir).unknown()?;

    for file in files {
        print_file_info(f, session, verbose)?;
    }

    Ok(())
}
```

src/command/shim.rs
```diff
@@ -100,38 +108,41 @@ Options:
...
+            .all(|ref result| result.as_ref().ok() == Some(&true))
         })
 }
 
+fn print_file_info(
+    file: fs::DirEntry,
+    session: &Session,
+    verbose: bool,
+) -> Result<bool, io::Error> {
+    file.path().file_name().map_or(Ok(false), |shim_name| {
```
dherman: A `DirEntry` always has a file name, so instead of using the `Option` that comes from `file.path().file_name()` you can use `file.file_name()` which directly returns an `OsString`.

src/command/shim.rs
```diff
@@ -100,38 +108,41 @@ Options:
...
+        if verbose {
+            match resolve_shim(session, &shim_name) {
+                Ok(shim_info) => println!("{} -> {}", shim_name.to_string_lossy(), shim_info),
+                Err(err) => {
+                    style::display_error(style::ErrorContext::Notion, &err);
```
dherman: I don't think we want to be doing this error handling logic here, as opposed to letting it propagate out and be handled at the top level.

src/command/shim.rs
```diff
@@ -100,38 +108,41 @@ Options:
...
+                file.and_then(|f| print_file_info(f, session, verbose))
+            })
+            .collect::<Vec<_>>()
+            .iter()
+            // return false if anything failed
+            .all(|ref result| result.as_ref().ok() == Some(&true))
         })
 }
 
+fn print_file_info(
```
dherman: Here's a simplified version of this function:

```rust
fn print_file_info(file: fs::DirEntry, session: &Session, verbose: bool) -> Fallible<()> {
    let shim_name = file.file_name();

    if verbose {
        let shim_info = resolve_shim(session, &shim_name)?;
        println!("{} -> {}", shim_name.to_string_lossy(), shim_info);
    } else {
        println!("{}", shim_name.to_string_lossy());
    }

    Ok(())
}
```

src/command/shim.rs
```diff
@@ -152,13 +163,18 @@ fn resolve_shim(session: &mut Session, shim_name: &OsStr) -> Fallible<ShimKind>
     }
 }
 
+fn node_is_available(project: &Project, session: &Session) -> Fallible<Option<Version>> {
```
dherman: Since this isn't a boolean predicate, maybe a better name would be something like `available_node_version`?

mikrostew: Yeah I agree that's a better name

src/command/shim.rs
mikrostew: Doh! I missed that in the docs somehow

src/command/shim.rs
mikrostew: That is so much simpler!

I made peace with the borrow checker, but lately I have been fighting Result and Option. This helps a lot!

src/command/shim.rs
mikrostew: fixed

src/command/shim.rs
mikrostew: So much better!

PR #105, opened 2018-08-01
Better messages for common errors

### LINKED ISSUE #70: error messages for known errors
For MVP, we should do a first pass to cover the most common known errors and ensure they are classified as known errors with user-friendly error messages.

Let's start a checklist and we can add to it as we discover more:

- [x] `NoNodeVersionFoundError`
- [ ] `VersionParseError`
- [ ] `UnimplementedError`
- [ ] `DownloadFailedError`
- [ ] `AmbiguousVersionError`
- [ ] `DependenciesMissingError`
- [ ] ...?

mikrostew:
I implemented error messages for common error cases where we usually get `error: an internal error occurred`, or `thread 'main' panicked at 'not yet implemented'`, which are not user-friendly or helpful.

Instead, we get these errors now:

```
$ notion install 6.5.4.3
error: the given version requirement is invalid
```

```
$ notion config list
error: command `list` is not yet implemented
```

```
$ notion install 8
error: Could not fetch public registry
https://nodejs.org/dist/index.json: failed to lookup address information: nodename nor servname provided, or not known
```

```
$ notion install 9
error: Failed to download version 9.11.2
https://nodejs.org/dist/v9.11.2/node-v9.11.2-darwin-x64.tar.gz: failed to lookup address information: nodename nor servname provided, or not known
```

```
$ notion uninstall 6
error: Could not parse version: Expected dot
```

```
$ notion shim -v
error: Could not read dependent package info: Could not read package info: No such file or directory (os error 2)
```

Most of these are just surfacing lower-level errors. If there are better messages we can use for these I am fine with changing them to be more helpful/descriptive.

Closes #70.

### DIFF
--- a/crates/notion-core/src/tool.rs
+++ b/crates/notion-core/src/tool.rs
@@ -50,6 +50,25 @@ impl NotionFail for BinaryExecError {
     }
 }
 
+#[derive(Fail, Debug)]
+#[fail(display = "this tool is not yet implemented")]
+pub(crate) struct ToolUnimplementedError;
+
+impl ToolUnimplementedError {
+    pub(crate) fn new() -> Self {
+        ToolUnimplementedError {}
+    }
+}
+
+impl NotionFail for ToolUnimplementedError {
+    fn is_user_friendly(&self) -> bool {
+        true
+    }
+    fn exit_code(&self) -> i32 {
+        4
+    }
+}
+
 /// Represents a command-line tool that Notion shims delegate to.
 pub trait Tool: Sized {
     fn launch() -> ! {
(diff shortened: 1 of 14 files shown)


crates/notion-core/src/tool.rs
```diff
@@ -50,6 +50,25 @@ impl NotionFail for BinaryExecError {
     }
 }
 
+#[derive(Fail, Debug)]
+#[fail(display = "this tool is not yet implemented")]
+pub(crate) struct ToolUnimplementedError;
+
+impl ToolUnimplementedError {
+    pub(crate) fn new() -> Self {
+        ToolUnimplementedError {}
```
dherman: Don't use the optional `{}` here -- you can just write `ToolUnimplementedError`.

PR #107, opened 2018-08-04
Move events reporting plugin to the user's config file

### LINKED ISSUE #106: Events reporting should be a config plugin
Right now, the events reporting plugin is configured in a project's `package.json`, but it should be in the user's config file.

dherman:
…instead of `package.json`.

Closes #106.

### DIFF
--- a/crates/notion-core/src/event.rs
+++ b/crates/notion-core/src/event.rs
@@ -8,6 +8,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use monitor::LazyMonitor;
 use notion_fail::{Fallible, NotionError};
 use session::ActivityKind;
+use plugin::Publish;
 
 // the Event data that is serialized to JSON and sent the plugin
 #[derive(Serialize)]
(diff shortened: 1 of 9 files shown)


crates/notion-core/src/session.rs
```diff
@@ -182,22 +183,20 @@ impl Session {
...
-    pub fn events_command(&self) -> Option<String> {
-        self.project
-            .as_ref()
-            .and_then(|project| project.manifest().events_plugin.as_ref())
-            .map(|plugin| plugin.to_string())
-    }
-
     pub fn exit(mut self, code: i32) -> ! {
-        self.send_events();
+        match publish_plugin(&self.config) {
```
dherman: I think it might be better to push this up to eagerly read the config file when creating the `Session`, since it's better to get an eager error about an invalid config file. But I'm filing that as a separate issue (#108).

crates/notion-core/src/event.rs
```diff
@@ -121,11 +122,15 @@ impl EventLog {
...
-    // send the events from this session to the monitor
-    pub fn send_events(&mut self, command: Option<String>) {
-        self.monitor
-            .get_mut(command)
-            .unwrap()
-            .send_events(&self.events);
+    pub fn publish(&mut self, plugin: Option<&Publish>) {
+        match plugin {
+            Some(&Publish::Url(_)) => { unimplemented!() }
+            Some(&Publish::Bin(ref command)) => {
```
mikrostew: I like the way this is setup now, but in the future I think it would be nice to support sending events to multiple places, for instance logging to a file as well as sending to a URL, to be more flexible.

dherman: Good point. Also at some point it needs some better data abstraction one way or another; there are too many types that are `Foo::{Url, Bin}`. I filed #113.

PR #110, opened 2018-08-05
Change `notion install` to `notion fetch`

dherman:
Now `notion install 6.11.1` becomes `notion fetch node 6.11.1`. Also supports `notion fetch yarn 1.7`.

This is in preparation for the new model of global toolchain management, which still needs an RFC. But I want to get the basic shape of the CLI closer to that model in preparation for our first level of MVP feedback.

There are a couple cross-cutting refactors here worth noting:

- What was called an `Installer` in the implementation is now called a `Distro` because setting it up is more an internal abstraction. Instead of thinking it as _installation_, the fetch operation should be thought of more as preparing the layout of a distribution. ("Installation" carries an implication of affecting the user-visible state of their system. By contrast, "fetch" is more just about ensuring that network operations have successfully completed; any other aspects of unpacking and preparing a particular version of the tool for usage is an implementation detail.)
- Similarly, lots of `install_XXX` methods become `fetch_XXX`.

Remaining work to do after this, which isn't yet done in the PR:

- Rename the current `notion use` command to `notion install`.
- Create a separate `notion use` command which updates `package.json` (we can use @stefanpenner's [detect-indent-rs](https://github.com/stefanpenner/detect-indent-rs) in conjunction with [serde_json with a custom indentation](https://play.rust-lang.org/?gist=009ef8f29aa6af44c26d32be5f3c9724) to preserve layout when modifying a manifest -- this is what npm and Yarn do, as I understand it)
- Delete the `notion uninstall` command, or do ... something sensible with it. Related, probably rename the `uninstall_XXX` methods to `remove_XXX` or something like that.


crates/notion-core/src/distro/node.rs
```diff
@@ -15,18 +15,18 @@ use semver::Version;
...
-pub struct NodeInstaller {
+/// A provisioned Node distribution.
+pub struct NodeDistro {
     archive: Box<Archive>,
     version: Version,
 }
 
-impl Install for NodeInstaller {
+impl Distro for NodeDistro {
     /// Provision an `Installer` from the public Node distributor (`https://nodejs.org`).
```
monshi: nitpick: `Installer` to `Distro`

PR #111, opened 2018-08-06
Changes to command UX

dherman:
Changes:
- Rename previous `notion use` to `notion install` but with behavior closer to previous `notion default`
- Remove `notion default`
- Remove `notion uninstall`
- Stub new `notion use` for modifying project settings

Still left to do:
- Create a separate `notion use` command which updates `package.json` (we can use @stefanpenner's [detect-indent-rs](https://github.com/stefanpenner/detect-indent-rs) in conjunction with [serde_json with a custom indentation](https://play.rust-lang.org/?gist=009ef8f29aa6af44c26d32be5f3c9724) to preserve layout when modifying a manifest -- this is what npm and Yarn do, as I understand it)


dherman:
The reason for removing `notion default` is that, with reproducible distributions, i.e. statelessness, I think the per-console state is less important than in other version managers. Instead of having to use multiple consoles to work on separate projects, you can move around between projects in a single console and never have to choose different state.

Also, it can be confusing and hard to remember the state in different consoles. So I think it's worth seeing how far we can get with just global installation state and nothing else. I'm leaving the ability to override the settings with environment variables, though -- so that way we can still easily experiment with workflows for temporary overrides.

src/command/install.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    fn run(self, session: &mut Session) -> Fallible<bool> {
+        session.add_event_start(ActivityKind::Install);
+        match self {
+            Install::Help => {
+                Help::Command(CommandName::Install).run(session)?;
+            }
+            Install::Node(requirements) => {
+                session.set_default_node(&requirements)?;
```
mikrostew: So the most recently version of `node` or `yarn` that has been manually installed becomes the default global version? I think that's fine, since we're trying to get away from depending on the global toolchain state. And if the user really needs to care about the global version they can `notion install node <version>` for now. Or maybe we need something like `notion use node <version> --global` for this case?

dherman: I'm not sure I follow -- can you say more?

mikrostew: (pasted in from chat)

ok, I think I'm just hung up on command syntax
so `notion install node x.x.x` will install that node version if it's not already installed
or just set that version to be the default if it's already installed
same for yarn or anything else
thinking about the use cases for this, it doesn't quite make sense to me
as a user, I `notion install node 10`
then some time later I `notion install node 8`
which becomes the default
after that I want to switch back to node 10 for whatever reason
so I would run `notion install node 10` again
even though it's already installed
I think the problem is that I know that node 10 is already installed, so why do I need to install it again
but maybe it comes back to the stateless model of notion
like, who cares what is already installed
running notion commands is declaring how I want my system (or project) to be configured
so running `notion install node 10` is saying "I want to run this system with node v10, I don't care how you do it, just make it happen"
is `install` the right verb for that?
I guess, I can't think of anything better haha

PR #129, opened 2018-08-24
Convert Notion::go and friends to Fallible<()>.

benblank:
Per discussion on [#114](https://github.com/notion-cli/notion/pull/114#discussion_r212484537).

### DIFF
--- a/src/command/current.rs
+++ b/src/command/current.rs
@@ -1,7 +1,7 @@
 use std::string::ToString;
 
 use notion_core::session::{ActivityKind, Session};
-use notion_fail::{ExitCode, Fallible};
+use notion_fail::{ExitCode, Fallible, NotionFail};
 
 use Notion;
 use command::{Command, CommandName, Help};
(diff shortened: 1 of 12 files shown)


src/command/current.rs
```diff
@@ -90,11 +98,14 @@ Options:
...
                 }
 
-                Ok(any)
+                any
             }
         };
         session.add_event_end(ActivityKind::Current, ExitCode::Success);
-        result
+        if !result {
+            throw!(NoVersionsFoundError);
```
mikrostew: It doesn't seem right to me that not having a currently selected node version is an error. But since command syntax is still up in the air, I'm fine with this for now.

dherman: Agreed, and we don't need to resolve this right this second, but just to explain the intuition I had here, it was so that you could use e.g. `notion current > /dev/null` in shell scripting to determine if there's a current Node version. IOW, a CLI "error" is also just a "falsey value" in the shell.

PR #133, opened 2018-08-25
support "latest" to be specified as version for node and yarn

monshi:
notion will resolve to a version that is the latest released version of node/yarn
implemented latest for following notion commands:
`notion fetch node latest`
`notion fetch yarn latest`
`notion install node latest`
`notion install yarn latest`
`notion use node latest`
`notion use yarn latest`

moved node resolve request code out from `resolve_public` to a new method `resolve_node_versions` which can be reused for `latest`.
Yarn has a API to simply fetch the latest version number `https://yarnpkg.com/latest-version`, but couldn't find such endpoint for node, so ended up finding the latest version from `https://nodejs.org/dist/index.json`

### DIFF
--- a/crates/notion-core/src/catalog.rs
+++ b/crates/notion-core/src/catalog.rs
@@ -35,6 +35,8 @@ const PUBLIC_NODE_VERSION_INDEX: &'static str = "https://nodejs.org/dist/index.j
 /// URL of the index of available Yarn versions on the public git repository.
 const PUBLIC_YARN_VERSION_INDEX: &'static str =
     "https://github.com/notion-cli/yarn-releases/raw/master/index.json";
+/// URL of the latest Yarn version on the public yarnpkg.com
+const PUBLIC_YARN_LATEST_VERSION: &'static str = "https://yarnpkg.com/latest-version";
 
 /// Lazily loaded tool catalog.
 pub struct LazyCatalog {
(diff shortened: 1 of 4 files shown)


crates/notion-core/src/catalog.rs
```diff
@@ -278,51 +280,7 @@ impl RegistryFetchError {
...
-                expiry.persist(path::node_index_expiry_file()?).unknown()?;
-
-                let serial: serial::index::Index =
-                    serde_json::de::from_str(&response_text).unknown()?;
-
-                spinner.finish_and_clear();
-                serial
-            }
-        }.into_index()?;
+        let index: Index = resolve_node_versions().unwrap().into_index()?;
```
mikrostew: You should use a question mark here instead of `unwrap()`, so the error from `resolve_node_versions` is propagated correctly

crates/notion-core/src/catalog.rs
```diff
@@ -433,3 +391,75 @@ fn max_age(response: &reqwest::Response) -> u32 {
...
+            spinner.finish_and_clear();
+            Ok(serial)
+        }
+    }
+}
+
+pub fn parse_node_version(src: String) -> Fallible<String> {
+    let mut version:String= src;
+    if version == "latest" {
+        let index = resolve_node_versions().unwrap().into_index()?;
```
mikrostew: Same here, question mark instead of `unwrap()`

crates/notion-core/src/catalog.rs
```diff
@@ -433,3 +391,75 @@ fn max_age(response: &reqwest::Response) -> u32 {
...
+        };
+        version = latest_version.to_string();
+    }
+    Ok(version)
+}
+
+pub fn parse_yarn_version(src: String) -> Fallible<String> {
+    let mut version:String = src;
+    if version == "latest" {
+        let mut response: reqwest::Response = reqwest::get(PUBLIC_YARN_LATEST_VERSION).unwrap();
```
mikrostew: Need to handle this error, like how it's done in `resolve_node_versions`:

```
let mut response: reqwest::Response = reqwest::get(PUBLIC_YARN_LATEST_VERSION)
    .with_context(RegistryFetchError::from_error)?;
```

monshi: thanks! good call

crates/notion-core/src/catalog.rs
monshi: 👍 good to learning, updated the diff

Issue #134, opened 2018-08-27
Use checksums for cached distributions


mikrostew:
We should verify the checksums of the distributions downloaded by Notion, to:
* verify the downloaded distro is correct, and
* verify the cached distro is valid

## Node

Node distributions have a corresponding sha256sum file at https://nodejs.org/dist/vx.y.z/SHASUMS256.txt (see https://github.com/nodejs/node#verifying-binaries).

We should download that checksum, and can verify it using the [sha2 crate](https://docs.rs/sha2/0.7.1/sha2/).

## Yarn

Short-term, we will need to add checksums to https://github.com/notion-cli/yarn-releases, and verify those.

Long-term, we will probably get yarn from github (at https://github.com/yarnpkg/yarn/releases). But they don't include checksums with those releases, or through the API (e.g. https://api.github.com/repos/yarnpkg/yarn/releases/latest) &ndash; not sure what we can use for that.

PR #135, opened 2018-08-28
Testing 1/3: Add a bunch of unit tests

mikrostew:
As the first part of improving the testing for Notion, I have added a bunch of unit tests.

Next up is creating an acceptance testing harness to tackle #7.


crates/notion-core/src/distro/node.rs
```diff
@@ -24,6 +24,7 @@ pub struct NodeDistro {
 
 /// Check if the cached file is valid. It may have been corrupted or interrupted in the middle of
 /// downloading.
+// FIXME(#134) - verify checksum
```
dherman: Thanks for citing an issue! Let's use consistent formatting and write `ISSUE(#134)` instead of `FIXME`.

src/command/shim.rs
```diff
@@ -108,6 +108,7 @@ Options:
     }
 }
 
+// FIXME: all the logic here should be moved to notion-core
```
dherman: One last `FIXME` left :)

crates/notion-core/src/serial/manifest.rs
```diff
@@ -166,3 +181,173 @@ where
         Ok(bin_map)
     }
 }
+
+#[cfg(test)]
+pub mod tests {
```
dherman: I don't have a strong opinion either way, but you could consider splitting these out into a separate file. You'd create a `manifest/` subdirectory, move `manifest.rs` into `manifest/mod.rs`, split out this module into `manifest/test.rs`, and then change this code to

```rust
#[cfg(test)]
pub mod tests;
```

But I don't know if that's better or worse, just a thought.

crates/notion-core/src/distro/node.rs
mikrostew: done

src/command/shim.rs
mikrostew: done

crates/notion-core/src/serial/manifest.rs
mikrostew: Good idea. And since you already refactored manifest.rs to manifest/mod.rs, this is an easy change

PR #138, opened 2018-08-31
Ensure the cache dir exists when caching things

### LINKED ISSUE #136: Notion fails with an unknown error if the cache directory does not exist
The `~/.notion/cache` directory is created by the install script, so this is not usually an issue. But if you blow away that cache dir, Notion fails:

```
$ rm -rf ~/.notion/cache/

$ notion use node 9
error: an internal error occurred

Notion is still a pre-alpha project, so we expect to run into some bugs,
but we'd love to hear about them so we can fix them!

Please feel free to reach out to us at @notionjs on Twitter or file an issue at:

    https://github.com/notion-cli/notion/issues

```

mikrostew:
Closes #136

### DIFF
--- a/crates/notion-core/src/lib.rs
+++ b/crates/notion-core/src/lib.rs
@@ -49,3 +49,39 @@ extern crate notion_fail_derive;
 
 #[macro_use]
 extern crate cfg_if;
+
+use std::fs;
+use std::io;
+use std::path::Path;
+
+use notion_fail::{ExitCode, Fallible, NotionFail, ResultExt};
+
+#[derive(Debug, Fail, NotionFail)]
+#[fail(display = "Could not create directory {}: {}", dir, error)]
+#[notion_fail(code = "FileSystemError")]
+pub(crate) struct CreateDirError {
+    pub(crate) dir: String,
+    pub(crate) error: String,
+}
+
+impl CreateDirError {
+    pub(crate) fn for_dir(dir: String) -> impl FnOnce(&io::Error) -> CreateDirError
+    {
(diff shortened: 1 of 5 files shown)


crates/notion-core/src/lib.rs
```diff
@@ -49,3 +49,39 @@ extern crate notion_fail_derive;
...
+impl CreateDirError {
+    pub(crate) fn for_dir(dir: String) -> impl FnOnce(&io::Error) -> CreateDirError {
+        move |error| CreateDirError {
+            dir,
+            error: error.to_string(),
+        }
+    }
+}
+
+/// If the input path is a directory, it creates that directory. If the input path is a file, it creates the parent directory of that file.
```
dherman: This API doesn't quite make sense as is. The only way `Path::is_dir()` returns true is if the directory already exists in the filesystem and is a directory. (A `Path` isn't inherently a file or directory, it's just a path.) So that means the first branch of the conditional is always recreating a directory that already exists.

But I don't think you need the dynamic behavior anyway (and it makes the API less predictable and also not concurrency-safe). IINM, every use case in this PR is a file that wants its parent directory created. So I'd just make the API a narrower `ensure_containing_dir_exists` function that always assumes it's being given a path to a file. (If `p.parent()` returns `None` you can produce an internal error, since that's a programming error inside Notion.)

dherman: Also just a code organization suggestion: these definitions (`ensure_dir_exists` and `CreateDirError`) might be a little nicer in a helper module called `fs` or `path` or something.

PR #139, opened 2018-09-01
notion-core cleanup

dherman:
General cleanup of `notion-core`:
- Moved serialization logic into crate-internal submodules of each datatype's module
- Moved filesystem utility functions from scattered modules into `notion_core::fs`
- Created `notion_core::version` module with a `VersionSpec` type for parsing CLI version specifiers (e.g. `"latest"`)


dherman:
@mikrostew Sorry this one is a _little_ bit big. Once I'd refactored the serialization modules, I felt I needed to create the top-level `VersionSpec` type so that I could make `notion_core::version::serial` crate-internal just like the other `::serial` modules, and that led to another wave of refactoring.

I've got more refactoring I want to do but I stopped it there so things didn't get out of hand.

PR #144, opened 2018-09-06
Improvements to `latest`

dherman:
A few little fixes:

- Fixes the `latest` calculation which I broke in #139. (@mikrostew's test harnesses would've caught this!).
- Make the version argument to `notion install` optional, and default to `latest`.
- Stores the index as a `Vec` instead of a `BTreeMap` since we don't use any map methods; we're only operating on an iterator.
- A couple method renames to align with current terminology.

### DIFF
--- a/src/command/install.rs
+++ b/src/command/install.rs
@@ -1,12 +1,12 @@
-use notion_core::version::VersionSpec;
 use notion_core::session::{ActivityKind, Session};
+use notion_core::version::VersionSpec;
 use notion_fail::{ExitCode, Fallible};
 
 use result::ResultOptionExt;
 
+use CommandUnimplementedError;
 use Notion;
 use command::{Command, CommandName, Help};
-use CommandUnimplementedError;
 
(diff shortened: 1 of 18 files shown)


src/command/install.rs
```diff
@@ -47,16 +49,21 @@ Supported Tools:
...
             "yarn" => {
-                Ok(Install::Yarn(VersionSpec::parse(&arg_version)?))
+                Ok(Install::Yarn(version))
             },
-            ref tool => Ok(Install::Other {
-                name: tool.to_string(),
-                version: VersionSpec::parse(&arg_version)?,
+            ref package => Ok(Install::Other {
+                package: package.to_string(),
+                version: version,
```
mikrostew: nit: can just do `version,` here (I see frequent compiler warnings about these)

dherman: Yeah I'm confused about when those warnings happen—for some reason I'm not seeing them on my machine.

PR #163, opened 2018-10-12
Project autoshim

benblank:
This is a speculative implementation of [RFC 23](https://github.com/notion-cli/rfcs/pull/23).

A notable change with this PR is the grammar for the `notion shim` command, which has been moved behind the `notion-dev` feature flag.

* `notion shim` => `notion shim list`
* `notion shim foo` => `notion shim create foo`
* `notion shim foo --delete` => `notion shim delete foo`

### DIFF
--- a/crates/notion-core/src/tool.rs
+++ b/crates/notion-core/src/tool.rs
@@ -337,14 +337,12 @@ impl Tool for Yarn {
 
     /// Perform any tasks which must be run after the tool runs but before exiting.
     fn finalize(session: &Session, maybe_status: &io::Result<ExitStatus>) {
-        if let Ok(status) = maybe_status {
-            if status.success() {
-                if let Some(project) = session.project() {
-                    let errors = project.autoshim();
+        if let Ok(_) = maybe_status {
+            if let Some(project) = session.project() {
+                let errors = project.autoshim();
 
-                    for error in errors {
-                        display_error(&error);
-                    }
+                for error in errors {
+                    display_error(&error);
                 }
             }
         }
--- a/crates/notion-core/src/tool.rs
+++ b/crates/notion-core/src/tool.rs
@@ -5,7 +5,7 @@ use std::ffi::{OsStr, OsString};
 use std::io;
 use std::marker::Sized;
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, ExitStatus};
 
 use notion_fail::{ExitCode, FailExt, Fallible, NotionError, NotionFail};
 use path;
@@ -86,10 +86,14 @@ pub trait Tool: Sized {
     /// Extracts the `Command` from this tool.
     fn command(self) -> Command;
 
+    /// Perform any tasks which must be run after the tool runs but before exiting.
+    fn finalize(_session: &Session, _maybe_status: &io::Result<ExitStatus>) {}
+
     /// Delegates the current process to this tool.
     fn exec(self, mut session: Session) -> ! {
         let mut command = self.command();
         let status = command.status();
+        Self::finalize(&session, &status);
         match status {
             Ok(status) if status.success() => {
                 session.add_event_end(ActivityKind::Tool, ExitCode::Success);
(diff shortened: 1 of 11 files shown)


crates/notion-core/src/tool.rs
```diff
@@ -330,4 +334,17 @@ impl Tool for Yarn {
...
     }
+
+    /// Perform any tasks which must be run after the tool runs but before exiting.
+    fn finalize(session: &Session, maybe_status: &io::Result<ExitStatus>) {
+        if let Ok(_) = maybe_status {
+            if let Some(project) = session.project() {
+                let errors = project.autoshim();
+
+                for error in errors {
+                    display_error(&error);
```
mikrostew: Since autoshimming is running essentially behind the scenes, and these errors are not affecting the status, should these be displayed as warnings instead of errors?

src/command/use_.rs
```diff
@@ -85,6 +86,17 @@ Options:
...
+            let errors = project.autoshim();
+
+            for error in errors {
+                if error.is_user_friendly() {
+                    display_error(ErrorContext::Notion, &error);
+                } else {
+                    display_unknown_error(ErrorContext::Notion, &error);
+                }
+            }
+        }
```
mikrostew: Again here, I think these should be displayed to the user as warnings instead of errors, since autoshimming was not  explicitly invoked by the user in this context, and they don't affect the actual command status.

crates/notion-core/src/tool.rs
benblank: I disagree.  Auto-shimming is an important part of Notion's core functionality and I think it needs to be called out as a significant problem when it fails.  Doubly so *because* it won't affect the tool's exit status.

src/command/use_.rs
benblank: Auto-shimming will never be invoked by the user, because it is an implicit action.  I was actually on the fence as to whether or not to have `notion use` fail when errors occur while auto-shimming, but decided to keep it consistent with the behavior of tool shims.

src/command/shim.rs
```diff
@@ -131,15 +188,61 @@ fn print_file_info(file: fs::DirEntry, session: &Session, verbose: bool) -> Fall
...
 
 fn create(_session: &Session, shim_name: String, _verbose: bool) -> Fallible<()> {
-    shim::create(&shim_name)?;
+    let result = shim::create(&shim_name)?;
+
+    if result == shim::ShimResult::AlreadyExists {
+        throw!(ShimAlreadyExistsError {
+            name: shim_name,
+        });
+    }
```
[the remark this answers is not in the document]
benblank: Done.

src/command/shim.rs
```diff
@@ -131,15 +188,61 @@ fn print_file_info(file: fs::DirEntry, session: &Session, verbose: bool) -> Fall
...
 
 fn delete(_session: &Session, shim_name: String, _verbose: bool) -> Fallible<()> {
-    shim::delete(&shim_name)?;
+    let result = shim::delete(&shim_name)?;
+
+    if result == shim::ShimResult::DoesntExist {
+        throw!(ShimDoesntExistError {
+            name: shim_name,
+        });
+    }
```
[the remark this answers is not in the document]
benblank: Done.

benblank:
In the call this morning, Michael raised a corner case I don't think the PR currently covers: running `notion use` before dependencies are installed.

crates/notion-core/src/tool.rs
benblank: This was resolved in last week's call.

src/command/use_.rs
benblank: This was resolved in last week's call.

benblank:
Missing dependencies are no longer reported as errors.

crates/notion-core/src/tool.rs
dherman: I think I'm starting to doubt the part of the exit codes RFC that says that we should preserve the exit code of the underlying tool even if extra logic afterwards fails. We don't need to block this PR on this, but I [filed an RFC issue](https://github.com/notion-cli/rfcs/issues/26) to follow up.

dherman:
I pushed a change to the one nit (`// TODO` ==> `// ISSUE`).

Issue #170, opened 2018-10-19
Separate Inventory and Toolchain types


dherman:
Remove the "catalog" concept and replace it with the concepts of the _inventory_ (i.e., the set of tools that have been fetched locally) and the _toolchain_ (i.e., the set of tools currently actively installed for the user).

dherman:
Implemented in #181.

PR #181, opened 2018-10-21
Separate Inventory and Toolchain types

dherman:
This PR implements #170, which involves a few changes:

## Removed the "catalog"

The "catalog" had a few problems. The terminology obscured the fact that it's on the local machine, and it elided two distinct concepts: the set of tools that have been fetched, and the set of tools that the user has currently installed. These two concepts are now called the _inventory_ and the _user toolchain_, respectively.

This PR creates an `Inventory` and `Toolchain` type for these two concepts. It also changes the directory layout, in a few ways:

  - The `cache/node` directory is moved to `tools/inventory/node`.
  - The `catalog.toml` is removed in favor of:
    - The available Node versions are unpacked at `tools/user/image/node`.
    - The available Yarn versions are unpacked at `tools/user/image/yarn`.
    - The user toolchain's platform image, stored at `tools/user/platform.toml`. 

## Platform images include the npm version

A platform image contains a `NodeVersion`, which is now a pair of versions: the Node runtime version, and the npm version. The directory layout of a Node version -- for now -- is `tools/user/image/node/$runtime/$npm` where `$runtime` is the Node runtime version and `$npm` is the npm version. This will allow users to choose different versions of npm without overwriting shared Node installations (i.e., it obsoletes the `npm i -g npm` workflow).

It's a space-inefficient representation, so we'll want to come up with a way to share the contents of the vast majority of files. Git really is perfect for this...

Another thing this PR does with npm is extract the default npm version when fetching a Node distro, and saving that version in a text file at `tools/inventory/node-$runtime-npm` for easy retrieval. That way when operations only specify a Node version without an npm version, npm can default to that version.

## Get tests working

In order to get the tests working I had to make some more changes:

  - Update mockito to 0.14.0 so we can mock binary files and get Windows working.
  - Eliminated the `method_name` macro which seems to have been a necessary workaround for 64-bit Windows in older mockito versions but was making it fail now.
  - Turned on more detailed error reporting in CI with `NOTION_DEV` and `RUST_BACKTRACE`, which makes it easier to debug failing CI runs.
  - Refactored the `node-archive` crate to `archive`, which can serve _both_ zip and tarball archives, since on Windows we need zip files for Node distros but tarballs for Yarn.

Checklist of what I did step by step:

  - [x] Rename `Catalog` to `Inventory`
  - [x] Create `Toolchain` type
  - [x] `Session` contains one of both
  - [x] Move `set_user_node` to `toolchain.install_node`
  - [x] Move `set_user_yarn` to `toolchain.install_yarn`
  - [x] Rename `~/.notion/cache` to `~/.notion/tools/inventory`
  - [x] Rename `~/.notion/versions` to `~/.notion/tools/image`
  - [x] Remove fetched versions list from `catalog.toml`, just read list from directory
  - [x] Move `catalog.toml` to `~/.notion/tools/user/platform.toml`
  - [x] Eliminate catalog from acceptance test sandbox, use platform.toml instead
  - [x] Comment read_dir_eager
  - [x] Comment inventory::serial::versions_matching
  - [x] Change platform.toml to `{ runtime = '10.9.0', npm = '6.2.0' }`
  - [x] Align representation of node/npm pair between platform.toml and package.json?
  - [x] ~Split up (NodeDistro as Distro)::fetch into smaller methods~ (filed #210 to tackle separately)
  - [x] Update Unix installer to create new directory structure
  - [x] Some manual smoke testing on a local machine


crates/notion-core/src/manifest/serial.rs
```diff
@@ -93,6 +94,8 @@ impl Manifest {
             return Ok(Some(image::Image {
                 node: VersionSpec::parse_version(&toolchain.node)?,
                 node_str: toolchain.node.clone(),
+                npm: VersionSpec::parse_version(&toolchain.npm)?,
```
charlespierce: I'm still coming up to speed on both Rust and the structure of Notion, but does this mean that if `npm` is not present in `toolchain` that loading the manifest will return an error? Which will mean that none of the tools will be able to execute because they won't see an active image for the current directory?

I see there's code later on that when you pin the version of Node, it will also write the NPM version, but if I'm correct in my interpretation above, this means that devs won't be able to just manually edit `package.json` to add `toolchain: { node: "10.0.0" }` or whatever, without also including the exact version of NPM that is bundled with that version of Node. Instead, the only simple way to enable Notion in a project is with the `pin` command.

From an ergonomics standpoint, it seems like it would be nice to continue to allow a minimal `node` configuration so that Notion can be enabled manually instead of essentially requiring the user to run a command. I could also see confusion for a user that is using `yarn`, asking "Why is there an NPM configuration added to my project when I don't use NPM for anything?"

dherman: @charlespierce Great question! There's a slightly uncomfortable mismatch here, in that the toolchain key is meant to fully pin all the information, so it's more like a lockfile, and yet it's stored in `package.json`, which is a human-owned file.

We could fudge a little bit and say that an absent `"npm"` key is either sugar for the default npm version that comes with the node version, or it means `"npm"` is not supported at all. Pro: lighter-weight syntax. Con: more variation means more ways to say the same thing, which also means automated commands for doing the same thing might either create churn in git commits or have to get smarter to preserve style.

But one decision we made earlier that I still think was the right one was disallowing loose semver specifications for the versions -- you can only fully specify an exact pinned version; semver resolution is never triggered by reading the manifest. This is the right call because it's the basis for reproducibility.

It's worth at least asking whether this all calls for a separate notion lockfile. I'm just very hesitant to impose another file on JS projects for such a small amount of data.

So maybe a good middle ground to start with for now (which we can revisit after usage experience) is:

- `"npm"` defaults to the default version that comes with Node, because that's the most common usage pattern
- `"yarn"` defaults to not-present, since it's not Node's default package manager
- when Notion writes to `package.json`, it should stylistically prefer omitting the `"npm"` key if it's the same as the default version, to create a pleasant canonical style and minimize churn in git commits
 
Thoughts?

charlespierce: Those seem like good defaults that maintain the simple path for devs while keeping the necessary specificity that we want inside of Notion.

One thought I had from looking at the way the `notion_home` folder is structured with this PR, is that an unspecified `npm` version in `package.json` could be mapped to a folder named e.g. `default`. That way internally the representation can still be that there is no specific `npm` version, but it will correctly map to a filesystem location.

crates/notion-core/src/inventory/serial.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub(crate) fn load() -> Fallible<Self> {
+        let re = Regex::new(r"(?x)
+            node
+            -
+            v(?P<version>\d+\.\d+\.\d+) # Node version
+            -
+            (?P<os>[a-z]+)              # operating system
+            -
+            (?P<arch>[a-z0-9]+)         # architecture
+            \.(zip|tar\.gz)
```
mikrostew: I was looking at download options to make sure this RE isn't missing anything, and it looks like there are `tar.xz` archives available for OSX and Linux, which are smaller download sizes. Might be an option for us to use in the future - I created #211 to track that.

crates/notion-core/src/path/windows.rs
```diff
@@ -27,43 +27,47 @@ cfg_if! {
...
 
-// C:\
-//     ProgramData\
-//         Notion\
-//             cache\                                  cache_dir
-//                 node\                               node_cache_dir
-//                     node-v4.8.4-win-x64.zip         archive_file("4.8.4")
-//                     node-v6.11.3-win-x64.zip
-//                     node-v8.6.0-win-x64.zip
+// C:\Users\dherman\AppData\Local\
```
mikrostew: Probably won't be in `dherman` folder for most people 😄

crates/notion-core/src/distro/yarn.rs
```diff
@@ -36,13 +36,13 @@ pub struct YarnDistro {
...
 /// downloading.
 // ISSUE(#134) - verify checksum
-fn cache_is_valid(cache_file: &PathBuf) -> bool {
-    if cache_file.is_file() {
-        if let Ok(file) = File::open(cache_file) {
-            match node_archive::load(file) {
+fn distro_is_valid(file: &PathBuf) -> bool {
+    if file.is_file() {
+        if let Ok(file) = File::open(file) {
+            match Tarball::load(file) {
```
charlespierce: Just to double-check, the `node` versions are downloaded as either tarball (for Unix) or Zip (for Windows), while the `yarn` versions are just always tarballs, even on Windows?

crates/notion-core/src/toolchain/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn current() -> Fallible<Toolchain> {
+        let path = user_platform_file()?;
+        let src = touch(&path)?.read_into_string().unknown()?;
+        let serial: serial::Platform = toml::from_str(&src).unknown()?;
+        Ok(Toolchain {
+            platform: serial.into_image()?
+        })
+    }
+
+    pub fn get_installed_node(&self) -> Option<NodeVersion> {
```
charlespierce: It seems like `installed` might not be the right adjective here, since technically all of the versions in the inventory are "installed" in the sense that they are available without a download.

Perhaps something like `active` e.g. `get_active_node`?

crates/notion-core/src/distro/yarn.rs
dherman: That's correct.

crates/notion-core/src/toolchain/mod.rs
dherman: I can see how someone reading the code could find that ambiguous. "Active" sounds pretty good to me!

dherman:
@mikrostew @charlespierce Addressed your comments so far!

tests/acceptance/support/sandbox.rs
```diff
@@ -426,11 +452,9 @@ impl Sandbox {
             .env("HOME", home_dir())
             .env("USERPROFILE", home_dir()) // windows
             .env("NOTION_HOME", notion_home())
-            .env("NOTION_DATA_ROOT", notion_home()) // windows
             .env("PATH", &self.path)
             .env("NOTION_POSTSCRIPT", notion_postscript())
-            .env_remove("NOTION_DEV")
-            .env_remove("NOTION_NODE_VERSION")
+            //.env_remove("NOTION_DEV")
```
mikrostew: Remove this commented-out code

crates/notion-core/src/distro/node.rs
```diff
@@ -96,12 +128,18 @@ impl Distro for NodeDistro {
...
-    fn fetch(self, collection: &NodeCollection) -> Fallible<Fetched> {
+    fn fetch(self, collection: &NodeCollection) -> Fallible<Fetched<NodeVersion>> {
         if collection.contains(&self.version) {
-            return Ok(Fetched::Already(self.version));
+            let filename = path::node_npm_version_file_name(&self.version.to_string());
+            let npm = path::node_inventory_dir()?.join(&filename);
+
+            return Ok(Fetched::Already(NodeVersion {
+                runtime: self.version,
+                npm: read_to_string(npm).unknown()?.parse().unknown()?
```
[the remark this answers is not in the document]
dherman: Great points. I left a few thoughts on #215 and I'll add development info to the README in this PR.

crates/notion-core/src/path/unix.rs
```diff
@@ -44,27 +44,40 @@ cfg_if! {
...
+//                 yarn/                                   yarn_image_root_dir
+//                     1.7.0/                              yarn_image_dir("1.7.0")
+//             user/                                       user_toolchain_dir
+//                 bins/
+//                     ember ~> ../packages/ember-cli
+//                 packages/
+//                     ember-cli/
+//                         package.toml
+//                         contents/
+//                 platform.toml                           user_platform_file
```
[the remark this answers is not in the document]
dherman: Yeah, we can update the issue after we merge this.

dherman:
🚀

PR #185, opened 2018-10-22
Update os_info to released version.

rwjblue:
The changes that we needed (https://github.com/DarkEld3r/os_info/pull/88) have been released in 1.1.0.


mikrostew:
Thanks Rob!

PR #187, opened 2018-10-24
Cleanup formatting and enforce.

rwjblue:
* Adds `./dev/unix/rustfmt.sh --check` to CI runs
* Runs `./dev/unix/rustfmt.sh` across the codebase to reformat as needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: required
+rust: stable
 language: rust
 cache: cargo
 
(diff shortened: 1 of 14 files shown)


rwjblue:
Thanks for reviewing @mikrostew! I updated with some verbiage to the README, and rebased...

dherman:
Sorry we let this one bitrot! Now that [Rust has stable rustfmt built-in](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html#new-tools), I think we should do this slightly differently:

- Delete `rustfmt.sh`
- Change the minimum Rust requirement to 1.31
- Have the CI do `rustup component add rustfmt` and then `cargo fmt --all -q -- --check`
- Update the README's contribution section to state the minimum Rust requirement and the requirement to do `cargo fmt` as above

We probably want to merge a couple of the larger PRs in the queue before doing this so there's not _too_ much rebasing churn.

PR #188, opened 2018-10-31
[WIP] Smoke Tests

### LINKED ISSUE #162: Smoke tests for basic workflows
The acceptance tests are not testing the actual directories used by Notion, because the tests are sandboxed. We should have some tests that use these directories. These should only run in CI, because they will need to clear out the directories used by Notion and reset them to a default state before each test.

mikrostew:
Closes #162


tests/smoke/notion_fetch.rs
```diff
@@ -0,0 +1,31 @@
...
+fn fetch_node() {
+    let s = sandbox()
+        .build();
+
+    assert_that!(
+        s.notion("fetch node 10.4.1"),
+        execs()
+            .with_status(0)
+    );
+    assert_eq!(s.node_version_is_fetched("10.4.1"), true);
```
benblank: Can we get comments/documentation on why the node version is different in each of the smoke tests?  And should the yarn versions be different as well?

.travis.yml
```diff
@@ -18,7 +18,9 @@ jobs:
         # run single-threaded because some tests change environment variables, which is not thread-safe
         cargo test --package notion-core --package notion-fail --package node-archive --package notion-fail-derive --package progress-read -- --test-threads=1 &&
         # the acceptance tests, using network mocks
-        cargo test --features mock-network
+        cargo test --features mock-network &&
```
dherman: Heads up, this'll be a tiny merge conflict since I merged #191.

tests/smoke/support/sandbox.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// files and dirs in the sandbox
+
+fn package_json_file(mut root: PathBuf) -> PathBuf {
+    root.push("package.json");
+    root
+}
+
+pub struct Sandbox {
```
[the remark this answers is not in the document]
dherman: I think it'd be good to call this something different from the other Sandbox. Maybe something like TempProject{Builder}.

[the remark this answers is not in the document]
mikrostew: I like TempProject, just changed it to that.

.travis.yml
mikrostew: Yup, fixed that, and some other merge conflicts from the latest PR merges.

tests/smoke/notion_fetch.rs
mikrostew: I am using different versions of node and yarn for these tests just in case the environment is not cleaned up correctly, to help with test isolation. I'll add some comments about that.

tests/smoke/autodownload.rs
```diff
@@ -0,0 +1,60 @@
...
+    assert_that!(
+        p.node("--version"),
+        execs()
+            .with_status(0)
+            .with_stdout_contains("v10.11.0")
+    );
+}
+
+#[test]
+#[ignore] // ISSUE (#227) - This fails in CI because of the github API rate limit
```
dherman: Ooh I like how you have an issue comment next to each `#[ignore]`d test. That seems like a good policy for us. Maybe also worth commenting on the issue that when the issue is fixed we should un-`#[ignore]` the tests blocked on it.

mikrostew: Just commented in that issue

PR #201, opened 2018-11-27
Add 'notion activate' command

### LINKED ISSUE #194: There should be a `notion activate` command
Since there is `notion deactivate`, there should also be `notion activate` to re-enable Notion in the current shell. This should add Notion shims back to the PATH, and set NOTION_HOME, and undo whatever else was done by `notion deactivate`.

charlespierce:
Fixes #194 

Adds a `notion activate` command that essentially un-does the `notion deactivate` command, restoring the `NOTION_HOME` environment variable and ensuring the shims directory is in the `PATH`.

If run when Notion has not been `deactivate`d, the variables should be unchanged.

Tested:
* Running `notion deactivate` followed by `notion activate` correctly restores the ENV.
* Running `notion activate` without first running `deactivate` doesn't change the ENV.

Inside a project with Notion `toolchain` set to a different version of Node than is globally installed on the system:
* Running `node --version` shows the Notion-installed version.
* Running `notion deactivate` followed by `node --version` shows the globally installed version.
* Running `notion activate` followed by `node --version` again shows the Notion-installed toolchain version.

Note: I took on this issue to make sure I could reason about Rust well enough for more complicated tasks, so please point out anything that isn't idiomatic Rust or that could be done in a cleaner way.


crates/notion-core/src/image/mod.rs
```diff
@@ -66,6 +66,20 @@ impl System {
...
 
+    /// Reproduces the Notion-enabled `PATH` environment variable for situations where
+    /// Notion has been deactivated
+    pub fn enabled_path() -> Fallible<OsString> {
+        let old_path = envoy::path().unwrap_or(envoy::Var::from(""));
+        let shim_dir = path::shim_dir()?;
+
+        let mut parts = old_path.split();
+
+        if !old_path.split().any(|part| part == shim_dir) {
```
mikrostew: No need to split twice, you can simplify this a little by doing:

```rust
if !parts.any(|part| part == shim_dir) {
```

charlespierce: My understanding was that `any()` was one of the methods that consumed the iterator, so if we did `parts.any(...)` then later on the `parts.join()` call would fail / not include all of the items in the iterable.

charlespierce: Confirmed, switching to `if !parts.any(...)` causes the tests to fail, as the `any` consumes part of the iterator, leaving the `parts.join()` to only return part of the `PATH`, instead of everything that we want it to.

dherman: This might be slightly cleaner:

```rust
        if !old_path.split().any(|part| part == shim_dir) {
            old_path.split().prefix(vec![shim_dir]).join().unknown()?
        } else {
            old_path
        }
```

The double split smells to me like the `envoy` API isn't quite ideal, but I'm not exactly sure what the right thing is. For now, splitting twice seems unavoidable.

charlespierce: Updated

PR #205, opened 2018-11-28
Add support for npm and npx

### LINKED ISSUE #112: Support npm
Notion support for `npm` in toolchain same as `yarn`.
- Download the matching version of `npm` based on toolchain config in package.json
- Download the latest version of `npm` according to `https://registry.npmjs.org/npm/latest`
- `npm` shim executable

Open question: should we use `npm` version that is bundled with `node.js` by default? for instance, if I use `node@10.8.0` in my toolchain I'd get `npm@6.2.0` by default, however, I can change that by pinning a different version in toolchain config.

charlespierce:
Info
-----
Fixes #112 

Given the issues that have been seen in beta testing with scripts directly running `npm`, it seems that having the `npm` shim throw an error when `yarn` is enabled would still leave us in the same place. Instead, the `npm` shim will show a warning the user when it is run in a yarn-enabled project, letting them know that they should be running yarn instead of npm.

Changes
-----
* Added tools for `npm` and `npx`, including compiled binaries.
* Both tools take advantage of the fact that `npm` and `npx` are installed along side `node` and so are already in the execution path.
* Included a helpful error message for `npx` when invoked on projects that have `node < 8.2.0`, letting the user know that `npx` is only available in `node >= 8.2.0` and which version their project is using.
* Updated bash install scripts to include and install the `npm` and `npx` binaries, instead of symlinking to `launchscript`.

Tested
-----
* Installed locally and verified that `npm` and `npx` commands work as intended, pulling their version from that bundled with `node`.

Notes
-----
* I was unable to reproduce the reported `pre-commit` / scripts running `npm` errors, as whenever `node` is run by `notion`, the `path` that it's executed under already includes `npm` so it would find it even on existing versions of `notion`.
* I couldn't see where (if any) in the Windows installer the `npm` and `npx` shims are installed / set up, so I didn't modify those install projects at all.


crates/notion-core/src/tool.rs
```diff
@@ -348,3 +355,94 @@ impl Tool for Yarn {
...
+    fn new(session: &mut Session) -> Fallible<Self> {
+        session.add_event_start(ActivityKind::Npx);
+
+        let mut args = args_os();
+        let exe = arg0(&mut args)?;
+        if let Some(ref platform) = session.current_platform()? {
+            session.prepare_image(platform)?;
+
+            // npx was only included with Node >= 8.2.0. If less than that, we should include a helpful error message
+            let required_node = VersionSpec::parse_requirements(">= 8.2.0")?;
```
benblank: While `npx` wasn't *included* until Node 8.2.0, it's *compatible with*, and can be installed under, versions at least as far back as 4.9.1.  I don't think we should be blocking its use here on older versions where it's been installed.

charlespierce: Based on the discussion in the issue #112, the `npm` (and `npx`) versions that are supported right now are just those that come bundled with `node`. There isn't a separate entry in `toolchain` for them, nor any way to install a specific version that isn't bundled with the selected version of Node.

Given that, if the selected `node` version is older than `8.2.0`, then `npx` won't be one of the binaries installed, so it won't be available on the path (unless the user has already installed it globally).

My thinking was it's better to have a nice error message that lets the user know what is going on instead of whatever generic error message is thrown when the command can't be found.

benblank: I wouldn't expect `npx` to be pinned to a package; I'm more thinking of the `notion install npx latest` case.

Which, come to think of it, raises another question — if we have an `npx`-specific shim, should `notion install npx` be using that, or the generic shim?  Perhaps this should be tabled for now, after all.  😄

charlespierce: Yeah, I believe that currently `notion install npx ...` won't work, as we only support installing `node` and `yarn`. This isn't changing that model at all, just supporting the fact that `npm` and `npx` are binaries that are available when you install `node`.

I agree when we have a more robust implementation that allows directly managing the versions of `npm` (and presumably `npx`), then we should remove this block, but for the current approach where we know ahead of time it's going to fail, may as well be nice about it 😄

benblank: I've created #206 to track the question.

crates/notion-core/src/tool.rs
```diff
@@ -348,3 +355,94 @@ impl Tool for Yarn {
...
+        session.add_event_start(ActivityKind::Npm);
+
+        let mut args = args_os();
+        let exe = arg0(&mut args)?;
+        if let Some(ref platform) = session.current_platform()? {
+            session.prepare_image(platform)?;
+
+            if platform.yarn.is_some() {
+                // If Yarn is enabled for the current project, warn the user that they shouldn't be using npm
+                style::display_warning(&String::from("This project is configured to use yarn, npm should be avoided"));
```
dherman: I'm pretty wary of adding warnings to Notion. Warnings are a risk in a tool that gets used regularly, because when users learn to ignore them, they depreciate the value of all console output.

Also, there may be some unintended consequences here. For example, you may be in a Yarn-based project but run a third-party script that does something benign with npm, like use it to query information about the package's version, and then that script will suddenly start generating this warning.

crates/notion-core/src/tool.rs
```diff
@@ -348,3 +355,94 @@ impl Tool for Yarn {
...
+
+    fn from_components(exe: &OsStr, args: ArgsOs, path_var: &OsStr) -> Self {
+        Npm(command_for(exe, args, path_var))
+    }
+
+    fn command(self) -> Command {
+        self.0
+    }
+
+    /// Perform any tasks which must be run after the tool runs but before exiting
```
dherman: I think you don't need to repeat the rustdoc comment here, since it should already be documented in the trait declaration.

crates/notion-core/src/tool.rs
```diff
@@ -348,3 +355,94 @@ impl Tool for Yarn {
...
+
+            if platform.yarn.is_some() {
+                // If Yarn is enabled for the current project, warn the user that they shouldn't be using npm
+                style::display_warning(&String::from("This project is configured to use yarn, npm should be avoided"));
+            }
+            Ok(Self::from_components(&exe, args, &platform.path()?))
+        } else {
+            // Using 'Node' as the tool name since the npm version is derived from the Node version
+            // This way the error message will prompt the user to add 'Node' to their toolchain, instead of 'npm'
+            throw!(NoSuchToolError {
```
dherman: I'm not 100% confident this is the right UX but we don't need to change it for now -- we can ship this and play with it.

dherman:
Oops, I chose the wrong review status but I can't figure out how to change it. You can look at the comments to see which changes I requested.

dherman:
Merging!!

PR #221, opened 2018-12-18
Robust progress

### LINKED ISSUE #214: Robustly compute uncompressed size of tarballs
Currently we make three separate HTTP connections -- sequentially -- to fetch a tarball and compute its uncompressed size. And one of them is a HEAD request, which doesn't seem to work for GitHub releases, and that's caused us to create a Yarn releases proxy repo, which is a pain to maintain, at least at this early stage of Notion development.

I think a more efficient and robust approach would be to immediately fetch the full payload, which if I'm not mistaken can synchronously return header information while it's still fetching the payload. We can then use those headers to determine if it accepts byte range requests and then compute the uncompressed size with a second request.

I did a proof of concept here, and you can see in the tests that it works for the standard Yarn releases:

https://github.com/dherman/headless-tarball-fetch

dherman:
Fixes #214.

This PR streamlines the implementation of progress reporting for fetching and unpacking tarballs: it eliminates the unnecessary extra `HEAD` request and gets both the `Content-Length` and `Accepts` headers from a `GET` request of the full payload. Since [reqwest fetches payloads by-need](https://twitter.com/seanmonstar/status/1073634882469781504), we can kick of a second `GET` request to compute the uncompressed size before fetching the full tarball payload.

This means once we ship this, we no longer need to maintain the [Yarn release proxy repo](https://github.com/notion-cli/yarn-releases/), because the implementation just fetches releases directly from the GitHub releases API of the official public Yarn repo.

### DIFF
--- a/crates/notion-core/src/inventory/serial.rs
+++ b/crates/notion-core/src/inventory/serial.rs
@@ -134,8 +134,7 @@ impl YarnEntry {
         let release_filename = &format!("yarn-{}.tar.gz", self.tag_name)[..];
         self.assets
             .iter()
-            .find(|&&YarnAsset { ref name }| { name == release_filename })
-            .is_some()
+            .any(|&YarnAsset { ref name }| { name == release_filename })
     }
 }
 
--- a/crates/notion-core/src/inventory/serial.rs
+++ b/crates/notion-core/src/inventory/serial.rs
@@ -74,35 +74,86 @@ impl YarnCollection {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
-pub struct Index(Vec<Entry>);
+pub struct NodeIndex(Vec<NodeEntry>);
 
 #[derive(Serialize, Deserialize, Debug)]
-pub struct Entry {
+pub struct NodeEntry {
     pub version: String,
     pub npm: Option<String>,
     pub files: Vec<String>,
 }
 
-impl Index {
-    pub fn into_index(self) -> Fallible<super::Index> {
+fn trim_version(s: &str) -> &str {
+    let s = s.trim();
+    if s.starts_with('v') {
+        s[1..].trim()
+    } else {
+        s
+    }
+}
+
+impl NodeIndex {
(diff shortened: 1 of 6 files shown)


crates/notion-core/src/inventory/serial.rs
```diff
@@ -74,35 +74,86 @@ impl YarnCollection {
...
+}
+
+impl YarnEntry {
+    /// Is this entry a full release, i.e., does this entry's asset list include a
+    /// proper release tarball?
+    fn is_full_release(&self) -> bool {
+        let release_filename = &format!("yarn-{}.tar.gz", self.tag_name)[..];
+        self.assets
+            .iter()
+            .find(|&&YarnAsset { ref name }| { name == release_filename })
```
dherman: Lol, reviewing myself... I should use `Iterator::any()` here instead of `find().is_some()`

crates/notion-core/src/inventory/serial.rs
```diff
@@ -74,35 +74,85 @@ impl YarnCollection {
...
+    let s = s.trim();
+    if s.starts_with('v') {
+        s[1..].trim()
+    } else {
+        s
+    }
+}
+
+impl NodeIndex {
+    pub fn into_index(self) -> Fallible<super::NodeIndex> {
```
mikrostew: And having `NodeIndex::into_index` (and `YarnIndex::into_index`) is also a little confusing, since it is already an index.

crates/notion-core/src/inventory/serial.rs
```diff
@@ -74,35 +74,85 @@ impl YarnCollection {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
-pub struct Index(Vec<Entry>);
+pub struct NodeIndex(Vec<NodeEntry>);
 
 #[derive(Serialize, Deserialize, Debug)]
-pub struct Entry {
+pub struct NodeEntry {
```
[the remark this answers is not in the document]
dherman: That's a good point. That's been the convention for all the serialization types, but I can come up with a more distinct naming convention.

PR #224, opened 2018-12-19
Use `NOTION_HOME` for all sandboxing

### LINKED ISSUE #207: Use NOTION_HOME for test sandbox
The sandboxing logic we use for testing is a bit messier than it could be. I think we can clean it up by eliminating the `NOTION_SANDBOX` env var and not changing `HOME` or `USERPROFILE`, and just use `NOTION_HOME` as the main hook for sandboxing.

Rough plan:

- don't change `HOME` or `USERPROFILE` -- don't need to isolate behavior of tools we're delegating to
- use `NOTION_HOME` as only sandboxing hook
  * unix: replaces `~/.notion`
  * windows: replaces `C:\Users\$username\AppData\Local\Notion`
- test sandbox should just set `NOTION_HOME` to `$unique/.notion` on all platforms
- `path::mod` should define notion_home
  * if `NOTION_HOME` is set, use that
  * otherwise use `default_notion_home()`
- `path::$os` should define `default_notion_home`
  * unix: `${dirs::home_dir()}/.notion` (see [`dirs::home_dir()`](https://docs.rs/dirs/1.0.4/dirs/fn.home_dir.html))
  * windows: `${dirs::data_local_dir()}/Notion` (see [`dirs::data_local_dir()`](https://docs.rs/dirs/1.0.4/dirs/fn.data_local_dir.html))
- all the path functions should root off of `notion_home()`
- eliminate `NOTION_SANDBOX`

It might be best to wait until #181 is done to implement this, since that touches these files a fair amount.

dherman:
Fixes #207.

Reduces the OS-specific logic by building everything on top of a `notion_home()` function. Most of the directory layout is OS-independent.

Also eliminates the Rust compiler warning for the deprecated `std::env::home_dir()` function, replacing it with the recommended `dirs` crate.

This PR is the first step towards making a more abstracted and manageable approach to the Notion directory layout. My goal is to create a `Layout` data structure with methods for extracting all the paths. This will then be something that can be shared by not only Notion, but also the test harness's sandbox setup code, as well as a migration tool for upgrading between different layouts.

### DIFF
--- a/crates/notion-core/src/image/mod.rs
+++ b/crates/notion-core/src/image/mod.rs
@@ -84,13 +84,20 @@ mod test {
     use std;
     use std::path::PathBuf;
 
+    #[cfg(unix)]
+    use dirs;
+
     #[cfg(windows)]
     use winfolder;
 
     fn notion_base() -> PathBuf {
         #[cfg(unix)]
+<<<<<<< HEAD
         return PathBuf::from(std::env::home_dir().expect("Could not get home directory"))
             .join(".notion");
+=======
+        return PathBuf::from(dirs::home_dir().expect("Could not get home directory")).join(".notion");
+>>>>>>> Fix Windows compile errors and warnings, and eliminate another use of `std::env::home_dir()`.
 
         #[cfg(windows)]
         return winfolder::Folder::LocalAppData.path().join("Notion");
--- a/crates/notion-core/src/image/mod.rs
+++ b/crates/notion-core/src/image/mod.rs
@@ -152,7 +152,7 @@ mod test {
     #[cfg(windows)]
     fn test_image_path() {
         let mut pathbufs: Vec<PathBuf> = Vec::new();
-        pathbufs.push(shim_dir());
+        pathbufs.push(shim_dir().unwrap());
         pathbufs.push(PathBuf::from("C:\\\\somebin"));
         pathbufs.push(PathBuf::from("D:\\\\ProbramFlies"));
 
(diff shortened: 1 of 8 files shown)


dherman:
With this PR I think we are 100% warnings-clean on macOS and Linux. (A few warnings left in test modules on Windows.) Progress!

crates/notion-core/src/image/mod.rs
```diff
@@ -89,12 +89,15 @@ mod test {
...
+    #[cfg(unix)]
+    use dirs;
+
     #[cfg(windows)]
     use winfolder;
 
     fn notion_base() -> PathBuf {
         #[cfg(unix)]
-        return PathBuf::from(std::env::home_dir().expect("Could not get home directory")).join(".notion");
+        return PathBuf::from(dirs::home_dir().expect("Could not get home directory")).join(".notion");
```
charlespierce: Do we want to have separate `notion_base` and `shim_dir` functions here for these tests? Can we just replace these with the `path::notion_home()` and `path::shim_dir()` functions in the tests?

It seems like with the two separate implementations, the tests could fail if the `NOTION_HOME` environment variable is set to something other than the default, since the actual implementation will use the correct value, while the test will be checking for the default value.

dherman: I'm not sure I understand this comment? Which tests are you referring to? This code isn't test code. I'm a little confused, sorry!

dherman: > Doesn't the #[cfg(test)] on line 84 above this indicate that the whole block is test code? 

Yeeeeessss it does and I always forget to look at the collapsed code >_<

dherman: > so is there an explicit reason why we aren't just using those functions

Now I get it! Great point, I'll fix that.

PR #228, opened 2018-12-20
Skip Scanning Index with Exact Version

### LINKED ISSUE #227: Prevent Scanning Index for Exact Versions
Currently, if we have an exact version of a tool pinned in `package.json`, but that tool hasn't been downloaded to the local inventory, Notion will first fetch the index of versions for that tool, and then scan the index for a version that matches the exact version we have specified. It will then use that version to build the download URL.

This process makes sense for when we are matching a semver string, as we want to find the first version that matches, but when we already know the exact version, we should be able to shortcut and go straight to downloading. We do need to make sure to capture a 404 response and translate it into a `NoNodeVersionFoundError` / `NoYarnVersionFoundError`

charlespierce:
Closes #227 

Prevent an unnecessary scan of the tool index (either Node or Yarn) by adding a `VersionSpec::Exact` type to represent a known exact version. Then when resolving the version we can shortcut and just return that version instead of doing a scan.

Also updated the error messages in `distro` to handle the 404 response, which will most often occur when an exact version that doesn't exist is specified. Note: I'm especially interested in feedback on the approach taken to refactor the error messages, as I feel like there could be a more elegant way to approach the problem.


crates/notion-core/src/distro/error.rs
```diff
@@ -1,27 +1,69 @@
...
+        tool: Tool,
         version: String,
-        from_url: String,
+        from_url: String
     ) -> impl FnOnce(&failure::Error) -> DownloadError {
-        move |error| DownloadError {
-            version: version,
-            from_url: from_url,
-            error: error.to_string(),
+        move |error| match error.downcast_ref::<HttpError>() {
```
dherman: I might separate the `match` to a new line and brace the closure body:

```rust
move |error| {
    match error.downcast_ref::<HttpError>() {
        // ...
    }
}
```

But otherwise this looks fine to me.

crates/notion-core/src/distro/error.rs
```diff
@@ -1,27 +1,69 @@
...
+            Some(HttpError { code }) if *code == StatusCode::NotFound => DownloadError::NotFound {
```
dherman: If this is what was bothering you, it's fine. :) I'd probably prefer `if`-`let` but you can't concisely check the status code without the [if-let-chains](https://github.com/rust-lang/rust/issues/53667) feature, which hasn't been implemented yet.

crates/notion-core/src/distro/error.rs
```diff
@@ -1,27 +1,69 @@
 //! Provides error types for the installer tools.
 
+use archive::HttpError;
 use notion_fail::{ExitCode, NotionFail};
 
 use failure;
+use reqwest::StatusCode;
+use std::fmt;
+
+// Once Issue #173 is implemented, we can use the ToolSpec struct to differentiate tools
```
dherman: It'd be nice to use a standard syntax for issue comments, esp for greppability. We've been using:

```rust
// ISSUE #173: Once it's implemented we can use the ToolSpec struct etc etc
```

charlespierce:
Updated with the style changes. There wasn't anything specific bothering me about it, just mostly looking for a sanity check that there wasn't a simpler way that I was overlooking :)

dherman:
@charlespierce This looks good. It's kind of a shame `semver::VersionReq` doesn't provide the ability to test for exact versions and extract them as `semver::Version`s, since this means we end up double-parsing a bit. But these are very small strings so it doesn't really matter.

PR #231, opened 2018-12-26
[WIP] Layout abstraction

dherman:
This is just an experimental PR for now, but I thought I'd submit it so others can take a look.

It defines a procedural macro called `layout!`, which lets you declaratively specify the directory layout schema. It makes describing and evolving the directory layout much more frictionless.

You can see the macro in use in v1.rs. It's kinda pretty :)

This should also bring some other benefits:
- Making it easy to created sandboxed layouts for testing
- Making it possible to build migration tools for auto-converting old layouts to new ones

Work checklist:
- [x] Macro syntax validation
- [x] Proper syntax spans for better error reporting (use `quote_spanned!`)
- [x] Rebase to master once we merge #224
- [x] Delete the `path` module
- [x] Refactor `cfg_if::cfg_if!` to `use cfg_if::cfg_if; ... cfg_if!` while I'm in here
- [x] ~Add logic to create and test the `.schema` file for compatibility~

Future work:
- Refactor the test sandbox to use this (especially the handy `.create()` method)
- Optimizations: generate the paths lazily since a given Volta execution may not use them all


crates/notion-core/src/path/windows.rs
```diff
@@ -128,69 +91,22 @@ pub fn node_image_3p_bin_dir(_node: &str, _npm: &str) -> Fallible<PathBuf> {
...
-    Ok(yarn_image_root_dir()?.join(version))
-}
-
-pub fn yarn_image_bin_dir(version: &str) -> Fallible<PathBuf> {
-    Ok(yarn_image_dir(version)?.join("bin"))
-}
-
 pub fn launchbin_file() -> Fallible<PathBuf> {
-    Ok(local_data_root()?.join("launchbin.exe"))
+    Ok(notion_home()?.join("launchbin.exe"))
```
mikrostew: As part of this refactoring you could also pull these executable paths up to `notion-core/src/path/mod.rs` and use `std::env::consts::EXE_SUFFIX` like in the layout code below, to eliminate some duplication. But, since the `path` module will be going away I that's not necessary.

crates/notion-layout/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+
+pub(crate) fn executable(name: &str) -> String {
+    format!("{}{}", name, std::env::consts::EXE_SUFFIX)
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        assert_eq!(2 + 2, 4);
```
mikrostew: 😆

dherman:
@mikrostew 

> I didn't see it in the description but it looks like one of the next steps is to read the `.schema` file and use the appropriate v0/v1/v(whatever) layout for the paths?

I was thinking something a little less ambitious: just a quick startup check to ensure the schema version matches what the current version of Notion expects, and fail if there's a mismatch.

We can also work on a migration tool that can upgrade old layouts to new ones, so that Notion can make seamless upgrades. We'd likely want to keep that in a separate binary (or binaries?) so the `notion` and `shim` executables don't need to grow without bound. But this is all still speculative. I think for right now, just failing fast is a good way to give a clean, understandable error message.

> And since there is no `.schema` file for V0, that seems like it is there mostly for historical reasons but shouldn't actually be used?

Yeah, and I'm deleting it. I had that in the WIP for a while just to have a couple of test cases for the macro.

mikrostew:
As we discussed today, this PR has been open for many months now, and it is not going to be finished any time soon. There are a lot of good ideas here, and they should probably be part of the upgrade/update work, for whoever takes that on.

@dherman the weight is off your shoulders now

Issue #208, opened 2018-12-01
Make "npm" key optional


dherman:
From a [discussion thread](https://github.com/notion-cli/notion/pull/181#discussion_r237960057) in #181, we should make the `"npm"` key optional and default to whatever npm version comes by default with the specified version of Node. Whenever we write to `package.json`, we should prefer to omit the key if the version matches, to encourage minimal syntax.

Depends on #181.

charlespierce:
After a quick investigation, I think that implementing this cleanly depends on #169. Essentially what we want to do is make the Platform Specification allow the `npm` version to be optional, but then resolve that to the default version of npm.

However, currently we don't have a differentiation between Platform Spec and Image, so we run into a timing issue: If the npm version isn't specified, then we need to replace it with the default. When we are parsing `package.json`, there's no guarantee that we know what the default `npm` version is, since we might not yet have downloaded to specified version of `node`. So we really need to wait until we call `Session::prepare_image` before we lock in the version of `npm`, but `Session::prepare_image` requires the `Image` to execute.

After #169 is implemented, however, we can just make that resolution part of the `Platform::checkout` method, since we are making that separation between specification and resolved image.

PR #232, opened 2019-01-08
Platform Specs

### LINKED ISSUE #169: Platform specs
Separate out the concept of _platform image_ and _platform specifier_.

  - [ ] Rename `image::` to `platform::`
  - [ ] Create `platform::Platform` type that contains the versions
  - [ ] Create `Image` type that represents a checked out platform image and contains a `Platform`
  - [ ] Stub `Platform::checkout(self) -> Fallible<Image>`

charlespierce:
Closes #169 

Renamed the `image` module to `platform` and added a `PlatformSpec` type. Currently it is slightly redundant as it holds the exact same information as `Image`, but this will allow the two to diverge.

Updated the `manifest`, `project`, and `toolchain` types to hold a `PlatformSpec` instead of an `Image`, and then made `Session::prepare_image` actually return an `Image` object for the tools to use.

Note that since ensuring all the tools are downloaded and unpacked requires orchestration between different objects, the `checkout` method doesn't exist, and instead the behavior is still provided by `Session::prepare_image`


crates/notion-core/src/platform/mod.rs
```diff
@@ -8,6 +8,15 @@ use distro::node::NodeVersion;
 use notion_fail::{Fallible, ResultExt};
 use path;
 
+/// A specification of tool versions needed for a platform
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub struct PlatformSpec {
+    /// The pinned version of Node.
+    pub node: NodeVersion,
```
dherman: Is your thinking that this will get replaced by something where the npm version is optional? I was gonna say `Image` could contain a `PlatformSpec` instead of duplicating its contents, but if they're going to diverge then this is good as is.

charlespierce: Yeah, I was thinking that the two types are likely to diverge, but figured that for the initial separation keeping them the same made the most sense. After #208 it would probably look something like:

```
pub struct PlatformSpec {
    pub node_runtime: Version,
    pub npm: Option<Version>,
    pub yarn: Option<Version>,
}
```

Thinking about that though, I think once that starts to happen, the `checkout` operation becomes about more than just ensuring everything is downloaded, so it might be good to bring it back and have it be `fn checkout(&self, &Session);` where it then calls out to the `Session` for the "ensure node / yarn are available" part, but then performs the other operations within the `PlatformSpec` method.

I was going to tackle #208 next, since it's all fresh in my mind, I can probably include that refactor in that work, keeping this as a minimal separation that doesn't really change the code much other than some places swapping `Image` for `PlatformSpec`

crates/notion-core/src/session.rs
```diff
@@ -165,26 +165,28 @@ impl Session {
         self.config.get()
     }
 
-    /// Ensures that a platform image has been fully fetched and set up.
-    pub(crate) fn prepare_image(&mut self, platform: &PlatformSpec) -> Fallible<Image> {
+    /// Ensures that a specific Node version has been fetched and unpacked
+    pub(crate) fn ensure_node(&mut self, version: &Version) -> Fallible<()> {
```
dherman: Aside: I feel like this "prepare"/"ensure" operation needs a better name, but this code is going to be refactored a few more times and it's a purely internal operation so I'm not going to worry about it too much yet.

PR #233, opened 2019-01-09
Make "npm" key optional in package.json

### LINKED ISSUE #208: Make "npm" key optional

charlespierce:
Closes #208 

* Updated `PlatformSpec` to hold `node_runtime` and `npm` separately.
* Switched `PlatformSpec::npm` to be an `Option`.
* Included logic in `PlatformSpec::checkout()` that determines the default version of npm when creating an `Image`.
* Prevent writing of `npm` to `package.json` if it is the default for the selected `node` version when pinning the `node` version (currently that is always the case, so `npm` will never be written).
* Also did some minor refactors, removing a `Result` wrapper around `Session::current_platform` and related functions, since all code paths would return `Ok`

### DIFF
--- a/crates/notion-core/src/distro/node.rs
+++ b/crates/notion-core/src/distro/node.rs
@@ -50,6 +50,14 @@ pub struct NodeVersion {
     pub npm: Version,
 }
 
+impl NodeVersion {
+    /// Read the local npm version file to determine the default npm version for a given node version
+    pub fn default_npm_version(node_version: &Version) -> Fallible<Version> {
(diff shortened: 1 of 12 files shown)


charlespierce:
And Resolved Merge Conflicts

Issue #174, opened 2018-10-19
Generalize `Toolchain::install`


dherman:
Generalize the `install_foo` methods of `Toolchain` to one single `install` method.

  - [ ] Refactor `toolchain::install_{node, yarn}` to `toolchain::install(tool: Tool)`
  - [ ] Error on Npx (recommend selecting either a version of npm or a version of node)

Depends on:

  - [x] #170 Separate Inventory and Toolchain types
  - [ ] #173 Tool specs

PR #234, opened 2019-01-10
[WIP] Tool specs and fetch/install/pin refactoring

### LINKED ISSUE #173: Tool specs
Separate out the concept of a _tool_ from a _tool specifier_.

  - [ ] create `tool` directory for the `tool::` module
  - [ ] move shim stuff into `tool::ToolSpec::create_shim`, etc
  - [ ] create `tool::ToolSpec` enum (`Node | Yarn | Npm | Npx | Package(String)`)
  - [ ] change `notion::Install` to a struct of `{ tool: ToolSpec, version: VersionSpec }`
  - [ ] change `tool::Tool` to an enum (`Node | Yarn | Npm | Npx | PackageBinary(String)`)

mikrostew:
This turned into more of a refactor than I was anticipating - there are still a couple things left to finish to make everything work again.

Closes #173 and #174


crates/notion-core/src/inventory/mod.rs
```diff
@@ -108,32 +110,49 @@ impl Inventory {
         })
     }
 
-    /// Fetches a Node version matching the specified semantic versioning requirements.
-    pub fn fetch_node(&mut self, matching: &VersionSpec, config: &Config) -> Fallible<Fetched<NodeVersion>> {
-        let distro = self.node.resolve_remote(matching, config.node.as_ref())?;
-        let fetched = distro.fetch(&self.node).unknown()?;
+    /// Fetches a Tool version matching the specified semantic versioning requirements.
+    pub fn fetch(&mut self, _toolspec: &ToolSpec, _config: &Config) -> Fallible<Fetched<DistroVersion>> {
```
charlespierce: If the idea is to have this be a general-purpose tool `fetch` command, the return type should probably support all of those tools, so this `DistroVersion` will probably need to support all the different types of tools and their associated versions.

Related: It would be nice if we could type this in a way that whatever tool you request, you are required to get that tool's `Version` type back. Right now it looks like you could call `fetch(ToolSpec::Node)` and get a return value that is `DistroVersion::Yarn`, which wouldn't make sense. I'm not sure exactly how to type that, but I feel like it should be possible.

mikrostew: Good idea, it would be nice to use generics to do `fetch(ToolSpec<T>) -> DistroVersion<T>`, something like that. And I'm planning to extend DistroVersion to cover Node/Yarn/Npm/Npx/Package, so that should be enough to cover the things that can be fetched.

mikrostew: I couldn't figure out how to make `fetch` generic for this, because it calls `inventory.fetch()` whcih is not generic-alizable as far as I can tell? Maybe @dherman has some thoughts on this.

crates/notion-core/src/distro/mod.rs
```diff
@@ -33,9 +34,35 @@ impl<V> Fetched<V> {
     }
 }
 
-pub trait Distro: Sized {
-    type VersionDetails;
+/// Abstraction to contain info about Distro versions.
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub enum DistroVersion {
```
dherman: Something feels suspicious about this being an `enum` instead of (1) a trait (2) a separate type for each kind of distro and (3) an associated type in the `Distro` trait. By using an `enum`, we aren't using the type system to guarantee at compile time that e.g. a `NodeDistro` will only ever produce a `DistroVersion::Node`. I think maybe the runtime checks (`if let`) in `FetchResolve<NodeDistro>::fetch()` and `FetchResolve<YarnDistro>::fetch()` are a consequence of this.

crates/notion-core/src/distro/mod.rs
```diff
@@ -33,9 +34,35 @@ impl<V> Fetched<V> {
...
+    // the version of the Node runtime, and the npm version installed with that
+    Node(Version, Version),
+    Yarn(Version),
+    Npm(Version),
+    Npx(Version),
```
dherman: I wonder if we should not have the `Npx` variant, since if we want to support installing it independently from npm, it would just be as a package, so you could use the `Package` variant for that.

crates/notion-core/src/distro/mod.rs
dherman: Ah, I see @charlespierce made the same point. Probably easier to talk through this synchronously…

dherman: So in that thread, you were unsure how to make it generic since you need to call `inventory.fetch()`. One thought is that you could make each collection crate-visible, so the dynamic version is still `inventory.fetch()`, but when you're statically dealing with Node you can call `inventory.node.fetch()`, when you're statically dealing with Yarn you can call `inventory.yarn.fetch()`, etc. Would something like that work?

crates/notion-core/src/inventory/mod.rs
```diff
@@ -159,7 +149,14 @@ impl<D: Distro> Collection<D> {
     }
 }
 
-pub trait Resolve<D: Distro> {
+pub trait FetchResolve<D: Distro> {
+    /// Fetch a Distro version matching the specified semantic versioning requirements.
```
dherman: I don't fully understand why this trait needs the new `fetch` method. Is this to implement the `notion fetch` command? I'm not opposed to it, I'm just not fully understanding the trait.

crates/notion-core/src/project.rs
```diff
@@ -227,29 +225,37 @@ impl Project {
...
-        // update the toolchain yarn version
-        if let Some(image) = self.manifest().platform() {
-            let toolchain =
-                serial::Image::new(image.node.runtime.to_string(), image.node.npm.to_string(), Some(yarn_version.to_string()));
-            Manifest::update_toolchain(toolchain, self.package_file())?;
-            println!("Pinned yarn to version {} in package.json", yarn_version);
-        } else {
-            throw!(NoPinnedNodeVersion::new());
+    /// Writes the specified version of Node or Yarn to the `toolchain` in package.json.
+    pub fn pin_in_toolchain(&self, distro_version: &DistroVersion) -> Fallible<()> {
```
dherman: Maybe at this point the method could just be called `pin`?

crates/notion-core/src/session.rs
```diff
@@ -193,30 +191,27 @@ impl Session {
...
-        self.toolchain.set_active_node(version)?;
-        Ok(())
+        inventory.fetch(&tool, config)
     }
 
-    /// Updates toolchain in package.json with the Node version matching the specified semantic
+    /// Updates toolchain in package.json with the Tool version matching the specified semantic
     /// versioning requirements.
-    pub fn pin_node_version(&mut self, matching: &VersionSpec) -> Fallible<()> {
+    pub fn pin_tool(&mut self, toolspec: &ToolSpec) -> Fallible<()> {
```
dherman: Would this one also read nicely as just `pin`?

crates/notion-core/src/tool/mod.rs
```diff
@@ -54,7 +101,7 @@ impl ToolUnimplementedError {
 }
 
 /// Represents a command-line tool that Notion shims delegate to.
-pub trait Tool: Sized {
+pub trait CmdTool: Sized {
```
dherman: Why the rename?

dherman:
@mikrostew and I spent some time thinking about how to make it more static and it was nonobvious enough that I think it's better to go with the enum for now and file a code quality issue for making it more static later.

crates/notion-core/src/project.rs
mikrostew: fixed

crates/notion-core/src/session.rs
mikrostew: yup! fixed this as well

crates/notion-core/src/distro/mod.rs
mikrostew: removed the Npx variants - those can use the `Package` variant

crates/notion-core/src/inventory/mod.rs
mikrostew: I created this as a way to get rid of the `fetch_node` and `fetch_yarn` methods, and combine them into a single `fetch` function. And I figured that each `Distro` should have `fetch` methods as well as `resolve_*`.

dherman: OK, makes sense.

crates/notion-core/src/inventory/mod.rs
```diff
@@ -159,7 +148,14 @@ impl<D: Distro> Collection<D> {
...
```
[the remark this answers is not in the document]
dherman: Filed #244

PR #239, opened 2019-01-11
Ensure prerelease npm still allow npx usage

rwjblue:
Using `VersionReq::matches` will (intentionally) not match prerelease versions. Instead you must compare via `>=`.

Fixes https://github.com/notion-cli/notion/issues/238

### DIFF
--- a/crates/notion-core/src/tool.rs
+++ b/crates/notion-core/src/tool.rs
@@ -420,9 +420,10 @@ impl Tool for Npx {
         if let Some(ref platform) = session.current_platform()? {
             let image = platform.checkout(session)?;
 
-            // npx was only included with Node >= 8.2.0. If less than that, we should include a helpful error message
-            let required_node = VersionSpec::parse_requirements(">= 5.2.0")?;
-            if required_node.matches(&image.node.npm) {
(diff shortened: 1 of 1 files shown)


crates/notion-core/src/tool.rs
```diff
@@ -420,9 +420,10 @@ impl Tool for Npx {
         if let Some(ref platform) = session.current_platform()? {
             let image = platform.checkout(session)?;
 
-            // npx was only included with Node >= 8.2.0. If less than that, we should include a helpful error message
-            let required_node = VersionSpec::parse_requirements(">= 5.2.0")?;
-            if required_node.matches(&image.node.npm) {
+            // npx was only included with Node >= 8.2.0 which shipped with npm 5.2.0. If the npm
```
charlespierce: Super minor, but I would get rid of the mention of `Node` at all, since we're directly comparing `npm` versions. Something like:

```
npx was only included with npm 5.2.0 and above. If the npm version
is less than that, we should include a helpful error message
```

rwjblue: Good point, updated!

PR #241, opened 2019-01-16
Rename 'Plugin' & 'Config' to 'Hook' and 'HookConfig'

charlespierce:
Implements the changes discussed in https://github.com/notion-cli/rfcs/pull/25

Changes
-----
* Combined `config` and `plugin` modules into a single modules named `hook`.
* Updated `Inventory::fetch` and `Distro` construction to reference the hooks and use them to resolve the URLs necessary for locating the latest version, index and specific distro download.
* Added tests to confirm the behavior of the different hook variants (`Prefix` and `Template`).

Notes
-----
* The `node` hooks for `latest` and `index` are currently treated the same way, they download the index, which is then searched to find the appropriate version. This will probably need to be documented, or if Node has an URL we can hit in the same way as the Yarn `latest-version` endpoint to determine the latest version, we can migrate to that.
* This will likely have conflicts with #234, so once that is merged I will update to resolve conflicts.
* There is a slight issue with this and the Node Index Caching: The cache is just a single file, so there is no way currently for it to distinguish between sources. That means if the index is cached and the hook changes so that it should be downloading from somewhere else, the cache will still be used. Filed #242 to address a more robust cache.


charlespierce:
Resolved Merge Conflicts

crates/notion-core/src/hook/tool.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl DistroHook {
+    /// Performs resolution of the Distro URL based on the given
+    /// Version and File Name
+    pub fn resolve(&self, version: &Version, filename: &str) -> Fallible<String> {
+        match self {
+            &DistroHook::Prefix(ref prefix) => Ok(format!("{}{}", prefix, filename)),
+            &DistroHook::Template(ref template) => Ok(template
+                .replace(ARCH_TEMPLATE, ARCH)
```
dherman: I think I'd want to see this templating logic get pulled out into a separate "simple templating" crate. I spent some time researching existing templating engines in the Rust ecosystem, and they all seem too feature-rich for our purposes here. I think what we want is something minimal. But we still probably want a _tiny_ bit of functionality, e.g. the ability to escape `'{'` characters just in case. So I think it'd be a little cleaner to pull this out into a crate (in-repo at least for now). I don't care deeply whether we pull it out now or in a followup issue.

crates/notion-core/src/hook/tool.rs
```diff
@@ -0,0 +1,160 @@
...
+            &MetadataHook::Prefix(ref prefix) => Ok(format!("{}{}", prefix, filename)),
+            &MetadataHook::Template(ref template) => Ok(template
+                .replace(ARCH_TEMPLATE, ARCH)
+                .replace(OS_TEMPLATE, OS)),
+            &MetadataHook::Bin(ref bin) => execute_binary(bin, None),
+        }
+    }
+}
+
+fn execute_binary(bin: &str, extra_arg: Option<String>) -> Fallible<String> {
```
dherman: Add a brief doc comment to describe the purpose of this function.

crates/notion-core/src/hook/tool.rs
charlespierce: Filed #260 for pulling the behavior into a separate crate, as this PR is big enough already, and I think there are some outstanding questions for how we want the templating to work. To reduce the likelihood of needing an escape, however, I also updated the template strings to use `{{` and `}}`, instead of single braces.

Since it's currently using string replaces on a specific set of template strings, we don't have to worry about escaping arbitrary `{{` sequences. It will only be a problem if the user wants to have a literal `{{arch}}`, `{{os}}`, or `{{version}}` in their URL that they would need to escape.

PR #249, opened 2019-01-30
Consolidate errors to streamline nice error quest issue.

charlespierce:
Info
-----
* To make it easier to tackle improving the error messages across the app, we want to streamline the process of adding a new error message.
* One approach to that is to Consolidate all of the errors into a single `ErrorDetails` enum, so that new errors can simply be added to that enum, as opposed to creating entirely new `struct`s throughout the app.

Changes
-----
- [x] Add a new module `error` that exposes an `InternalError` enum which implements `NotionFail`
- [x] Move error objects in `fs` into the enum
- [x] Move error objects in `inventory` into the enum
- [x] Move error objects in `manifest` into the enum
- [x] Move error objects in `path` into the enum
- [x] Move error objects in `project` into the enum
- [x] Move error objects in `session` into the enum
- [x] Move error objects in `shell` into the enum
- [x] Move error objects in `shim` into the enum
- [x] Move error objects in `tool` into the enum
- [x] Move error objects in `version` into the enum
- [x] Move error objects in root `src` into the enum
- [x] Restore comments about the intent of some errors

Notes
-----
* I also removed the `shim` command, it wasn't compiling after other changes and will shortly be replaced by the Package Installs.
* The approach to adding a new Error message is now:
    1. Add a new entry to the `ErrorDetails` enum, including information you need for the error message if necessary.
    2. Update the `match` block in the `impl Display for ErrorDetails` block to include the error message for the new error.
    3. Update the `match` block in the `impl NotionFail for ErrorDetails` block to return the correct `ExitCode` for the new error.
    4. Replace any `.unknown()` calls that should wrap an failure in the new error with a `.with_context(|error| ErrorDetails::NewError)` call, setting any contextual values needed.


crates/notion-core/src/error.rs
```diff
@@ -204,6 +207,7 @@ impl NotionFail for ErrorDetails {
             ErrorDetails::DepPackageReadError { .. } => ExitCode::FileSystemError,
             ErrorDetails::DownloadToolNetworkError { .. } => ExitCode::NetworkError,
             ErrorDetails::DownloadToolNotFound { .. } => ExitCode::NoVersionMatch,
+            ErrorDetails::InvalidHookCommand { .. } => ExitCode::UnknownError,
```
benblank: We should probably clean up these `UnknownError`s at some point; IIRC, that code was meant to only be used for "unfriendly" errors (i.e. errors not migrated to this enum).

charlespierce: Agreed, the next step in this will be a quest issue to handle all of the "unknown" errors, which will include all of the calls to `.unknown()` as well as fixing up these `UnknownError` exit codes to be more descriptive.

benblank: 👍

crates/notion-core/src/shell/mod.rs
```diff
@@ -44,18 +37,12 @@ pub trait Shell {
...
-#[notion_fail(code = "EnvironmentError")]
-struct UnspecifiedShellError;
-
 impl CurrentShell {
     pub fn detect() -> Fallible<Self> {
         match env::shell_name() {
             Some(name) => Ok(name.parse()?),
             None => {
-                throw!(UnspecifiedShellError);
+                throw!(ErrorDetails::UnspecifiedShell);
```
chriskrycho: Another place that the change made me notice—here gain the combinator approach yields at least a *different* style:


```rust
pub fn detect() -> Fallible<Self> {
    env::shell_name()
        .ok_or(ErrorDetails::UnspecifiedShell.into())
        .and_then(|name| name.parse())
}
```

crates/notion-core/src/shell/mod.rs
```diff
@@ -73,24 +60,16 @@ impl Shell for CurrentShell {
...
-struct UnrecognizedShellError {
-    name: String,
-}
-
 impl FromStr for CurrentShell {
     type Err = NotionError;
 
     fn from_str(src: &str) -> Result<Self, NotionError> {
-        let postscript_path = env::postscript_path().ok_or(UnspecifiedPostscriptError)?;
+        let postscript_path = env::postscript_path().ok_or(ErrorDetails::UnspecifiedPostscript)?;
```
chriskrycho: This needs a merge from `master`, it looks like: the logic here doesn't include the changes I made for fish support.

crates/notion-core/src/shell/mod.rs
```diff
@@ -73,24 +60,16 @@ impl Shell for CurrentShell {
...
 
         Ok(CurrentShell(match src {
             "bash" => Box::new(Bash { postscript_path }),
             _ => {
-                throw!(UnrecognizedShellError {
+                throw!(ErrorDetails::UnrecognizedShell {
```
chriskrycho: While we're doing the work to include the changes made for adding fish support here, we should tweak this implementation: wrapping a `throw!` *inside* an `Ok(...)` is super confusing, at least to me. It took me a few to figure out exactly what I was seeing. The actual semantics here, and therefore, what I'd recommend, are:

```rust
fn from_str(src: &str) -> Result<Self, NotionError> {
    let postscript_path = env::postscript_path().ok_or(ErrorDetails::UnspecifiedPostscript)?;

    match src {
        "bash" => Ok(CurrentShell(Box::new(Bash { postscript_path }))),
        "fish" => Ok(CurrentShell(Box::new(Fish { postscript_path }))),
        _ => Err(ErrorDetails::UnrecognizedShell {
            name: src.to_string(),
        }
        .into()),
    }
}
```

If there's a preference for keeping `throw!` instead of using `.into()`, you can do this:

```rust
fn from_str(src: &str) -> Result<Self, NotionError> {
    let postscript_path = env::postscript_path().ok_or(ErrorDetails::UnspecifiedPostscript)?;

    match src {
        "bash" => Ok(CurrentShell(Box::new(Bash { postscript_path }))),
        "fish" => Ok(CurrentShell(Box::new(Fish { postscript_path }))),
        _ => throw!(ErrorDetails::UnrecognizedShell {
            name: src.to_string(),
        }),
    }
}
```

Personally, I find that less clear, simply because there's such a nice symmetry with `Ok` and `Err` in the basic version.

crates/notion-core/src/distro/mod.rs
```diff
@@ -73,3 +76,23 @@ pub trait Distro: Sized {
     /// to update its state after fetching succeeds.)
     fn fetch(self, collection: &Collection<Self>) -> Fallible<Fetched<DistroVersion>>;
 }
+
+fn error_for_tool(
+    toolspec: ToolSpec,
+    from_url: String,
+) -> impl FnOnce(&failure::Error) -> ErrorDetails {
+    move |error| {
```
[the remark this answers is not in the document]
charlespierce: Looks good, updated!

crates/notion-core/src/fs.rs
```diff
@@ -18,36 +16,20 @@ pub fn touch(path: &Path) -> Fallible<File> {
...
-impl CreateDirError {
-    pub(crate) fn for_dir(dir: String) -> impl FnOnce(&io::Error) -> CreateDirError {
-        move |error| CreateDirError {
-            dir,
-            error: error.to_string(),
-        }
+fn error_for_dir(dir: String) -> impl FnOnce(&io::Error) -> ErrorDetails {
+    move |error| ErrorDetails::CreateDirError {
+        dir,
+        error: error.to_string(),
```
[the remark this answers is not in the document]
charlespierce: This is a great idea for the general case! I think I'll update the plan for adapting the `.unknown()` calls to use an `impl From` approach, as it makes the call to `with_context` more expressive (`.with_context(ErrorDetails::from)`).

In this case, we actually need a bespoke function, because we need the additional context provided by the `dir` parameter to create the error.

I agree about losing the detail of the underlying error. There are a number of the existing `NotionFail`s that incorporate an `error` string, rather than the underlying error. Cleaning those up to more accurately include the full error in the failure is a good next step.

crates/notion-core/src/fs.rs
```diff
@@ -18,36 +16,20 @@ pub fn touch(path: &Path) -> Fallible<File> {
...
-#[fail(display = "`path` internal error")]
-#[notion_fail(code = "UnknownError")]
-pub(crate) struct PathInternalError;
-
 /// This creates the parent directory of the input path, assuming the input path is a file.
 pub fn ensure_containing_dir_exists<P: AsRef<Path>>(path: &P) -> Fallible<()> {
     if let Some(dir) = path.as_ref().parent() {
-        fs::create_dir_all(dir)
-            .with_context(CreateDirError::for_dir(dir.to_string_lossy().to_string()))
+        fs::create_dir_all(dir).with_context(error_for_dir(dir.to_string_lossy().to_string()))
```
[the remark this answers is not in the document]
charlespierce: I'm going to have to remember to think of the `Option::ok_or()` method more :) That's a good suggestion.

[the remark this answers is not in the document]
chriskrycho: Writing [a TypeScript library](https://github.com/true-myth/true-myth) based heavily on Rust's `Result` and `Option` types and using it heavily for the last two years has the hilarious side effect of meaning I know the available Rust combinators for these two absurdly well. 😆

crates/notion-core/src/shell/mod.rs
charlespierce: I'm not sure if this is particularly relevant, but one benefit to using the `match` approach is that the `ErrorDetails::UnspecifiedShell` object isn't created if there is no error. When using `.ok_or()`, we need to construct the error in order to pass the parameter. With the control flow from the `match` block, we only construct the error when we are about to return it.

We could get around that by using `.ok_or_else()`, but that would then create the closure function to pass as a parameter, so we're still constructing something.

All that said, this is almost assuredly premature optimization (and might not actually have any real benefits with compiler optimizations). Your formulation is a lot more readable, so I think until we notice major performance issues, going with readability is better :+1:

chriskrycho: Ah, yeah, that's definitely worth keeping in mind. One thing I keep reminding myself is that `match` is *good* – I have a bunch of deeply ingrained habits from living in JS/TS-land where there *is* no `match`, and am trying to remember: `match` is idiomatic! 😆

crates/notion-core/src/shell/mod.rs
charlespierce: Agreed, that's a lot more explicit about the flow.

crates/notion-core/src/fs.rs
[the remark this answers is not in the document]
charlespierce: Thinking on this more, we need to make sure we align what the errors are supposed to represent and don't throw away context. Specifically, the underlying errors that we are getting are almost always descriptions of what went wrong, while the `NotionError` we want to present to the user is "This is what we were trying to do, and here's what happened when we did that". We're always going to be adding that additional context about what we were trying to do, so I'm not sure that there will be a good general conversion between underlying errors and `NotionError`s.

That said, it's definitely good to keep an eye out for spots where we could use an idiomatic construct like `impl From`, as opposed to rolling our own.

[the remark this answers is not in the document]
chriskrycho: Yeah, I figured there'd be limitations to what we can do—and you certainly have a better sense of that space than I do at this point! Even if we end up throwing away most of it, using richer data to construct our own error when it's available (i.e. *not* just doing a `to_string()` if there's more we can use) seems like it'll be a nice win.

chriseppstein:
I want these nice errors. I am tired of the mean errors.

crates/notion-core/src/shim.rs
```diff
@@ -2,29 +2,18 @@
...
-    pub(crate) fn from_io_error(error: &io::Error) -> Self {
-        if let Some(inner_err) = error.get_ref() {
-            SymlinkError {
-                error: inner_err.to_string(),
-            }
-        } else {
-            SymlinkError {
-                error: error.to_string(),
-            }
+fn from_io_error(error: &io::Error) -> ErrorDetails {
```
dherman: Use a name like `symlink_error` instead, since you're trying to express what kind of context you're providing to `with_context`.

crates/notion-core/src/tool/mod.rs
```diff
@@ -81,23 +79,14 @@ impl Display for ToolSpec {
...
-    pub(crate) fn from_io_error(error: &io::Error) -> Self {
-        if let Some(inner_err) = error.get_ref() {
-            BinaryExecError {
-                error: inner_err.to_string(),
-            }
-        } else {
-            BinaryExecError {
-                error: error.to_string(),
-            }
+fn from_io_error(error: &io::Error) -> ErrorDetails {
```
dherman: Similar to symlink_error, this should be named something like `binary_exec_error`.

crates/notion-core/src/version/mod.rs
```diff
@@ -70,23 +68,14 @@ impl FromStr for VersionSpec {
...
-pub(crate) struct VersionParseError {
-    pub(crate) error: String,
-}
-
-impl VersionParseError {
-    fn from_req_parse_error(error: &ReqParseError) -> Self {
-        VersionParseError {
-            error: error.to_string(),
-        }
+fn from_req_parse_error(error: &ReqParseError) -> ErrorDetails {
```
dherman: I don't _love_ top-level functions with "from_xxx" names, and this is related to the other naming suggestions I made. But in this case you have two different semver parse errors going into one `ErrorDetails` variant, so you can't just have one name. Maybe just `version_req_parse_error` and `version_parse_error`.

src/error.rs
```diff
@@ -1,28 +1,18 @@
...
-        } else {
-            CliParseError {
-                usage: None,
-                error: error.to_string(),
-            }
+use failure::Context;
+use notion_core::error::ErrorDetails;
+use notion_fail::NotionError;
+
+pub(crate) fn from_docopt_error(error: &docopt::Error) -> ErrorDetails {
```
dherman: Again I don't think a top-level function should start with `from`. Probably `cli_parse_error` is good.

crates/notion-core/src/distro/mod.rs
```diff
@@ -73,3 +76,19 @@ pub trait Distro: Sized {
     /// to update its state after fetching succeeds.)
     fn fetch(self, collection: &Collection<Self>) -> Fallible<Fetched<DistroVersion>>;
 }
+
+fn error_for_tool(
```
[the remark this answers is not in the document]
dherman: After reading the full PR, I could be convinced either way about whether you make `fetch_error` a method of `ToolSpec` or a top-level function as you have it.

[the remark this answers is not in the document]
charlespierce: Agreed about the naming and DRYing the `.to_string()` call. I think at first pass that it makes more sense to remain as a top-level function. That matches the behavior elsewhere of using a conversion function to provide the needed error context. It also seems like adding it as a method to the `ToolSpec` would be more indirection than is appropriate.

charlespierce:
Updated to rename all the top-level error context functions :+1:

PR #250, opened 2019-01-31
Update to Rust 2018 Edition

### LINKED ISSUE #251: Update to Rust 2018
Update to the Rust 2018 Edition. Lots of things should get cleaner by doing so, *especially* paths!

chriskrycho:
Resolves #251.

- [x] Do `cargo fix --edition`
- [x] Update the edition flag in `Cargo.toml` files
- [x] Attempt `cargo fix --edition-idioms` (it may or may not work)
- [x] In line with Rust 2018 idiom (and as supported since Rust 1.30), switch from using `#[macro_use]` on imports to explicitly importing the macro to be applied.


charlespierce:
For `serde` specifically, it looks like we actually don't have to import `serde_derive` explicitly at all. I just saw this on Twitter today, appropriately: https://twitter.com/RReverser/status/1090967358049607680

chriskrycho:
Ooooh, snazzy. I'll update!

chriskrycho:
And, follow-up on that, I find that failure (at least the version we're on currently) doesn't work nicely without it. Going to see if upgrading that solves this.

chriskrycho:
@mikrostew we could potentially look at Clippy and carefully enable rules we care about rather than global rulesets, but that seems like a bigger discussion.

chriskrycho:
As I was pushing through to finish getting things in line with the 2018 Edition idioms in the areas mentioned above, I hit a number of quirky cases around `notion_fail` the *module* vs. `notion_fail` the macro – mostly have those resolved, and should wrap it up early on Monday.

chriskrycho:
Ugh. I broke something in the smoke tests while attempting to get rid of the rest of the extern crate and macro_use stuff. Will chase that down.

PR #257, opened 2019-02-06
tmp files in NOTION_HOME

dschmidt:


crates/notion-core/src/distro/yarn.rs
```diff
@@ -117,14 +117,20 @@ impl Distro for YarnDistro {
...
+
+        let dest = path::yarn_image_root_dir()?;
+
+        ensure_containing_dir_exists(&dest)?;
+
         rename(
-            dest.join(path::yarn_archive_root_dir_name(&version_string)),
+            temp.path()
+                .join(path::yarn_archive_root_dir_name(&version_string)),
             path::yarn_image_dir(&version_string)?,
```
charlespierce: The `dest` in this block should be the final destination of the unpacked archive, not the same as the old variable that was (unfortunately) called `dest`. To be analogous with the `node` version of this, `let dest = path::yarn_image_dir(&version_string)?;`, then you can just use `dest` inside the call to rename.

dschmidt: Ah, yes, that makes a lot more sense. Thanks!

charlespierce:
Thanks for reporting and tracking this down!

dschmidt:
Thanks for your patience and getting me started with rust :)

PR #171, opened 2018-10-19
[WIP] First pass at fish shell support

scalvert:
Adding fish shell support to notion (#152)

TODO:

- [x] Adding a backend (fish.rs) to crates/notion-core/src/shell/, with a type Fish that implements Shell
- [ ] Adding a bootup script (notion.fish) to shell/unix/, with a wrapper function notion to get installed at console boot time
- [ ] Edit install.sh.in to include installing notion.fish to ~/.config/fish/functions

Issue #99, opened 2018-07-20
escape `PATH` when writing postscript


dherman:
`notion deactivate` writes a `PATH` variable and should be careful to correctly escape the string.

PR #266, opened 2019-02-09
Add fish support 🐟 

### LINKED ISSUE #152: fish support
Lots of devs use the [fish](https://fishshell.com/) shell (hi @stefanpenner, hi @scalvert!). We should support it!

This will require:
- [ ] Adding a backend (`fish.rs`) to `crates/notion-core/src/shell/`, with a type `Fish` that implements `Shell`
- [ ] Adding a bootup script (`load.fish`) to `shell/unix/`, with a wrapper function `notion` to get installed at console boot time
- [ ] Adding an installer (`install.fish`) to `dev/unix/`

That last task is probably blocked on #151.

chriskrycho:
This PR builds on the work @scalvert started in #171 to add full fish support to Notion.

- closes #171
- resolves #152

## Testing

1. Check out this branch locally. If you're using [`hub`], `hub pr checkout 266`. (If you're not using `hub`… you should do that thing. And also alias it to `git` to make your life wonderful.)

2. Create a release build: `cargo build --release`.

3. Do the equivalent of running the install script: `./dev/unix/build.sh; and ./dev/unix/install.sh`.

4. Open a new shell, or refresh your current shell with the updated config: `source ~/.config/fish/config.fish`.

[`hub`]: https://hub.github.com

Then you should be able to use Notion!


crates/notion-core/src/shell/fish.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[test]
+    fn test_compile_postscript() {
+        let fish = CurrentShell::from_str("fish").expect("Could not create fish shell");
+
+        assert_eq!(
+            fish.compile_postscript(&Postscript::Deactivate("some:path".to_string())),
+            "set -U fish_user_paths 'some:path'\nset -e NOTION_HOME\n"
+        );
+
+        assert_eq!(
```
benblank: Just so it doesn't get lost:  the "ISSUE(99)" comment should probably go here, as well.

chriskrycho: Ah, good point; will do.

scalvert:
@chriskrycho thank you, sir!

chriskrycho:
@scalvert @chadhietala @locks can you all do me a favor and make sure this works?

## Testing

1. Check out this branch locally. If you're using [`hub`], `hub pr checkout 266`. (If you're not using `hub`… you should do that thing. And also alias it to `git` to make your life wonderful.)

2. Create a release build: `cargo build --release`.

3. Do the equivalent of running the install script: `./dev/unix/build.sh; and ./dev/unix/install.sh`.

4. Open a new shell, or refresh your current shell with the updated config: `source ~/.config/fish/config.fish`.

[`hub`]: https://hub.github.com

Then you should be able to use Notion!

chadhietala:
nod

chadhietala:
I think there might be some issues with `deactivate`?

```
set: Warning: $PATH entry "/Applications/Visual Studio Code.app/Contents/Resources/app/bin/:/Users/chietala/Code/binaryen/bin:/Users/chietala/.cargo/bin:/Users/chietala/.config/node/default/bin:/Users/chietala/.yarn/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/linkedin/bin:/export/content/linkedin/bin" is not valid (No such file or directory)
set: Did you mean 'set PATH $PATH /Users/chietala/Code/binaryen/bin:/Users/chietala/.cargo/bin:/Users/chietala/.config/node/default/bin:/Users/chietala/.yarn/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/linkedin/bin:/export/content/linkedin/bin'?
```

chriskrycho:
Ah, so, I *think* this is further up than the changes I made, but I know what the issue is: fish expects a path where entries are separated by spaces rather than `:`. The problem is the path as we're getting it in `System` ([here](https://github.com/chriskrycho/notion/blob/fish-support/crates/notion-core/src/platform/mod.rs#L87-L113)). And that in turn is from yet *further* upstream, in [envoy](https://crates.io/crates/envoy), which in turn just uses [`std::var_os`](https://doc.rust-lang.org/1.32.0/std/env/fn.var_os.html).

I think, to make this do the right thing, I'll just need to add some special handling here, because what we have here is just what the OS reports back to us out of its `getenv`, so far as I can see. I'll double-check that analysis tomorrow, and add some special handling for it if so.

Should just be able to do something like `path.replace(":", " ")` when we're in fish, but I'm not sure whether locations can includes `:` if quoted.

chriskrycho:
@chadhietala an update on the above: I can't reproduce the error you're seeing, but *also* running `notion activate` doesn't actually seem to be doing anything. 🤔

chriskrycho:
That path style actually seems to be fine; the real issue is that we're not quoting the value properly! I.e. this is part of #99.

chriskrycho:
@chadhietala the above *should* have fixed it (as well as, you know, just making `activate` and `deactivate` do what they should) – can you test again and confirm?

scalvert:
Checking!

chadhietala:
No dice.
screen shot 2019-02-14 at 9 25 49 am

I think everything might be hosed from when I tried to `deactivate` last time. Where does Notion write these shell files? I looked at `~/.config/fish/config.fish` does it write anywhere else?

chriskrycho:
Interesting. The quoting *should* have fixed this. 🤔 

It's not writing it to anywhere in the deactivate/activate phase; it's just changing what the exported `PATH` value is for the shell and any children.

chriskrycho:
@chadhietala and I did some digging, and the issue around `notion deactivate` is that versions of fish earlier than 3.0 complain if you attempt to set a value of `PATH` (except, apparently, in `.config/fish/config.fish`?) that includes a non-existent/non-directory location, it prints a warning, exits with status `1`, and does not update the `PATH`. [Fish 3.0.0] changed this behavior so it works correctly.

[Fish 3.0.0]: https://github.com/fish-shell/fish-shell/releases/tag/3.0.0

This isn't something Notion can (or even should) control or work around: the exact same thing happens if you just run `set -x PATH` on the CLI with an invalid location on older versions of fish.

As such, the best solution here seems to be documentation: let users know that the two solutions are:

- to edit their `PATH` in their `~/.config/fish/config.fish` file to remove any items which do not exist on their system
- to upgrade to fish ^3.0

@dherman, thoughts on where exactly we should document that? I was thinking in the `notion activate` and `notion deactivate` docs, since that's where this comes up?

charlespierce:
@chriskrycho Would it be possible to add something to the `load.fish` script that checks the exit code after `source $NOTION_POSTSCRIPT` and print a message if it finds a `1` there? That would be closest to the actual issue and would let a user know after they run `notion activate` or `notion deactivate` that it didn't work and they'll need to make changes

chriskrycho:
We’d have to put it in the fish postscript itself, actually; I considered that but wasn’t sure how much value that adds vs. just documenting it clearly and searchably. Let me experiment with doing just that and see how it comes out.

chriskrycho:
Okay, I've made that change and it looks like this:

screen shot 2019-02-14 at 13 36 40

Wording changes and formatting suggestions *very* welcome!

scalvert:
Everything is fine until the post-install step. Sourcing `~/.config/fish/config.fish` doesn't result in notion having been added to my path. If I `cat ~/.config/fish/config.fish`, it's unmodified. I would have expected notion's install to modify it.

chriskrycho:
Weird. It worked fine for me and others; I will follow up with you in the morning!

chriskrycho:
@scalvert and I synced up; the problem here was that his `$SHELL` was `/etc/bash` instead of fish. It did correctly set up his bash profile! Once he updated his shell everything Just Worked™.

scalvert:
So the issue I had ended up being that my default shell was bash. So, the install was assuming, via the `$SHELL` env var, that I was using bash, not fish. I use iTerm2, and have it configured to run fish, which is why my env var wasn't set. 

Setting `$SHELL` to fish solved the install problem.

stefanpenner:
🕺

PR #270, opened 2019-02-11
Multiple Linux builds

dherman:
Update the bootstrap installer (ie the https://get.notionjs.com script) to choose between multiple Linux builds for Notion >= 0.2, for different versions of OpenSSL.

### DIFF
--- a/dev/unix/boot-install.sh
+++ b/dev/unix/boot-install.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# This is the bootstrap Unix installer served by `https://get.notionjs.com`.
+# Its responsibility is to query the system to determine what OS (and in the
+# case of Linux, what OpenSSL version) the system has, and then proceed to
+# fetch and install the appropriate build of Notion.
+
 notion_get_latest_release() {
   curl --silent https://www.notionjs.com/latest-version
 }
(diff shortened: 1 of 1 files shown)


dev/unix/boot-install.sh
```diff
@@ -28,20 +33,29 @@ notion_warning() {
...
-  unset -f notion_get_latest_release notion_eprintf notion_info notion_error notion_warning notion_exit notion_cleanup
+# determines the major and minor version of OpenSSL on the system
+notion_get_openssl_version() {
+  local LIB
+  local FULLVERSION
+  local MAJOR
+  local MINOR
+  # By default, we'll guess OpenSSL 1.0.1.
+  LIB="$(openssl version 2>/dev/null || echo 'OpenSSL 1.0.1')"
+  FULLVERSION="$(echo $LIB | awk '{print $2;}')"
```
charlespierce: This gives an unexpected Version if the user is using `LibreSSL` instead of `OpenSSL`. For example, calling `openssl version` on my Mac Laptop returns `LibreSSL 2.6.5`. Obviously that's not directly relevant since the MacOS installer is separate, but for Linux users on LibreSSL, this will probably result in a download failure when trying to download `linux-openssl-2.6` which doesn't exist.

I'm not sure if we want to try to support LibreSSL on Linux in the installer right now (Do we need to build a separate binary to handle those cases as well?), just something to think about.

dherman: @charlespierce Good point. How about for now we just fail fast if the name isn't OpenSSL, and we can file an issue to support it eventually.

charlespierce: Yeah, sounds like the cleanest approach :+1:

PR #271, opened 2019-02-12
Non-interactive installer

dherman:
The installer can't be interactive since it's piped from curl. And it's not a good idea to be trying to get too clever about uninstalling (`rm` is scary).

So this PR changes the installer to make it a hard error if an old Notion installation is found. The check is moved into the bootstrap script to avoid users having to download the (quite big) installer twice.

### DIFF
--- a/dev/unix/boot-install.sh
+++ b/dev/unix/boot-install.sh
@@ -24,13 +24,46 @@ notion_info() {
 notion_error() {
   command printf '\033[1;31mError\033[0m: ' 1>&2
   notion_eprintf "$1"
-  notion_eprintf ''
 }
 
 notion_warning() {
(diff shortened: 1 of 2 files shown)


dev/unix/boot-install.sh
```diff
@@ -24,7 +24,6 @@ notion_info() {
 notion_error() {
   command printf '\033[1;31mError\033[0m: ' 1>&2
   notion_eprintf "$1"
-  notion_eprintf ''
```
charlespierce: The other printing methods (including the newly added `notion_request()`) still have this extra empty line. Is there a benefit to having them all be consistent with how they output text, either with or without the additional empty line?

dherman: It was just a result of tweaking the aesthetics until I could get the error message looking good to my eye:

image

PR #273, opened 2019-02-12
Move preventing globals behind a feature flag

charlespierce:
As noted in #272, we shouldn't intercept globals and direct the user to `notion install <package>` before the general package installs are working. Update to lock that behavior behind a feature flag (`intercept-globals`) until we get the package installs working.


dherman:
Looks like you need to disable the tests as well.

charlespierce:
Whoops, updated!

PR #283, opened 2019-02-21
Replace Docopt with StructOpt

### LINKED ISSUE #285: Consistency with version argument (-V)
Currently notion supports -V (capital v) for "version" and -v (lowercase v) for "verbose". This seems to be inverted and goes against common norms for example:

`npm -v`, `yarn -v`,`node -v`, `ember -v`, etc all support -v as version.

This seems like a nice UX improvement to stay consistent with other tools.

chriskrycho:
- implements notion-cli/rfcs#30
- resolves #285

Below are examples of the improvements, with nice colors, built-in help output, and support for both using and generating tab completions.

<details>
<summary><code>notion</code> (or <code>notion help</code>) output</summary>
Screen Shot 2019-03-11 at 13 09 11
</details>

<details>
<summary><code>notion help fetch</code> output</summary>
Screen Shot 2019-03-11 at 13 10 23
</details>

<details>
<summary><code>notion</code> tab completion</summary>
Screen Shot 2019-03-11 at 13 09 31
</details>

<details>
<summary><code>notion help completions</code> output</summary>
Screen Shot 2019-03-11 at 13 09 55
</details>


src/cli.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[derive(StructOpt)]
+#[structopt(
+    name = "Notion",
+    about = "The hassle-free JavaScript toolchain manager",
+    author = "",
+    usage = "notion [flags] [<command> <args> ...]
+
+    To install a tool in your user toolchain, use `notion install`.
+    To pin a tool in a project toolchain, use `notion pin`.",
```
charlespierce: Can you speak to the rationale behind making this part of the general help message, instead of being part of the `Use` command? It seems a little odd to have a few specific subcommands called out in the `USAGE` block, while everything is shown in the `SUBCOMMANDS` block.

src/command/use_.rs
```diff
@@ -1,50 +0,0 @@
-// Rust doesn't allow using keywords as module names so we have to call this `use_`.
```
charlespierce: Related to the above comment, is there a way in StructOpt to have "undocumented" commands that parse correctly but aren't included in the help message? Specifically, when we renamed from `use` to `pin`, we left this command around with the help message that guides the user to the correct subcommand based on what they are trying to do. Is there a way to recreate that behavior?

src/cli.rs
chriskrycho: Ah, I see: I misunderstood what was going on there! Here and for your question below: we can totally have a hidden subcommand that displays this for backwards compatibility. I'll push a commit to that effect shortly!

src/command/use_.rs
chriskrycho: There is, and I've used a `macro_rules`-style macro so the same text gets dumped both for help output and when the command executes; see 7589e8f! Thanks again for catching this.

dherman:
Some suggestions for the top-level help text:

> To install a tool in your toolchain, use `notion install`.
> To pin your project's runtime or package manager, use `notion pin`.

Suggestion for the `install` subcommand help text:

> install        Install a tool in your toolchain.

Suggestion for the `pin` subcommand help text:

> pin      Pin your project's runtime or package manager.

src/command/use.rs
```diff
@@ -0,0 +1,29 @@
...
+use notion_core::session::{ActivityKind, Session};
+use notion_fail::{ExitCode, Fallible};
+
+#[macro_export]
+macro_rules! usage {
+    () => {
+        "notion-use
+
+DEPRECATED:
+    To install a tool in your user toolchain, use `notion install`.
```
dherman: This text should also be updated, similar to the top-level help text.

chriskrycho: Done!

PR #293, opened 2019-03-07
Add `notion which` command

### LINKED ISSUE #167: `notion which`
As discussed in the [package binaries RFC thread](https://github.com/notion-cli/rfcs/pull/21#issuecomment-430455523), we need a `notion which` command for "seeing through a shim."

charlespierce:
Closes #167 

Info
-----
Adds a command `notion which <BINARY>` that shows the user which binary will actually be run, effectively seeing through the shims. It will also resolve any other binary that happens to be on the path, so it works as a drop-in replacement for `which`:

```bash
$ which node
/Users/me/.notion/bin/node
$ notion which node
/Users/me/.notion/tools/image/node/10.15.3/6.4.1/bin/node
$ which ls
/bin/ls
$ notion which ls
/bin/ls
```

Notes
-----
* Currently this command will use the system path (with Notion shim directory removed) as the `PATH` if there is no platform _or_ if there are any errors when attempting to retrieve the platform image. Errors with determining the system path or the current working directory will still be exposed as errors, however.
* This will likely need some updated logic to resolve package binaries from wherever they are installed, since they aren't directly on the PATH.

### DIFF
--- a/src/command/which.rs
+++ b/src/command/which.rs
@@ -3,9 +3,9 @@ use std::env;
 use structopt::StructOpt;
 use which::which_in;
 
-use notion_core::error::ErrorDetails;
+use notion_core::platform::System;
 use notion_core::session::{ActivityKind, Session};
-use notion_fail::{throw, ExitCode, Fallible, ResultExt};
+use notion_fail::{ExitCode, Fallible, ResultExt};
 
 use crate::command::Command;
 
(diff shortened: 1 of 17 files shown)


charlespierce:
Updated to use the System path (with Notion shim directory removed) as the fallback path, reducing the potential error cases and allowing the command to more closely mimic the expected behavior. For example, if the user has not run `notion install node` and runs `notion which node`, the result will be no output, the same as `which node` would do if the user had never installed a global `node` executable.

src/command/which.rs
```diff
@@ -0,0 +1,49 @@
...
+            Ok(result) => {
+                println!("{}", result.to_string_lossy());
+            }
+            Err(_) => {
+                // `which_in` Will return an Err if it can't find the binary in the path
+                // In that case, we want to do nothing, instead of showing the user an error
+            }
+        }
+
+        session.add_event_end(ActivityKind::Which, ExitCode::Success);
```
mikrostew: I think this should return an `ExitCode` other than success if the binary is not on the path, to mimic `which`

```
$ which doesnotexist

$ echo $?
1
```

charlespierce: Good catch, I didn't think to check the exit code of `which` when it can't find anything.

charlespierce:
Updated to use exit code 1 when `notion which` doesn't find the tool in the path. This required a small refactor to all of the commands, to allow them to return a non-zero exit code without an error result that would cause the `error:` output, matching the behavior of built-in `which`

PR #308, opened 2019-03-13
Set up CI with Azure Pipelines

dherman:


chriskrycho:
@dherman I think we can close this now?

PR #310, opened 2019-03-18
Don't print subcommand versions.

chriskrycho:
Blocked by TeXitoi/structopt#171.


dherman:
I figured out how to remove the versions from subcommands, by setting them explicitly to `""` for each one, in #318. Looks like these settings would be more concise but I'll close this PR for now and we can wait until that issue in StructOpt is fixed to clean this up.

PR #313, opened 2019-03-19
Allow version to be optional in 'notion pin'

### LINKED ISSUE #312: 'notion pin' Should default to 'latest' if no version specified.
Currently, running `notion pin node` results in an error:

```bash
error: The following required arguments were not provided:
    <version>
```

`notion pin` Should work in the same way as `notion install`: If the version isn't specified, we default to "latest" and pin that value.

charlespierce:
Fixes #312 

`notion pin` was requiring the `version` argument to be specified, instead of defaulting to "latest" like `notion install`. Update the `pin` command to match the `install` command with an Optional `version` argument and resolving the latest if the version is not specified.

### DIFF
--- a/src/command/pin.rs
+++ b/src/command/pin.rs
@@ -14,14 +14,18 @@ pub(crate) struct Pin {
     tool: String,
 
     /// The version of the tool to install, e.g. `1.2.3` or `latest`
-    version: String,
+    version: Option<String>,
 }
 
 impl Command for Pin {
     fn run(self, session: &mut Session) -> Fallible<ExitCode> {
         session.add_event_start(ActivityKind::Pin);
 
-        let version = VersionSpec::parse(&self.version)?;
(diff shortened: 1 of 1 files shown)


src/command/pin.rs
```diff
@@ -14,14 +14,18 @@ pub(crate) struct Pin {
...
-    version: String,
+    version: Option<String>,
 }
 
 impl Command for Pin {
     fn run(self, session: &mut Session) -> Fallible<ExitCode> {
         session.add_event_start(ActivityKind::Pin);
 
-        let version = VersionSpec::parse(&self.version)?;
+        let version = match self.version {
```
dherman: You could use `.map_or_default()` to eliminate boilerplate here, I think. You might need a `.transpose()` to dal with the combination of `Option` and `Result`.

PR #321, opened 2019-03-22
Package uninstall

### LINKED ISSUE #319: Support `notion uninstall` for user toolchain
Today, if you `notion install` a command, you can only `notion install` *other* versions of that command; you cannot *uninstall* it. This can be especially frustrating when tools themselves do not (yet 😄) specify a toolchain, as `notion install`-ing them will use the current user toolchain (a reasonable default), which may or may not be desired.

For example, if you have set up your global toolchain with Node 11, current versions of ember-cli warn during installation because Node 11 is not yet supported. Switching to a different user toolchain and running install again does not seem to help, because Notion correctly identifies that you already have the specified version of the tool installed!

## Related issues

- #103 
- #180

mikrostew:
This implements package uninstall: #319.

```
$ ./target/debug/notion uninstall cowsay
Removed executable 'cowthink' installed by 'cowsay'
Removed executable 'cowsay' installed by 'cowsay'
Package 'cowsay' uninstalled
```

```
$ ./target/debug/notion uninstall cowsay --full
Removed cached files for 'cowsay'
Package 'cowsay' uninstalled
```

Removes:
* [x] the json config files
* [x] the shims
* [x] the unpacked and initialized package

Optionally remove (with the `--full` flag):
* [x] the downloaded tarball and shasum files

Closes #319

### DIFF
--- a/crates/notion-core/src/distro/package.rs
+++ b/crates/notion-core/src/distro/package.rs
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 use std::ffi::OsStr;
-use std::fs::{rename, File};
+use std::fs::{self, rename, File};
 use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
@@ -14,7 +14,7 @@ use sha1::{Digest, Sha1};
 
 use crate::distro::{download_tool_error, Distro, Fetched};
 use crate::error::ErrorDetails;
-use crate::fs::{ensure_containing_dir_exists, read_dir_eager, read_file_opt};
+use crate::fs::{ensure_containing_dir_exists, files_matching, read_dir_eager, read_file_opt};
 use crate::hook::ToolHooks;
 use crate::inventory::Collection;
 use crate::manifest::Manifest;
@@ -29,6 +29,7 @@ use archive::{Archive, Tarball};
 use tempfile::tempdir_in;
 
 use notion_fail::{throw, Fallible, ResultExt};
+use regex::Regex;
 
 fn install_error(error: &io::Error) -> ErrorDetails {
     if let Some(inner_err) = error.get_ref() {
@@ -39,6 +40,12 @@ fn install_error(error: &io::Error) -> ErrorDetails {
         ErrorDetails::PackageInstallIoError {
             error: error.to_string(),
         }
+    }
+}
+
+fn file_deletion_error(err: &io::Error) -> ErrorDetails {
+    ErrorDetails::FileDeletionError {
+        error: err.to_string(),
     }
 }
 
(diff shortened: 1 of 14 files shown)


crates/notion-core/src/tool/mod.rs
```diff
@@ -55,6 +55,19 @@ impl ToolSpec {
         }
         Ok(())
     }
+
+    pub fn uninstall(&self, session: &mut Session, remove_all: bool) -> Fallible<()> {
+        match self {
+            ToolSpec::Node(_version) => unimplemented!("Uninstalling node not supported yet"),
+            ToolSpec::Yarn(_version) => unimplemented!("Uninstalling node not supported yet"),
```
arthirm: Isn't this supposed to uninstall Yarn and Npm?

crates/notion-core/src/distro/package.rs
```diff
@@ -363,6 +370,87 @@ impl PackageVersion {
...
+            }
+
+            fs::remove_file(package_config_file).with_context(file_deletion_error)?;
+        } else {
+            // there is no package config - check for orphaned binaries
+            let user_bin_dir = path::user_bin_dir()?;
+            if user_bin_dir.exists() {
+                let orphaned_bins = binaries_from_package(&name)?;
+                for bin_name in orphaned_bins {
+                    let shim = path::shim_file(&bin_name)?;
```
arthirm: Do you think we can extract this into a separate function or rearrange the code to not repeat this block?

src/command/uninstall.rs
```diff
@@ -0,0 +1,32 @@
...
+impl Command for Uninstall {
+    fn run(self, session: &mut Session) -> Fallible<ExitCode> {
+        session.add_event_start(ActivityKind::Uninstall);
+
+        let version = VersionSpec::default();
+        let tool = ToolSpec::from_str_and_version(&self.tool, version);
+
+        tool.uninstall(session)?;
+
+        session.add_event_end(ActivityKind::Uninstall, ExitCode::Success);
```
arthirm: What about session event for failure case ?

charlespierce: That's handled by the caller, if you return an `Err`. See: https://github.com/notion-cli/notion/blob/master/src/main.rs#L21

crates/notion-core/src/distro/package.rs
mikrostew: Oh good call, I meant to extract that and forgot - I'll update that

crates/notion-core/src/tool/mod.rs
mikrostew: This is only to implement #319, which is uninstall for package binaries. I created a new issue #327 to track implementing uninstall for Node and Yarn.

crates/notion-core/src/distro/package.rs
mikrostew: done!

crates/notion-core/src/distro/package.rs
```diff
@@ -363,6 +370,88 @@ impl PackageVersion {
...
+        println!("Removed executable '{}' installed by '{}'", bin_name, name);
+        Ok(())
+    }
+}
+
+/// Reads the contents of a directory and returns a Vec containing the names of
+/// all the binaries installed by the input package.
+pub fn binaries_from_package(package: &str) -> Fallible<Vec<String>> {
+    let bin_config_dir = path::user_bin_dir()?;
+    Ok(read_dir_eager(&bin_config_dir)?
```
charlespierce: This and the implementation of `files_matching` in `fs` are very similar. Would it be possible to abstract them in some way to prevent duplication of the boilerplate parts of it?

```
read_dir_eager(dir)?
    .filter(|(_, metadata)| metadata.is_file())
    .filter_map(|(entry, _)| {
    })
    .collect::<Vec<String>>())
```

Perhaps by making `files_matching` into a more general implementation that accepts a `FnOnce(PathBuf) -> Option<String>` and calls that, so we can implement each of these filters (config or regex) with only their specific logic?

crates/notion-core/src/tool/mod.rs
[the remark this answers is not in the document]
mikrostew: Good catch, copy-paste error on my part 😅

crates/notion-core/src/distro/package.rs
mikrostew: That's a good idea - I'll see if I can abstract that out

crates/notion-core/src/tool/mod.rs
arthirm: I actually meant the message. I think its fixed now :)

src/command/uninstall.rs
arthirm: Got it thanks. I was actually trying to search for it :)

crates/notion-core/src/tool/mod.rs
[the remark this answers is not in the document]
mikrostew: fixed

crates/notion-core/src/distro/package.rs
mikrostew: I was able to refactor this into a `dir_entry_match` function that takes `FnMut(DirEntry) -> Option<T>`, so it can be used by both implementations.

mikrostew:
I created #330 to track implementation of a command to remove the cached files.

PR #322, opened 2019-03-25
Support Verbose Errors and Logs

charlespierce:
Info
-----
* As part of improving errors, we want to support the `--verbose` flag to print additional error details.
* We also want to log those additional details, when we have them, regardless of the existence of the `--verbose` flag, so that users can access the more involved error details.

Changes
-----
* Added an `ErrorReporter` type, responsible for composing the error message, additional details, and error log.
* Restructured the `style` module to only be responsible for the styling of the error messages, not the composition of various pieces.
* Updated the two binaries (`main.rs` and `shim.rs`) to use `ErrorReporter` for reporting any errors that are encountered.

Tested
-----
* Added acceptance tests for the verbose flag and error logs.
* Manually tested that the verbose flag outputs as expected, and the error logs are written.

Notes
-----
* The `ErrorReporter` is configured to treat the `NOTION_DEV` environment variable the same as the `--verbose` flag.
* Currently, when an error log is written, only the file name is outputted to the console, not the full path. All of the logs are written to `NOTION_HOME/log/`, so we will need to document that fact for users, or switch to outputting the entire path (could be long, especially on windows).
* The error log is only written if there is an underlying error, so that informational errors thrown directly from within Notion don't cause an unnecessary error log to be created.

Screenshots
-----
<details>
  <summary>Concise error with underlying cause</summary>

  Concise error with underlying cause
</details>

<details>
  <summary>Verbose error with underlying cause</summary>

  Verbose error with underlying cause
</details>


src/main.rs
```diff
@@ -14,8 +14,14 @@ pub fn main() {
     session.add_event_start(ActivityKind::Notion);
 
     let notion = cli::Notion::from_args();
+    let verbose = notion.verbose;
     let exit_code = notion.run(&mut session).unwrap_or_else(|err| {
-        display_error(ErrorContext::Notion, &err);
+        let reporter = if verbose {
```
dherman: I find myself kinda wanting to see this logic get folded into a single constructor method for `ErrorReporter` so here you could write:

```rust
ErrorReporter::from_flag(notion.verbose).report(ErrorContext::Notion, &err);
```

And you could rename the existing `new` constructor `from_env()`, and `from_flag(false)` would call that.

Just a thought!

crates/notion-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "notion-core"
-version = "0.1.0"
+version = "0.2.2"
```
dherman: Is this important? I've never updated the version numbers of the internal crates since it just seems like busywork. I don't care too much either way but if we're keeping all the internal crates up to date with the main `notion` crate we ought to automate the updating as part of the release process.

crates/notion-core/src/error/reporter.rs
```diff
@@ -0,0 +1,113 @@
...
+    writeln!(log_file)?;
+    writeln!(log_file, "{}", details)?;
+
+    Ok(file_name)
+}
+
+/// Combines all the arguments into a single String
+fn collect_arguments() -> String {
+    args_os()
+        .map(|arg| arg.into_string().unwrap_or(String::from("<UNKNOWN>")))
```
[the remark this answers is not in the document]
charlespierce: Good point! I'll look into if there's a crate that can format the arguments in a correct way. If not I'll file an issue to fix it.

src/main.rs
charlespierce: That's a nice approach, I'll do that. I was trying to avoid passing a boolean in, but still having 2 constructors just with better names makes a lot more sense. My main issue was I didn't want to do `ErrorReporter::new(false)` in `shim.rs`, because magic boolean values are difficult to maintain.

crates/notion-core/Cargo.toml
charlespierce: The reason it's important is that I want to output the Notion version as part of the log file, but `env!("CARGO_PKG_VERSION")` gets the version of the current package. That said, I'm not a huge fan of having to keep the version up-to-date, so I think passing it into the constructor is fine in this case.

crates/notion-core/src/error/reporter.rs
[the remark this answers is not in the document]
charlespierce: Filed #323 to fix this

PR #326, opened 2019-03-27
List all causes in verbose errors

charlespierce:
Info
-----
* The verbose error output (and error logs) currently only shows the first level underlying cause of a given error. If the error is more deeply nested, with multiple wrapper `Error`s, then only the first will be listed.

Changes
-----
* Updated the `compose_error_details` method to traverse the entire tree of `.cause()` calls, outputting all of the causes.
* Also refactored the code a bit so `compose_error_details` returns an `Option`, instead of unwrapping the cause outside.

### DIFF
--- a/crates/notion-core/src/error/reporter.rs
+++ b/crates/notion-core/src/error/reporter.rs
@@ -7,7 +7,7 @@ use crate::fs::ensure_containing_dir_exists;
 use crate::path::log_dir;
 use crate::style::{format_error_cause, format_error_message};
 use chrono::Local;
-use failure::{Error, Fail};
+use failure::Error;
 use notion_fail::NotionError;
 
 const NOTION_DEV: &'static str = "NOTION_DEV";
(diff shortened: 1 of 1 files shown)


crates/notion-core/src/error/reporter.rs
```diff
@@ -98,8 +95,31 @@ impl ErrorReporter {
...
+        None => {
+            return None;
+        }
+    };
+    let mut details = String::new();
+
+    // Walk up the tree of causes and include all of them
+    loop {
+        write!(details, "{}", format_error_cause(current))
+            .expect("write! to a String doesn't fail");
```
mikrostew: You can avoid the `write!` and `expect()` here by using string concatenation:

```rust
details += &format_error_cause(current);
```

Or you could use `push_str`

```rust
details.push_str(&format_error_cause(current));
```

crates/notion-core/src/error/reporter.rs
```diff
@@ -98,8 +95,31 @@ impl ErrorReporter {
...
+
+        match current.cause() {
+            Some(cause) => {
+                write!(details, "\n\n").expect("write! to a String doesn't fail");
```
mikrostew: Same thing here

crates/notion-core/src/error/reporter.rs
charlespierce: Good call! I was so caught up thinking about formatting that I completely missed that I didn't need the power of the `format!` macros when appending a string value. Updated!

crates/notion-core/src/error/reporter.rs
charlespierce: Updated

Issue #295, opened 2019-03-08
Audit existing error messages to improve helpfulness


charlespierce:
As part of improving error messages #294, we need to audit the existing error messages and improve their helpfulness. Currently, many of them are simply writing the underlying error message out to the console, which is often useless to the end user. We should add the context of what Notion was trying to do when the error occurred, and if possible, what the user can do to fix the problem on their own.

PR #328, opened 2019-03-28
First pass at auditing existing error messages

charlespierce:
See #295 

Info
-----
* With the ability to provide verbose error messages in place, we can start to actually tackle improving the error messages.
* To start, we want to audit the existing error messages and update them to be more helpful.
* There are many error messages in the program currently, so to start I focused on two areas:
    1. Informational error messages thrown by the Tool shims, indicating a problem with the users' configuration (platform or project).
    2. Errors that included the text of an underlying error inside of them (via an `error` property on the `ErrorDetails` variant).

Changes
-----
* Updated error messages in `error/details.rs` to match the new style, including CTAs to help guide the user to the correct action.
* Updated all call sites to correctly reference the new `ErrorDetails` variants, providing the correct context to the error message.
* Minor refactors in functionality to provide more granular errors.
* Updated tests to reflect the new error output.

### DIFF
--- a/crates/notion-core/src/inventory/mod.rs
+++ b/crates/notion-core/src/inventory/mod.rs
@@ -184,10 +184,13 @@ pub trait FetchResolve<D: Distro> {
     ) -> Fallible<D::ResolvedVersion>;
 }
 
-fn registry_fetch_error(error: &reqwest::Error) -> ErrorDetails {
-    ErrorDetails::RegistryFetchError {
-        error: error.to_string(),
-    }
+fn registry_fetch_error(
+    tool: impl AsRef<str>,
+    from_url: impl AsRef<str>,
+) -> impl FnOnce(&reqwest::Error) -> ErrorDetails {
+    let tool = tool.as_ref().to_string();
+    let from_url = from_url.as_ref().to_string();
+    |_| ErrorDetails::RegistryFetchError { tool, from_url }
 }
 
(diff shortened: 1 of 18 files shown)


crates/notion-core/src/inventory/mod.rs
```diff
@@ -496,9 +502,9 @@ impl FetchResolve<PackageDistro> for PackageCollection {
         if let Some(entry) = entry_opt {
             Ok(entry)
         } else {
-            throw!(ErrorDetails::NoPackageFound {
+            throw!(ErrorDetails::PackageVersionNotFound {
                 name: name.to_string(),
-                matching: VersionSpec::Latest,
+                matching: matching.to_string(),
```
mikrostew: good catch!

crates/notion-core/src/inventory/mod.rs
```diff
@@ -527,9 +533,9 @@ impl FetchResolve<PackageDistro> for PackageCollection {
         if let Some(entry) = entry_opt {
             Ok(entry)
         } else {
-            throw!(ErrorDetails::NoPackageFound {
+            throw!(ErrorDetails::PackageVersionNotFound {
                 name: name.to_string(),
-                matching: VersionSpec::Latest,
+                matching: exact_version.to_string(),
```
mikrostew: and here 👍

PR #331, opened 2019-03-29
Support `notion install node lts`

### LINKED ISSUE #255: Support `notion install node lts`
Looks like an awesome tool, thanks!

One nice thing `nvm` gives you is the ability to default to LTS versions of Node. Most people probably want to be on latest LTS by default, so it would maybe even make sense to make `notion install node` install the latest LTS by default.

In any case an opt-in flag would be useful.

nvm docs link: https://github.com/creationix/nvm#long-term-support

willsoto:
Closes #255

### DIFF
--- a/crates/notion-core/src/inventory/mod.rs
+++ b/crates/notion-core/src/inventory/mod.rs
@@ -149,6 +149,7 @@ pub trait FetchResolve<D: Distro> {
     ) -> Fallible<D> {
         let version = match *matching {
             VersionSpec::Latest => self.resolve_latest(name.clone(), hooks)?,
+            VersionSpec::Lts => self.resolve_lts(name.clone(), hooks)?,
             VersionSpec::Semver(ref requirement) => {
                 self.resolve_semver(name.clone(), requirement, hooks)?
             }
@@ -174,6 +175,15 @@ pub trait FetchResolve<D: Distro> {
         matching: &VersionReq,
         hooks: Option<&ToolHooks<D>>,
     ) -> Fallible<D::ResolvedVersion>;
+
+    /// Resolves an LTS version for this tool
+    fn resolve_lts(
+        &self,
+        name: String,
+        hooks: Option<&ToolHooks<D>>,
+    ) -> Fallible<D::ResolvedVersion> {
+        return self.resolve_latest(name, hooks);
+    }
 
     /// Resolves an exact version of this tool
     fn resolve_exact(
@@ -273,6 +283,29 @@ impl FetchResolve<NodeDistro> for NodeCollection {
         } else {
             throw!(ErrorDetails::NodeVersionNotFound {
                 matching: matching.to_string()
+            })
+        }
+    }
+
+    fn resolve_lts(
+        &self,
+        _name: String,
+        hooks: Option<&ToolHooks<NodeDistro>>,
+    ) -> Fallible<Version> {
+        let url = match hooks {
+            Some(&ToolHooks {
+                index: Some(ref hook),
+                ..
+            }) => hook.resolve("index.json")?,
+            _ => public_node_version_index(),
+        };
+        let version_opt = match_node_version(&url, |&NodeEntry { lts: val, .. }| val)?;
+
+        if let Some(version) = version_opt {
+            Ok(version)
+        } else {
+            throw!(ErrorDetails::NodeVersionNotFound {
+                matching: "lts".to_string()
             })
         }
     }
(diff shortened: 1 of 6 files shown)


crates/notion-core/src/inventory/mod.rs
```diff
@@ -277,6 +287,29 @@ impl FetchResolve<NodeDistro> for NodeCollection {
...
+        hooks: Option<&ToolHooks<NodeDistro>>,
+    ) -> Fallible<Version> {
+        let url = match hooks {
+            Some(&ToolHooks {
+                index: Some(ref hook),
+                ..
+            }) => hook.resolve("index.json")?,
+            _ => public_node_version_index(),
+        };
+        let version_opt = match_node_version(&url, |&NodeEntry { lts: ref v, .. }| v.clone())?;
```
charlespierce: `lts` is a Boolean value, so I don't think we need to use the `ref v` and `v.clone()`, instead I think this should be doable as just:
`|&NodeEntry { lts: val, .. }| v`

tests/smoke/notion_install.rs
```diff
@@ -21,6 +21,18 @@ fn install_node() {
...
 
+#[test]
+fn install_node_lts() {
+    let p = temp_project().build();
+
+    assert_that!(p.notion("install node lts"), execs().with_status(0));
+
+    assert_that!(
+        p.node("--version"),
+        execs().with_status(0).with_stdout_contains("1.2.1")
```
charlespierce: Since the LTS version of Node will constantly change, we probably can't check for a specific version here. However, checking that both the `notion install node lts` command completed successfully and then that `node --version` completes successfully should be enough of a smoke test that the install is working.

crates/notion-core/src/inventory/serial.rs
```diff
@@ -108,6 +108,7 @@ pub struct NodeEntry {
     #[serde(with = "option_version_serde")]
     pub npm: Option<Version>,
     pub files: Vec<String>,
+    pub lts: bool,
```
charlespierce: It looks like the issue with the test is here: I didn't look as closely as I should have at the `index.json` for Node. It looks like the `lts` property in the Node index is actually of type `false | string`. If it's false, that means no LTS, if it's a string value (like "Dubnium"), then it's that named LTS group (Node 10 is `Dubnium`, Node 8 is `Carbon`, etc.)

It looks like we'll need to have a custom deserializer for that field, that can convert a string to a boolean value. You can take a look at [this block](https://github.com/notion-cli/notion/blob/master/crates/notion-core/src/version/mod.rs#L152) for implementing a custom deserializer. In our case, we probably want to try to deserialize the value as a string, and if that succeeds we return `Ok(true)`, otherwise `Ok(false)`. Then we can annotate this line with `#[serde(deserialize_with = "<NAME OF FUNCTION>")]`, and it should be able to parse it the way we want.

tests/smoke/notion_install.rs
willsoto: 😅 absolutely. I was putting a bad version in there so I could actually test that my changes were doing something but ended up getting a totally different error! Thanks for the suggestion on how to test this though 👍

crates/notion-core/src/inventory/serial.rs
willsoto: I've gone ahead and implemented my best attempt at this, please let me know what you think!

crates/notion-core/src/version/mod.rs
```diff
@@ -163,3 +167,17 @@ pub mod option_version_serde {
...
 }
+
+pub mod lts_version_serde {
+    use serde::{self, Deserialize, Deserializer};
+
+    pub fn deserialize<'de, D>(deserializer: D) -> Result<bool, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        match String::deserialize(deserializer) {
```
charlespierce: This looks perfect! The only comment I have is that I think this fits better with the parsing logic in `inventory/serial.rs`, as opposed to here with the version parsing. I believe it can also be a non-`pub` function in that module that is referenced directly in `deserialize_with`, instead of needing its own module.

willsoto:
Everything should be good now, thank you for the feedback! I moved `fn lts_version_serde` to the top of `inventory/serial.rs` near the only other free floating function I found in that file. I hope that is okay.

---

Unrelated but I noticed after I rebased the last time I got an error on this line telling me that `PackageConfigNotFound` is not found. Doesn't seem to be breaking anything, but wanted to let you know anyway.

https://github.com/notion-cli/notion/blob/d0fab179f19cad2cc8f791bbe86cc9c25c157be0/crates/notion-core/src/inventory/serial.rs#L255

PR #336, opened 2019-04-07
Finish auditing the existing error messages

### LINKED ISSUE #295: Audit existing error messages to improve helpfulness

charlespierce:
Closes #295 

Info
-----
* Builds on #328 
* Completes audit and update of existing error messages.
* All error messages in `error/details.rs` should now follow the new style or have a comment explaining why they don't have a CTA.

Changes
-----
* Updated all remaining errors in `error/details.rs` to reflect the new style, mostly by adding CTAs.
* Updated call sites where necessary to pass the correct context to the errors.
* Fixed tests to reflect the new error messages.


crates/notion-core/src/error/details.rs
```diff
@@ -343,10 +372,13 @@ Please ensure you have correct permissions to the Notion directory."#, name),
...
-            ErrorDetails::ToolNotImplemented => write!(f, "this tool is not yet implemented"),
-            ErrorDetails::UnrecognizedShell { name } => write!(f, "Unrecognized shell: {}", name),
+            ErrorDetails::UnrecognizedShell { name } => write!(f, "Unrecognized shell '{}'
+
+Please ensure you are using a supported shell.", name),
             ErrorDetails::UnspecifiedPostscript => {
-                write!(f, "Notion postscript file not specified")
+                write!(f, "Could not determine Notion postscript file.
+
+Please nesure Notion was installed correctly.")
```
chriskrycho: Typo: "nesure" → "ensure"

crates/notion-core/src/error/details.rs
```diff
@@ -184,26 +186,39 @@ impl fmt::Display for ErrorDetails {
...
-                version,
+                existing_package,
+                new_package,
             } => write!(
                 f,
-                "Conflict with bin '{}' already installed by '{}' version {}",
-                bin_name, package, version
+                "Executable '{}' is already installed by {}
+
+Remove {} before installing {}",
```
chriskrycho: For consistency with the other CTAs, this should probably be `Please remove`.

crates/notion-core/src/error/details.rs
charlespierce: Good catch, fixed!

crates/notion-core/src/error/details.rs
charlespierce: Very true, updated.

PR #337, opened 2019-04-08
Rework unknown errors in event module

### LINKED ISSUE #297: Update calls to `.unknown()` in `event` module

Update all of the calls to `.unknown()` in the `event` module (As of this writing there is 1 such call) to give better error messages for each.

charlespierce:
Closes #297 

Update the tests in the `event` module to not use the `.unknown()` method and instead instantiate an `ErrorDetails` object.


chriskrycho:
:shipit:

PR #340, opened 2019-04-08
Rework unknown errors in shell module

### LINKED ISSUE #304: Update calls to `.unknown()` in `shell` module

Update all of the calls to `.unknown()` in the `shell` module (As of this writing there are 2 such calls) to give better error messages for each.

charlespierce:
Closes #304 

Add a `CreatePostscriptError` to represent a failure to create the postscript file when executing `notion activate` or `notion deactivate`

### DIFF
--- a/crates/notion-core/src/error/details.rs
+++ b/crates/notion-core/src/error/details.rs
@@ -44,7 +44,9 @@ pub enum ErrorDetails {
     },
 
     /// Thrown when unable to create the postscript file
-    CreatePostscriptError,
+    CreatePostscriptError {
+        in_dir: String,
+    },
 
     CurrentDirError,
 
@@ -237,9 +239,10 @@ Please ensure you have correct permissions to the Notion directory.", path)
 
 Please ensure that you have the correct permissions.", dir)
             }
-            ErrorDetails::CreatePostscriptError => write!(f, "Could not create postscript file.
+            ErrorDetails::CreatePostscriptError { in_dir } => write!(f, "Could not create postscript file
+in {}
 
-Please ensure you have correct permissions to the Notion directory."),
+Please ensure you have correct permissions to the Notion directory.", in_dir),
             ErrorDetails::CurrentDirError => write!(f, "Could not determine current directory
 
 Please ensure that you have the correct permissions."),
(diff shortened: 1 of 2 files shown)


crates/notion-core/src/error/details.rs
```diff
@@ -224,6 +227,9 @@ Please ensure you have correct permissions to the Notion directory.", path)
 
 Please ensure that you have the correct permissions.", dir)
             }
+            ErrorDetails::CreatePostscriptError => write!(f, "Could not create postscript file.
+
+Please ensure you have correct permissions to the Notion directory."),
```
chriskrycho: One thing we could do throughout where we have to reference specific directories is, if possible, make the enum variant smart enough to pass that along, so that in cases like this we could do this:

```rust
 ErrorDetails::CreatePostscriptError { for_directory } => {
    write!(
        f,
        "Could not create postscript file.

Please ensure you have correct permissions to the Notion directory (`{}`).",
        for_directory
    )
},
```

I don't think we should block landing this on that, because we probably need to survey and fill that kind of info in *throughout*, but it caught my attention. (I'll open an issue to that effect to track it if you like!)

charlespierce: I actually tried to do that most of the time when referencing a specific file / directory (c.f. https://github.com/notion-cli/notion/blob/177322a81a37cd3c8e163b172dbc18c0a7b876dc/crates/notion-core/src/error/details.rs#L240)

In this case (as well as the `CreateTempDirError` and `CreateTempFileError`), however, I didn't include them because the postscript file (and temp dir / file) are randomly named, so it would be somewhat gibberish.

I didn't consider the idea of showing the containing directory, that's a great idea! Instead of showing

```
Could not create postscript file
at ~/.notion/tmp/gibberishlakshdf
```

We could just show:

```
Could not create postscript file
in ~/.notion/tmp
```

Which lets the user know what directory they need to check, without overloading them with extra irrelevant information about the specifics of a temporary file name. I'll update this and #344 to include those!

chriskrycho: Ah, that makes sense – and I noticed you'd done that when looking at others after this one. 👍

PR #342, opened 2019-04-08
Rework unknown errors in manifest module

### LINKED ISSUE #301: Update calls to `.unknown()` in `manifest` module

Update all of the calls to `.unknown()` in the `manifest` module (As of this writing there are 8 such calls) to give better error messages for each.

charlespierce:
Closes #301 

Update all of the calls to `.unknown()` in `update_toolchain` to use `.with_context()` instead, returning `ErrorDetails` variants. Also Added `PackageWriteError` and `StringifyToolchainError` to cover the error cases that had not already been covered.

### DIFF
--- a/crates/notion-core/src/manifest/mod.rs
+++ b/crates/notion-core/src/manifest/mod.rs
@@ -91,18 +91,18 @@ impl Manifest {
         toolchain: serial::ToolchainSpec,
         package_file: PathBuf,
     ) -> Fallible<()> {
+        // Helper for lazily creating the file name string without moving `package_file` into
+        // one of the individual `with_context` closures below.
         let get_file = || package_file.to_string_lossy().to_string();
 
         // parse the entire package.json file into a Value
-        let file = File::open(&package_file)
+        let contents = read_to_string(&package_file)
             .with_context(|_| ErrorDetails::PackageReadError { file: get_file() })?;
-        let mut v: serde_json::Value = serde_json::from_reader(file)
+        let mut v: serde_json::Value = serde_json::from_str(&contents)
             .with_context(|_| ErrorDetails::PackageParseError { file: get_file() })?;
 
         if let Some(map) = v.as_object_mut() {
             // detect indentation in package.json
-            let contents = read_to_string(&package_file)
-                .with_context(|_| ErrorDetails::PackageReadError { file: get_file() })?;
             let indent = detect_indent::detect_indent(&contents);
 
             // update the "toolchain" key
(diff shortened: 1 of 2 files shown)


crates/notion-core/src/manifest/mod.rs
```diff
@@ -92,27 +91,33 @@ impl Manifest {
         toolchain: serial::ToolchainSpec,
         package_file: PathBuf,
     ) -> Fallible<()> {
-        // parse the entire package.json file into a Value
-        let file = File::open(&package_file).unknown()?;
-        let mut v: serde_json::Value = serde_json::from_reader(file).unknown()?;
+        let get_file = || package_file.to_string_lossy().to_string();
```
chriskrycho: A comment here explaining that we do it this way to allow easy repeated creation of the file path without moving it into the *other* closures below would be super helpful for future maintainers. It took me a minute of poking at it to figure out why we needed this.

crates/notion-core/src/manifest/mod.rs
```diff
@@ -92,27 +91,33 @@ impl Manifest {
...
 
-        // detect indentation in package.json
-        let mut contents = String::new();
-        let mut indent_file = File::open(&package_file).unknown()?;
-        indent_file.read_to_string(&mut contents).unknown()?;
-        let indent = detect_indent::detect_indent(&contents);
+        // parse the entire package.json file into a Value
+        let file = File::open(&package_file)
```
chriskrycho: We can simplify this flow *and* reduce the number of I/O operations and allocations all in a go, because we're *basically* always going to have to allocate this string in any successful case, and in the error case, serde would have been doing similar under the hood anyway:

```diff
-       let file = File::open(&package_file)
+       let contents = read_to_string(&package_file)
            .with_context(|_| ErrorDetails::PackageReadError { file: get_file() })?;
-       let mut v: serde_json::Value = serde_json::from_reader(file)
+       let mut v: serde_json::Value = serde_json::from_str(&contents)
            .with_context(|_| ErrorDetails::PackageParseError { file: get_file() })?;

        if let Some(map) = v.as_object_mut() {
            // detect indentation in package.json
-           let contents = read_to_string(&package_file)
-               .with_context(|_| ErrorDetails::PackageReadError { file: get_file() })?;
            let indent = detect_indent::detect_indent(&contents);
```

crates/notion-core/src/manifest/mod.rs
charlespierce: Great idea, updated!

crates/notion-core/src/manifest/mod.rs
charlespierce: Good catch, that is a lot simpler

PR #368, opened 2019-04-19
Fix annoying package install errors

### LINKED ISSUE #352: `notion install` fails if the same version has already been installed
After running `notion install <package>` once, if it is run again and the same version is found, the install will show an error on this line: https://github.com/notion-cli/notion/blob/cd37603b3f78e5bfbf4af7a40e52e7ced1463b0d/crates/notion-core/src/distro/package.rs#L168

Since the image directory is not empty. We should detect that the requested version is already installed and not attempt to re-install the same package. Related to #288

### LINKED ISSUE #353: `notion install` fails for new version of existing package
If you attempt to run `notion install <package>` for a package that you already installed, but that has a new version, the install fails with the error:

```bash
$ notion install ember-cli
error: Executable 'ember' is already installed by ember-cli

Please remove ember-cli before installing ember-cli
```

We should detect that the package is the same as the overlapping one and overwrite the existing install with the new version, instead of requiring an uninstall + reinstall.

### LINKED ISSUE #371: Interrupting package install leaves the filesystem in a bad state
Found this working on a slow WiFi this afternoon. This is similar to #200, but only for packages.

Interrupting the package install leaves the filesystem in a state where that version of the package is not able to be installed:

```
$ notion install cowsay 1.4.0
  Unpacking cowsay-v1.4.0  [========================================] 100%
[hit ^C here to interrupt the npm install]

$ notion install cowsay 1.4.0
  Unpacking cowsay-v1.4.0  [========================================] 100%
error: an internal error occurred

Notion is still a pre-alpha project, so we expect to run into some bugs,
but we'd love to hear about them so we can fix them!

Please feel free to reach out to us at @notionjs on Twitter or file an issue at:

    https://github.com/notion-cli/notion/issues

Error details written to: notion-error-2019-04-19_15_28_12.293.log

$ notion install cowsay 1.3.0
  Unpacking cowsay-v1.3.0  [========================================] 100%
npm notice created a lockfile as package-lock.json. You should commit this file.
added 9 packages from 2 contributors and audited 1535 packages in 4.964s
found 0 vulnerabilities
```

mikrostew:
## Bug Fixes

Closes #352 
Now this doesn't error if the same version has already been installed:

```
$ notion install cowsay
  Unpacking cowsay-v1.4.0  [========================================] 100%
npm notice created a lockfile as package-lock.json. You should commit this file.
added 9 packages from 2 contributors and audited 1535 packages in 3.261s
found 0 vulnerabilities

$ notion install cowsay
Package `cowsay` is up-to-date, version 1.4.0 already installed
```

Closes #353 
Don't fail for a different version of an existing package:

```
$ notion install ember-cli 3.8.1
  Fetching ember-cli-v3.8.1  [========================================] 100%
[lots out output not included]

$ ember --version
WARNING: Node v11.14.0 is not tested against Ember CLI on your platform. We recommend that you use the most-recent "Active LTS" version of Node.js. See https://git.io/v7S5n for details.
ember-cli: 3.8.1
node: 11.14.0
os: darwin x64

$ notion install ember-cli 3.9
  Fetching ember-cli-v3.9.0  [========================================] 100%
[lots log output not included]

$ ember --version
ember-cli: 3.9.0
node: 11.14.0
os: darwin x64
```

Closes #371
Now this doesn't fail to install if the previous install was cancelled:

```
$ notion install cowsay
  Unpacking cowsay-v1.4.0  [========================================] 100%
^C░░░░░░░░░░░░░░░░░⸩ ⠧ fetchMetadata: sill pacote range manifest for istanbul-reports@^2.2.2 fetched in 146ms
[I hit ^C to cancel the npm install]

$ notion install cowsay
  Unpacking cowsay-v1.4.0  [========================================] 100%
npm notice created a lockfile as package-lock.json. You should commit this file.
added 9 packages from 2 contributors and audited 1535 packages in 4.717s
found 0 vulnerabilities
```

## Other Stuff

* Cleaned up some descriptions in the comments for `distro::Fetched` to be more accurate.

### DIFF
--- a/crates/notion-core/src/session.rs
+++ b/crates/notion-core/src/session.rs
@@ -207,7 +207,6 @@ impl Session {
         let package_version = fetched_package.version();
 
         // if the package is already installed, don't re-install it
-        // TODO: accept --force for this?
         if let Fetched::Installed(pkg_version) = fetched_package {
             let version = pkg_version.version.clone();
             println!(
--- a/crates/notion-core/src/session.rs
+++ b/crates/notion-core/src/session.rs
@@ -149,11 +149,9 @@ impl Session {
 
         if !inventory.node.contains(version) {
             let hooks = self.hooks.get()?;
-            inventory.node.fetch(
-                "node".to_string(),
-                &VersionSpec::exact(version),
-                hooks.node.as_ref(),
-            )?;
+            inventory
+                .node
+                .fetch("node", &VersionSpec::exact(version), hooks.node.as_ref())?;
         }
 
         Ok(())
@@ -165,11 +163,9 @@ impl Session {
 
         if !inventory.yarn.contains(version) {
             let hooks = self.hooks.get()?;
-            inventory.yarn.fetch(
-                "yarn".to_string(),
-                &VersionSpec::exact(version),
-                hooks.yarn.as_ref(),
-            )?;
+            inventory
+                .yarn
+                .fetch("yarn", &VersionSpec::exact(version), hooks.yarn.as_ref())?;
         }
 
         Ok(())
@@ -203,7 +199,7 @@ impl Session {
     /// Fetch, unpack, and install a package matching the input requirements.
     pub fn install_package(&mut self, name: String, version: &VersionSpec) -> Fallible<Version> {
         // fetches and unpacks package
-        let fetched_package = self.fetch_package(name.clone(), version)?;
+        let fetched_package = self.fetch_package(&name, version)?;
         let package_version = fetched_package.version();
 
         // if the package is already installed, don't re-install it
@@ -251,7 +247,7 @@ impl Session {
         let hooks = self.hooks.get()?;
         inventory
             .node
-            .fetch("node".to_string(), &version_spec, hooks.node.as_ref())
+            .fetch("node", &version_spec, hooks.node.as_ref())
     }
 
     /// Fetches a Yarn version matching the specified semantic versioning requirements.
@@ -260,7 +256,7 @@ impl Session {
         let hooks = self.hooks.get()?;
         inventory
             .yarn
-            .fetch("yarn".to_string(), &version_spec, hooks.yarn.as_ref())
+            .fetch("yarn", &version_spec, hooks.yarn.as_ref())
     }
 
     /// Fetches a Npm version matching the specified semantic versioning requirements.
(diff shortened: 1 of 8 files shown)


crates/notion-core/src/session.rs
```diff
@@ -203,9 +203,19 @@ impl Session {
...
     pub fn install_package(&mut self, name: String, version: &VersionSpec) -> Fallible<Version> {
         // fetches and unpacks package
-        let fetched_package = self.fetch_package(name, version)?;
+        let fetched_package = self.fetch_package(name.clone(), version)?;
         let package_version = fetched_package.version();
 
+        // if the package is already installed, don't re-install it
+        if let Fetched::Installed(pkg_version) = fetched_package {
+            let version = pkg_version.version.clone();
+            eprintln!(
```
charlespierce: I think we can write this message to `stdout` instead of `stderr`, since it's an informative message about the fact that we didn't need to do anything.

```suggestion
            println!(
```

crates/notion-core/src/session.rs
```diff
@@ -203,9 +203,19 @@ impl Session {
     /// Fetch, unpack, and install a package matching the input requirements.
     pub fn install_package(&mut self, name: String, version: &VersionSpec) -> Fallible<Version> {
         // fetches and unpacks package
-        let fetched_package = self.fetch_package(name, version)?;
+        let fetched_package = self.fetch_package(name.clone(), version)?;
```
charlespierce: I'm not sure of the exact way to resolve it (if there even is one), but it seems like there are a lot of calls to `clone` in this method. In my (admittedly limited) experience with Rust, that tends to point to the data model not being ideal. It's not a big issue or a blocker for this PR, especially since the objects we're cloning here are small (versions and Strings), but just something that stuck out to me.

Perhaps @dherman or @chriskrycho can weigh in, having significantly more Rust knowledge?

chriskrycho: Two notes:

1. We can almost certainly plumb references down through as need be to make it work to just use references here (though I'd file that as a tech debt issue)—it should be fairly mechanical, and in fact I don't see (on an initial perusal) anywhere we *can't* just use an `&str` instead that we're passing down `String`s through this chain currently.

2. It's true that it's nice (and generally preferable) to avoid `clone`-ing unnecessarily, but that's often best left for cleanup/optimization phases. We'll use *slightly* more memory by allocating copies of this `String`, but it's going to be a pretty small amount and is very unlikely to be especially meaningful in the use of the tool. It's a good thing to do, but it's context-dependent how worthwhile it is to invest there.

chriskrycho: I just checked… and the Rust compiler is so ridiculously good; it was indeed trivial to plumb this through. [Opened a PR against this branch](https://github.com/mikrostew/notion/pull/1) that you can merge or ignore as you like, @mikrostew!

chriskrycho: I, uhh, will neither confirm nor deny reports that I got carried away and removed *a bunch* of moves and allocations on that branch/PR. 😛

crates/notion-core/src/session.rs
mikrostew: Sounds good to me

crates/notion-core/src/session.rs
mikrostew: Thanks @chriskrycho, just merged your PR. That cleans this up nicely

chriskrycho:
Rebase/merge and then :shipit:!

mikrostew:
rebased!

PR #370, opened 2019-04-19
Rename Notion to Jetson.

chriskrycho:
We can land this as soon as we're ready with the org move – as it stands, it updates all the links as well as the names in the README. I did not touch RELEASES.


.gitignore
```diff
@@ -1,9 +1,12 @@
 /target/
 **/*.rs.bk
 Notion.msi
```
mikrostew: Remove the `Notion.*` ignored files here

mikrostew:
And this will need to be re-visited before merging, because some open PRs add Notion text, e.g. https://github.com/notion-cli/notion/pull/344/files

.gitignore
chriskrycho: Good catch—but I actually left this in place on purpose so that if someone happens to have them hanging around after a rename locally, they don't accidentally check them in!

chriskrycho:
Yep, I was planning to do a rebase once we landed everything else and once @dherman is ready with the other bits. I'm just going to rebase rebase rebase till it's a green 🚥.

.gitignore
mikrostew: ah gotcha

dherman:
@chriskrycho Are you blocked on me for anything else in this PR?

chriskrycho:
Nope, just waiting till we land the Windows I staled PR. Once we do that, I’ll rebase one last time and we can merge it.

chriskrycho:
Closing; will recreate the same thing when we have a new name. 😢

Issue #362, opened 2019-04-17
--verbose does not seem to work


thoov:
Running `notion --verbose install node 8` does not include any information.

I would assume that it would print out things like:

- The version that got resolved (ie 8 -> 8.11.2) 
- The url (and status code) of where notion is attempting to download from
- Where is notion placing node

charlespierce:
`--verbose` Currently only affects error reporting, writing additional details in the case of an error. We've talked about including other reporting, but haven't implemented anything else so far. Thanks for the examples of info you'd like to see / would be useful when doing `--verbose`

thoov:
@charlespierce When running `notion --verbose pin node 10` with an invalid `package.json` I get:

```
error: Could not parse project manifest
at /private/tmp/foobaz/package.json

Please ensure that the file is correctly formatted.

cause: expected `,` or `}` at line 53 column 5
Error details written to: notion-error-2019-04-17_10_34_56.719.log
```

It seems like the only thing that --verbose added was the "cause" line. That should be in the default error message and not behind the flag. To me conceptually `--verbose` should tell the user **what** notion is doing "behind the scenes" and not preventing users from actionable information.

cc: @dherman

charlespierce:
That's true, that's what verbose errors does is show the underlying cause (and further, if there are nested causes). There are a couple of issues with showing all of the root causes by default:

* It makes the error messages longer and less concise. We explicitly wanted the errors to be concise to prevent the explosion of spam you see in e.g. `npm` when something goes wrong. We have the flag to allow that extra info to be shown, but by default want to keep things to-the-point.
* We don't have any control over the underlying error messages, they are generated by library dependencies, so their usefulness is questionable and not consistent. For example, one of the biggest complaints we got before the error message revamp was about how useless it was to see `not found (os error 2)`. That is one of the underlying error messages that we don't want to surface all the time, because it's just meaningless noise.

In the case of parsing errors specifically, I agree that providing some context of where the problem was makes sense, and we have an issue to track that: #280 Given that the parsing library we're using seems to give useful messages, it's likely that we can just include them.

PR #372, opened 2019-04-19
Allow Tool executions to pass-through to existing PATH if no platform found

### LINKED ISSUE #64: resolve to a default version for each toolchain if already exists
I installed `notion` on my mac  🎉🚀 and got so excited 💃🕺so that the first thing I did after that was to run `node -v` in the same directory and I got this error: `error: An unknown error has occurred`.
I was assuming that if there is no `package.json` or `notion` config block then `notion` will either fall back to a default version of `node` or will respect my previously installed version of `node`. By that I mean, when the notion is installed on a machine it can detect and install the global version of my installed toolchain (`node`, `yarn` ...) and set them as default versions.

If the expected behavior is not as I assumed, then maybe another option would be showing some instructions/helpful errors instead of `error: An unknown error has occurred`. I'll be happy to help fixing it as well :)

### LINKED ISSUE #345: Pass-through to already installed tools if no toolchain available
If executing a tool and there is no toolchain available, we should attempt to call the tool in the user's existing `PATH`. We should, however, store the "No Toolchain" error, so that if executing the command fails (i.e. if the user didn't have the tool available already), then we can still show the Notion error.

That way we get the benefits of error messages that tell the users how to get something working within Notion, but also the ability to seamlessly work with what the user already had installed when they installed Notion.

charlespierce:
Closes #345 
Closes #64 

Info
-----
* If no toolchain is found for a given command that we are shimming, we should default to passing through to the rest of the user's `PATH`.
* This will allow, for example, users who have `node` already installed to install Jetson and continue using their global `node` before they select a version of `node` with Jetson.
* We do, however, want to keep the error message, in case we are unable to find the command we are passing through, so that we can still show the Jetson CTAs about how to install a version of Node.

Changes
-----
* Refactored the tool-launching code to remove the `Tool` trait, instead creating a single enum `ToolCommand`, which has variants for `Direct` and `Passthrough`.
* Replaced the individual tool implementations with `node_command` functions, that perform the same checks that the `Tool::new` implementations were doing before.
* When creating a command would have returned a `NoPlatform` or `BinaryNotFound` error, it instead returns a `Passthrough` command, with that error as the default error.

Tested
-----
* Updated tests that called `npm` and `yarn` to fix their expected behaviors.
* Manually tested that the commands are passed through to the existing path if Jetson can't find the platform.


crates/notion-core/src/tool/npx.rs
```diff
@@ -1,50 +1,35 @@
...
             // npx was only included with npm 5.2.0 and higher. If the npm version is less than that, we
             // should include a helpful error message
             let required_npm = VersionSpec::parse_version("5.2.0")?;
             if image.node.npm >= required_npm {
-                Ok(Self::from_components(
-                    OsStr::new("npx"),
-                    args,
-                    &image.path()?,
-                ))
+                let path = image.path()?;
```
chriskrycho: Since we're doing this here, do you mind making the same change in the other tool invocations where you didn't already? I was going to suggest it as a slightly clarity improvement anyway.

crates/notion-core/src/tool/binary.rs
```diff
@@ -1,83 +1,70 @@
...
 
-        // try to use the user toolchain
-        if let Some(user_tool) = session.get_user_tool(&params.executable)? {
-            return Ok(Self::from_components(
-                &user_tool.bin_path.as_os_str(),
-                params.args,
-                &user_tool.image.path()?,
-            ));
+            // if there's no user platform selected, pass through to existing PATH.
+            return ToolCommand::passthrough(&exe, args, ErrorDetails::NoPlatform);
```
mikrostew: I think this return should be removed, so the code falls through to looking for a user tool (instead of trying to pass through to the system at this point).

In the following scenario:
* the user installs Jetson
* the user runs `notion install tsc`, for instance

At that point, `tsc` is installed as a user tool, but there is no default version of Node set for the user (because package install doesn't set the default, it just makes sure there is a platform downloaded).

Then when the user runs `tsc <whatever>` in the project, it will fail, because it doesn't get far enough to look for the user tool.

crates/notion-core/src/tool/binary.rs
```diff
@@ -1,83 +1,70 @@
...
+use super::ToolCommand;
 use crate::error::ErrorDetails;
 use crate::session::{ActivityKind, Session};
 
-use notion_fail::{throw, Fallible};
+use notion_fail::Fallible;
 
-/// Represents a delegated binary executable.
-pub struct Binary(Command);
+pub(super) fn binary_command<A>(
```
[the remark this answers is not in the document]
charlespierce: Good idea, I like that consistency

crates/notion-core/src/tool/mod.rs
```diff
@@ -97,44 +97,51 @@ pub fn execute_tool(session: &mut Session) -> Fallible<ExitStatus> {
...
-                executable: exe,
-                args,
-            },
-            session,
-        )?
-        .exec(),
-    }
-}
+    let command = match &exe.to_str() {
+        Some("node") => node_command(args, session)?,
```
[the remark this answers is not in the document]
charlespierce: Done

crates/notion-core/src/tool/binary.rs
```diff
@@ -1,83 +1,70 @@
...
 
-    fn command(self) -> Command {
-        self.0
-    }
+    // at this point, there is no project or user toolchain
+    // Pass through to the existing PATH
+    ToolCommand::passthrough(
+        &exe,
+        args,
+        ErrorDetails::BinaryNotFound {
```
mikrostew: If the above is changed, then this error message should be changed to indicate that there was no platform found and the binary was not found in the system.

crates/notion-core/src/tool/npx.rs
charlespierce: Done.

crates/notion-core/src/tool/binary.rs
charlespierce: My only concern is that we're then executing the user version of a tool inside of a package that we know for a fact has that tool as a dependency, because we don't have a way to invoke that local version. Of note, the current change matches the previous behavior, where we would throw an error here instead of falling through to checking for a user tool.

I'm a bit on the fence about whether making it so `jetson install typescript` lets you run `tsc` anywhere makes more sense than preventing users from accidentally running a global tool in a local workspace, which is one of the goals of Jetson as a product. That said, we are opening that protection up a little anyway by using the passthrough, so I suppose it's not unreasonable.

What about a third option: We check for the user tool settings, and if we find them, we execute the _local_ version (using `path_to_bin`) with the platform image provided by the `bin_config`. That matches the approach we would take if we found the user platform but not a local platform, just lining it up with the tool version. In fact, that might actually be better as the 2nd check, instead of the 3rd.

mikrostew: That might actually be a better option.

After thinking more about what I was saying above, I'm not sure it's a good idea to run the user tool version in this case, because it somewhat breaks the Jetson model of how things should work.

chriskrycho: I like @charlespierce's angle here, but I'd suggest *warning* in that case. It makes things *work* so users' flow isn't disrupted, but lets them know that there's something unusual in the flow.

charlespierce: I'm a little wary of warnings in places that will likely be called often, as that tends to result in 1 of 2 reactions:

1. The user sees the warning, thinks everything is broken, and files an issue or stops using Jetson.
2. The user realizes the warning isn't relevant and mentally tunes it out, making future warnings less impactful.

Though thinking on it a bit more, I'm actually not sure that looking for a user tool is the right approach at all. If we are in a project that has a local version of a tool, we should definitely be calling that local version. In order to minimize the chance of weird node version mismatch bugs, we should try to use the same version of Node with which the local version was installed:

- If the project doesn't have a `toolchain` key, then it most likely wasn't installed using a project-local Node version.
- If the user doesn't have a default Node version, then it wasn't installed using the user default Node version.
- The only thing left is that it was installed by `npm` or `yarn` that already existed on the system, either before Jetson was installed or through the passthrough mechanism.

In the last case, we should try to use the same Node version that already existed by passing through as well. We should, however, still be executing the local tool version, so we should update this passthrough to use the full path to the binary, instead of just the tool name (`bin_path` instead of `exe` in the code).

The user tools are installed in a sandbox that isn't connected to the versions of `npm` or `yarn` that the user executes, while local tools are always installed through `npm install` or `yarn`, so I don't think we should be looking towards the sandbox when executing local tools.

chriskrycho: Fair point on warnings; my inclination is generally to give more rather than less information. It's *useful* to provide info and not only silently do the "right" thing where possible, because it gives the user opportunities to opt into (what we would describe as) better alternatives. However, you're quite right about the two failures modes, and designing a solution that solves all of those is non-trivial.

I'll keep mulling on the design question here, and I'm also now thinking on how we might give users one-time or silence-able warnings for scenarios like that.

charlespierce: For providing information, I think this would be a great use-case for an expanded `--verbose` flag, as described in #362, if the user opts into the additional information, we can list which platform we are choosing and where it came from as part of the output.

For tools that's a little tricky, we may need an environment variable like `JETSON_VERBOSE` or something to switch that on, but we could provide a way for users to opt-in to seeing more information about the decisions Jetson is making

charlespierce:
Updated to resolve merge conflicts. The AppVeyor failure appears to be an intermittent network issue, not a failure with the code at all.

crates/notion-core/src/tool/binary.rs
mikrostew: Sorry I lost track of this PR. Just looked at the code again to refresh my memory on this flow. I think this echoes what @charlespierce is saying above.

At this point, the user is:
* in a project,
* running a shimmed executable,
* where that executable is a dependency of that project,
* and the project does not have `toolchain`,
* and the user does not have a default `node`

The reason there is a shim at this point is that the user ran `notion install <package>`, so they have this executable available as a user tool.

But:
* this project does not use Notion (there's no `toolchain`),
* this project's dependencies were not installed with the user version of `node`, because there is no default

So I think we should do what @charlespierce is suggesting above - try to execute the locally installed version of the executable, using the system Node, which is presumably what was used to install it.

I would prefer to give some kind of info/warning that Notion is running the local version, with a CTA of `Run 'notion install node' or 'notion pin node' to fix this warning`, something like that, while still trying to do the right thing.

crates/notion-core/src/tool/binary.rs
mikrostew: This doesn't make sense now given that I don't think we should fall through to the user tool above

crates/notion-core/src/tool/binary.rs
charlespierce: On the warning subject, what we do in this PR is keep the error we would have thrown before the passthrough (`NoPlatform` in this case), which tells the user that `node` isn't installed and calls out `notion install node` / `notion pin node`. Then if the passthrough fails for any reason, we show that error. So if the user has their system set up and hasn't opted in to using Notion fully, then it should continue to work exactly as they expect. If it doesn't, then we can guide them to how to fix it.

I'm wary of providing warnings in cases where it actually works, because then users get annoyed at seeing warnings where they don't actually affect the execution of the tool.

charlespierce: @mikrostew And yes, I didn't say in the last comment, but your summary describes exactly the situation we are in, and the current solution to it.

mikrostew: 👍

Issue #32, opened 2018-02-10
cache installers as they're fetched


dherman:
`notion_core::installer::node::Installer::remote()` should put a TeeReader around its input source to save the results to a cache file. The cache file should be stored in a directory specified by `notion_core::path::node_cache_dir()`.

dherman:
This issue must have been filed in a moment of forgetfulness — both the tarball and zip versions already generate the cache file. The logic lives in the `fetch` function.

PR #383, opened 2019-04-24
Update version handling.

chriskrycho:
- Implements [RFC #32](https://github.com/notion-cli/rfcs/pull/32)

### DIFF
--- a/crates/notion-core/src/tool/mod.rs
+++ b/crates/notion-core/src/tool/mod.rs
@@ -7,7 +7,11 @@ use std::marker::Sized;
 use std::path::Path;
 use std::process::{Command, ExitStatus};
 
+use lazy_static::lazy_static;
 use regex::Regex;
+
+use notion_fail::{Fallible, ResultExt};
+use validate_npm_package_name::{validate, Validity};
 
 use crate::command::create_command;
 use crate::env::UNSAFE_GLOBAL;
@@ -15,7 +19,6 @@ use crate::error::ErrorDetails;
 use crate::path;
 use crate::session::Session;
 use crate::version::VersionSpec;
-use notion_fail::{Fallible, ResultExt};
 
 mod binary;
 mod node;
@@ -28,6 +31,12 @@ use self::node::Node;
 use self::npm::Npm;
 use self::npx::Npx;
 use self::yarn::Yarn;
+
+lazy_static! {
+    static ref TOOL_SPEC_PATTERN: Regex =
+        Regex::new("^(?P<name>(?:@([^/]+?)[/])?([^/]+?))(@(?P<version>.+))?$")
+            .expect("regex is valid");
+}
 
 /// Specification for a tool and its associated version.
 ///
@@ -82,13 +91,25 @@ impl ToolSpec {
         Ok(())
     }
 
+    /// Try to parse a tool and version from a string like `<tool>[@<version>].
     pub fn try_from_str(tool_spec: &str) -> Fallible<Self> {
-        let re = Regex::new("(?P<name>@?[-/_A-z]+)(@(?P<version>.+))?").expect("regex is valid");
-        let captures = re
-            .captures(tool_spec)
-            .ok_or(ErrorDetails::ParseToolSpecError {
+        let captures =
+            TOOL_SPEC_PATTERN
+                .captures(tool_spec)
+                .ok_or(ErrorDetails::ParseToolSpecError {
+                    tool_spec: tool_spec.into(),
+                })?;
+
+        // Validate that the captured name is a valid NPM package name.
+        let name = &captures["name"];
+        if let Validity::Invalid { errors, warnings } = validate(name) {
+            dbg!(errors);
+            dbg!(warnings);
+            return Err(ErrorDetails::ParseToolSpecError {
                 tool_spec: tool_spec.into(),
-            })?;
+            }
+            .into());
+        }
 
         let version = captures
             .name("version")
@@ -96,7 +117,7 @@ impl ToolSpec {
             .transpose()?
             .unwrap_or_default();
 
-        Ok(match &captures["name"] {
+        Ok(match name {
             "node" => ToolSpec::Node(version),
             "npm" => ToolSpec::Npm(version),
             "yarn" => ToolSpec::Yarn(version),
--- a/crates/notion-core/src/tool/mod.rs
+++ b/crates/notion-core/src/tool/mod.rs
@@ -29,6 +29,15 @@ use self::npm::Npm;
 use self::npx::Npx;
 use self::yarn::Yarn;
 
+/// Specification for a tool and its associated version.
+///
+/// Since [`Ord`] is implemented for `ToolSpec`, we can use `.sort` on any
+/// `Vec<ToolSpec>`, and the order of the enum variants in the declaration
+/// determines the sorting order, which lets us guarantee (for example) that
+/// Node will always be prioritized over other tools in the toolchain when
+/// dealing with multiple tools.
+///
+/// [`Ord`]: https://doc.rust-lang.org/1.34.0/core/cmp/trait.Ord.html
 #[derive(PartialEq, Eq, PartialOrd, Ord)]
 pub enum ToolSpec {
     Node(VersionSpec),
@@ -41,8 +50,8 @@ impl ToolSpec {
     pub fn from_str_and_version(tool_name: &str, version: VersionSpec) -> Self {
         match tool_name {
             "node" => ToolSpec::Node(version),
-            "yarn" => ToolSpec::Yarn(version),
             "npm" => ToolSpec::Npm(version),
+            "yarn" => ToolSpec::Yarn(version),
             package => ToolSpec::Package(package.to_string(), version),
         }
     }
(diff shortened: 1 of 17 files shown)


crates/notion-core/src/tool/mod.rs
```diff
@@ -174,3 +199,214 @@ fn intercept_global_installs() -> bool {
     // We should only intercept global installs if the NOTION_UNSAFE_GLOBAL variable is not set
     env::var_os(UNSAFE_GLOBAL).is_none()
 }
+
+#[cfg(test)]
+mod tests {
+    mod try_from_str {
```
chriskrycho: The extra module here makes the test output look nice: `tools::tests::try_from_str::parses_bare_node`, etc.

crates/notion-core/src/tool/mod.rs
```diff
@@ -69,6 +72,28 @@ impl ToolSpec {
         }
         Ok(())
     }
+
+    pub fn try_from_str(tool_spec: &str) -> Fallible<Self> {
+        let re = Regex::new("(?P<name>@?[-/_A-z]+)(@(?P<version>.+))?").expect("regex is valid");
```
chriskrycho: I just realized: I need to double-check this regex against the actual spec for package names!

src/command/install.rs
```diff
@@ -2,33 +2,32 @@ use structopt::StructOpt;
...
-            Some(version_string) => VersionSpec::parse(version_string)?,
-            None => VersionSpec::default(),
-        };
-        let tool = ToolSpec::from_str_and_version(&self.tool, version);
+    fn run(mut self, session: &mut Session) -> Fallible<ExitCode> {
+        self.tools.sort();
 
-        tool.install(session)?;
+        for tool in self.tools {
+            session.add_event_start(ActivityKind::Install);
```
chriskrycho: Question: should these be inside the loop or run once per `run` invocation?

crates/notion-core/src/tool/mod.rs
```diff
@@ -27,10 +29,11 @@ use self::npm::Npm;
 use self::npx::Npx;
 use self::yarn::Yarn;
 
+#[derive(PartialEq, Eq, PartialOrd, Ord)]
```
chriskrycho: Including all of these up to `Ord` means the sequence defined here will be used for sorting when we have a `Vec<ToolSpec>`. I note that this is not obvious, so will add a comment explaining as much here.

crates/notion-core/src/tool/mod.rs
```diff
@@ -69,6 +72,28 @@ impl ToolSpec {
...
+
+        let version = captures
+            .name("version")
+            .map(|version| VersionSpec::parse(version.as_str()))
+            .transpose()?
+            .unwrap_or_else(VersionSpec::default);
+
+        Ok(match &captures["name"] {
+            "node" => ToolSpec::Node(version),
+            "yarn" => ToolSpec::Yarn(version),
```
mikrostew: Change this to match the new ordering in `ToolSpec` above?

crates/notion-core/src/tool/mod.rs
mikrostew: Ah, didn't see your edit while I was looking at this. Looks like that should fix the issues I found.

src/command/pin.rs
```diff
@@ -3,42 +3,40 @@ use structopt::StructOpt;
...
-            ToolSpec::Yarn(version) => session.pin_yarn(&version)?,
-            // ISSUE(#292): Implement install for npm
-            ToolSpec::Npm(_version) => unimplemented!("Pinning npm is not supported yet"),
-            ToolSpec::Package(name, _version) => {
-                throw!(ErrorDetails::CannotPinPackage { package: name })
+    fn run(mut self, session: &mut Session) -> Fallible<ExitCode> {
+        self.tools.sort();
+
+        for tool in self.tools {
+            session.add_event_start(ActivityKind::Pin);
```
charlespierce: Sorry, misclicked and accidentally deleted your comment, instead of the one I had written 😱 

The question was whether we should do this inside the loop or once per invocation of `run`.

I don't think it matters too much at the moment, the event reporting system isn't really used and hasn't been particularly maintained. I would lean towards once per invocation of `run`, mostly because of the inclusion of an ExitCode in the call to `add_event_end`. Then when we flesh out the event reporting we can add additional `ActivityKind`s to represent the intermediate steps (and likely refactor the events to not always require an exit code to end).

src/command/pin.rs
```diff
@@ -3,42 +3,40 @@ use structopt::StructOpt;
...
+            match tool {
+                ToolSpec::Node(version) => session.pin_node(&version)?,
```
charlespierce: I think the fail early approach is reasonable for the MVP, but that long-term we might want to be more robust (that may require some changes to the error reporting to support multiple errors). We can flesh it out more in the RFC, but that shouldn't be a blocker for this implementation.

crates/notion-core/src/tool/mod.rs
```diff
@@ -69,6 +72,28 @@ impl ToolSpec {
...
+            .captures(tool_spec)
+            .ok_or(ErrorDetails::ParseToolSpecError {
+                tool_spec: tool_spec.into(),
+            })?;
+
+        let version = captures
+            .name("version")
+            .map(|version| VersionSpec::parse(version.as_str()))
+            .transpose()?
+            .unwrap_or_else(VersionSpec::default);
```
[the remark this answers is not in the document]
chriskrycho: Today I learned! That's excellent! Just changed it and will include it.

src/command/pin.rs
chriskrycho: Sounds good; I will push it to the outer part of the `run` commands!

src/command/pin.rs
chriskrycho: Seems reasonable to me. I was thinking similarly. We should also in principle be able to parallelize those with something like [Rayon](https://crates.io/crates/rayon) and then capture the results in a nice way.

crates/notion-core/src/tool/mod.rs
chriskrycho: Great suggestion; I'm doing likewise throughout the module.

crates/notion-core/src/version/mod.rs
```diff
@@ -10,7 +10,7 @@ use notion_fail::{Fallible, ResultExt};
 
 use self::serial::parse_requirements;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
```
charlespierce: Is there a need for the `PartialEq, Eq, PartialOrd, Ord` derivations on `VersionSpec`? I don't see any other changes that would require comparing / sorting them. Not that it's a big issue either way, just curious

chriskrycho: Yes: because `VersionSpec`s are used in `ToolSpec`s. For a type to implement these traits, the types it depends on must *also* implement them.

charlespierce: Ah, makes perfect sense, thanks!

crates/notion-core/src/tool/mod.rs
chriskrycho: I ended up solving this by using the relatively loose regex I had here, just with a `^` to make sure it only matches the start of the string and tweaked slightly to match the regex used by npm, and then implementing more rigorous testing in a way we can later upstream in the new `validate-npm-package-name` crate I added (and which we can extract to its own crate at will).

src/command/pin.rs
mikrostew: Filed issue #389 for this so it doesn't fall through the cracks

mikrostew:
LGTM, I would just like to see some tests around `validate-npm-package-name`

crates/notion-core/src/tool/mod.rs
```diff
@@ -60,15 +81,49 @@ impl ToolSpec {
...
+                .captures(tool_spec)
+                .ok_or(ErrorDetails::ParseToolSpecError {
+                    tool_spec: tool_spec.into(),
+                })?;
+
+        // Validate that the captured name is a valid NPM package name.
+        let name = &captures["name"];
+        if let Validity::Invalid { errors, warnings } = validate(name) {
+            dbg!(errors);
+            dbg!(warnings);
```
charlespierce: Should these debugging prints be removed? It looks from [the docs](https://doc.rust-lang.org/std/macro.dbg.html) that it isn't removed on production builds, it's primarily a shortcut for `eprintln!("{:?}", expr)`

crates/notion-core/src/tool/mod.rs
```diff
@@ -60,15 +81,49 @@ impl ToolSpec {
...
+            return Err(ErrorDetails::ParseToolSpecError {
+                tool_spec: tool_spec.into(),
```
charlespierce: Since we're doing the separate check for the validity of the tool name, can we have a different error type that lets the user know the issue is with the name of the tool? It would be confusing to do `notion install _invalid@5` and be told `Please supply a spec in the form <tool name>[@<version>]`

crates/notion-core/src/tool/mod.rs
chriskrycho: Oh, yes! Thank you, and good catch!

crates/notion-core/src/tool/mod.rs
chriskrycho: Wondered the same; I’m happy to change it!

chriskrycho: Done!

chriskrycho:
Here's how the output looks with the updated handling for validating package names, @charlespierce:

output

Issue #151, opened 2018-09-18
fetch binaries instead of embedding them in installer


dherman:
The `get.notionjs.com` installation flow already has a wrapper script that fetches the installer, so it's really pointless to embed the binaries in the shell script itself. And the template expander script is committing some _heinous_ crimes against software engineering. We should simplify the install script to just be a simpler script that curls all the binaries it needs, and eliminate all the templating entirely.

PR #386, opened 2019-04-26
Release script, and script tests in CI

mikrostew:
Start to implement #151 by packaging the compiled binaries and shell scripts for release.

```
$ ./dev/unix/release.sh
    Building Notion for notion-0.3.0-macos
    Finished release [optimized] target(s) in 0.69s
   Packaging the compiled binaries and shell scripts
a notion
a shim
a load.fish
a load.sh
   Completed release in file target/release/notion-0.3.0-macos.tar.gz
```

```
$ ./dev/unix/release.sh --dev
    Building Notion for notion-0.3.0-macos
    Finished dev [unoptimized + debuginfo] target(s) in 0.43s
   Packaging the compiled binaries and shell scripts
a notion
a shim
a load.fish
a load.sh
   Completed release in file target/debug/notion-0.3.0-macos.tar.gz
```

```
$ ./dev/unix/release.sh -h
release.sh

Compile and package a release for Notion

USAGE:
    ./dev/unix/release.sh [FLAGS] [OPTIONS]

FLAGS:
    -h, --help          Prints this help info

OPTIONS:
        --release       Build artifacts in release mode, with optimizations (default)
        --dev           Build artifacts in dev mode, without optimizations
```

Also, this uses [bats](https://github.com/bats-core/bats-core) to test this script in CI.

### DIFF
--- a/dev/unix/release.sh
+++ b/dev/unix/release.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+# Script to build the binaries and package them up for release.
+# This should be run from the top-level directory.
+
+notion_info() {
+  local ACTION="$1"
+  local DETAILS="$2"
+  command printf '\033[1;32m%12s\033[0m %s\n' "${ACTION}" "${DETAILS}" 1>&2
+}
+
+notion_error() {
+  command printf '\033[1;31mError\033[0m: %s\n' "$1" 1>&2
+}
+
+bold() {
+  command printf '\033[1m%s\033[0m' "$1"
+}
+
+# parse the 'version = "0.3.0"' line from the input text
+parse_version() {
+  local contents="$1"
+
+  while read -r line
+  do
+    if [[ "$line" =~ ^version\ =\ \"(.*)\" ]]
+    then
+      echo "${BASH_REMATCH[1]}"
+      return 0
+    fi
+  done <<< "$contents"
+
+  notion_error "Could not determine the current version"
+  return 1
+}
+
+# returns the os name to be used in the packaged release,
+# including the openssl info if necessary
+parse_os_info() {
+  local uname_str="$1"
+  local openssl_version="$2"
+
+  case "$uname_str" in
+    Linux)
+      echo "linux-openssl-$(parse_openssl_version "$openssl_version")"
+      ;;
+    Darwin)
+      echo "macos"
+      ;;
+    *)
+      notion_error "Releases for '$uname_str' are not yet supported. Please modify this script and the install script to support this OS."
+      return 1
+  esac
+  return 0
+}
+
+# parse the OpenSSL version from the input text
+parse_openssl_version() {
+  let version_str="$1"
+
+  if [[ "$version_str" =~ ([^ ]*)\ (\d+).(\d+).(\d+) ]]
+  then
+    echo "matched $version_str" >&2
+    return 0
+  else
+    echo "could NOT match $version_str" >&2
+    return 1
+  fi
+}
+
+### END FUNCTIONS
+
+# read the current version from Cargo.toml
+cargo_toml_contents="$(<Cargo.toml)"
+NOTION_VERSION="$(parse_version "$cargo_toml_contents")"
+# TODO: check exit code
+
+# figure out the OS details
+os="$(uname -s)"
+openssl_version="$(openssl version)"
+NOTION_OS="$(parse_os_info "$os" "$openssl_version")"
+# TODO: check exit code
+
+release_filename="notion-$NOTION_VERSION-$NOTION_OS"
+
+# first make sure the release binaries have been built
+notion_info 'Building' "Notion release $(bold "$release_filename")"
+cargo build --release
+
+# then package up the binaries together
+target_dir="target/release"
+notion_info 'Packaging' "the compiled binaries"
+cd "$target_dir"
+tar -czf "$release_filename.tar.gz" notion shim
+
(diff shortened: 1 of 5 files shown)


dev/unix/tests/release-script.bats
```diff
@@ -0,0 +1,109 @@
+# test the release.sh script
+
+# load the functions from the script
+# (everything before the '### END FUNCTIONS' line)
+eval "$(cat dev/unix/release.sh | sed '/### END FUNCTIONS/q')"
```
mikrostew: Using `eval` here, because `source`-ing the file will execute all of the code after the functions, which is no good

dev/unix/release.sh
```diff
@@ -0,0 +1,110 @@
...
+# parse the OpenSSL version from the input text
+parse_openssl_version() {
+  local version_str="$1"
+
+  if [[ "$version_str" =~ ^([^\ ]*)\ ([0-9]+\.[0-9]+) ]]
+  then
+    # check that lib name is 'OpenSSL'
+    libname="${BASH_REMATCH[1]}"
+    if [[ "$libname" != "OpenSSL" ]]; then
+      notion_error "Releases for '$libname' not currently supported"
```
arthirm: You want to add what is supported (OpenSSL) in the message to make it easy for fixing this error.

dev/unix/release.sh
```diff
@@ -0,0 +1,110 @@
...
+      return 1
+    fi
+    echo "${BASH_REMATCH[2]}"
+    return 0
+  else
+    notion_error "Could not determine OpenSSL version for '$version_str'"
```
arthirm: Again you want to point out what format the version should be to make it easier for fixing the error?

dev/unix/release.sh
```diff
@@ -0,0 +1,110 @@
...
+
+# parse the OpenSSL version from the input text
+parse_openssl_version() {
+  local version_str="$1"
+
+  if [[ "$version_str" =~ ^([^\ ]*)\ ([0-9]+\.[0-9]+) ]]
+  then
+    # check that lib name is 'OpenSSL'
+    libname="${BASH_REMATCH[1]}"
+    if [[ "$libname" != "OpenSSL" ]]; then
```
arthirm: Can we add the supported libs in some array of some sort? So that when we start adding more libs we don't have to modify this `if` condition rather we can just add it to that array.

dev/unix/release.sh
```diff
@@ -0,0 +1,110 @@
...
+        return "$exit_code"
+      fi
+
+      echo "linux-openssl-$major_minor"
+      ;;
+    Darwin)
+      echo "macos"
+      ;;
+    *)
+      notion_error "Releases for '$uname_str' are not yet supported. You will need to modify this script and the install script to support this OS."
```
arthirm: Can we add more details on what needed to be modified on those scripts?

dev/unix/release.sh
mikrostew: Good point - added more to the error message

dev/unix/release.sh
mikrostew: I added some more detail about what to change to fix this error

dev/unix/release.sh
mikrostew: Good idea - I'll add an associative array to hold the supported libraries

dev/unix/release.sh
mikrostew: Added some more detail about what specifically needs to be changed

dev/unix/release.sh
```diff
@@ -0,0 +1,135 @@
...
+  done
+  return 1
+}
+
+# parse the OpenSSL version from the input text
+parse_openssl_version() {
+  local version_str="$1"
+
+  # array containing the SSL libraries that are supported
+  # would be nice to use a bash 4.x associative array, but bash 3.x is the default on OSX
```
mikrostew: why OSX why

dev/unix/release.sh
```diff
@@ -0,0 +1,135 @@
...
+  SUPPORTED_SSL_LIBS=( 'OpenSSL' )
+
+  if [[ "$version_str" =~ ^([^\ ]*)\ ([0-9]+\.[0-9]+) ]]
```
charlespierce: I'm not sure if the current release process supports this, but we probably need to support parsing the full 3-part version of OpenSSL, since `1.0.1`, `1.0.2`, `1.1.0`, and `1.1.1` are all distinct and incompatible with one another. Though we also need to make sure to drop any letters from the end of the version, as that appears to be how OpenSSL distinguishes patch versions e.g. `1.0.2b` (which do seem to be compatible with one another).

mikrostew: Ah, good point - I'll update this to use all three digits. I'm also working to revamp the install process, and will change that to use all three digits as well.

dev/unix/tests/release-script.bats
benblank: Would it be better to extract the functions into a "shared library", to avoid having to `sed` scripts?

mikrostew: Eventually I plan to put these functions into the install script, and will source that from this script, so that both scripts are using the same functions (and the install script can be downloaded and run individually). I'm currently working on that.

For now, I found a way to detect if the script is being sourced, and return if that is the case (adapted from https://stackoverflow.com/questions/2683279/how-to-detect-if-a-script-is-being-sourced)

PR #396, opened 2019-05-01
Ensure package binary files are executable on Unix

### LINKED ISSUE #351: Ensure 3rd party binaries are executable
When we execute a 3rd party binary, we are simply shelling out to whatever script is listed in the `bin` key in `package.json` for that package. In some cases (e.g. `ts-node`), that script file is not marked as executable, which causes `Permission denied` errors when trying to execute it. We should make sure when installing a package that all script files are marked as executable in the file system.

Note: This only affects Unix environments, as Windows doesn't have non-executable mode.

### LINKED ISSUE #385: notion install @sentry/craft fails
from the error log after running the installed `craft`:

```
cause: Directory not empty (os error 66)
```

This is with notion 0.3.0

charlespierce:
Closes #351 
Closes #385 

On unix, ensure that all 3rd-party binary files installed through `notion install <package>` are marked as executable in the file-system. This is necessary because we call the file directly when executing the shim, and if it is not executable, then we get a `permission denied` error.

On windows, this isn't a problem as there isn't a separate concept of "executable" permissions.

### DIFF
--- a/crates/notion-core/src/error/details.rs
+++ b/crates/notion-core/src/error/details.rs
@@ -101,6 +101,11 @@ pub enum ErrorDetails {
         command: String,
     },
 
+    /// Thrown when verifying the file permissions on an executable fails
+    ExecutablePermissionsError {
+        bin: String,
+    },
+
     /// Thrown when executing a hook command fails
     ExecuteHookError {
         command: String,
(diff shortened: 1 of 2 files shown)


crates/notion-core/src/error/details.rs
```diff
@@ -481,12 +486,15 @@ from {}
...
-Please ensure that the corrent command is specified.", command),
             ErrorDetails::ExecutablePathError { command } => write!(f, "Could not determine path to executable '{}'
 
 {}", command, REPORT_BUG_CTA),
+            ErrorDetails::ExecutablePermissionsError { bin } => write!(f, "Could not verify permissions for executable '{}'
+
+{}", bin, PERMISSIONS_CTA),
+            ErrorDetails::ExecuteHookError { command } => write!(f, "Could not execute hook command: '{}'
+
+Please ensure that the corrent command is specified.", command),
```
chriskrycho: Typo! You're looking for `correct` I think?

charlespierce: Good catch! Glad I moved that error to put it in the right place alphabetically!

PR #397, opened 2019-05-01
Add success messages to 'notion pin' and 'notion install'

### LINKED ISSUE #357: Improve the messaging with `notion install node`
After running `notion install node` nothing is reported back to the user. There should be some success messaging letting the users know that the command was successful.

Encountered this with notion v0.3.0.

cc: @dherman @charlespierce @mikrostew

charlespierce:
Closes #357 

Adds a success message to `notion install node`, `notion install yarn`, `notion pin node`, and `notion pin yarn`, so that the user is informed of success, instead of leaving the terminal completely blank.

### DIFF
--- a/crates/notion-core/src/session.rs
+++ b/crates/notion-core/src/session.rs
@@ -174,16 +174,30 @@ impl Session {
     /// Fetch and unpack a version of Node matching the input requirements.
     pub fn install_node(&mut self, version_spec: &VersionSpec) -> Fallible<()> {
         let node_distro = self.fetch_node(version_spec)?.into_version();
+        let success_message = format!(
+            "Installation successful. Default node version set to {}",
+            &node_distro.runtime
+        );
         let toolchain = self.toolchain.get_mut()?;
+
         toolchain.set_active_node(node_distro)?;
+        println!("{}", success_message);
+
         Ok(())
     }
 
     /// Fetch and unpack a version of Yarn matching the input requirements.
     pub fn install_yarn(&mut self, version_spec: &VersionSpec) -> Fallible<()> {
         let yarn_distro = self.fetch_yarn(version_spec)?.into_version();
+        let success_message = format!(
+            "Installation successful. Default yarn version set to {}",
+            &yarn_distro
+        );
         let toolchain = self.toolchain.get_mut()?;
+
         toolchain.set_active_yarn(yarn_distro)?;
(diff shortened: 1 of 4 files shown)


crates/notion-core/src/session.rs
```diff
@@ -174,16 +174,30 @@ impl Session {
     /// Fetch and unpack a version of Node matching the input requirements.
     pub fn install_node(&mut self, version_spec: &VersionSpec) -> Fallible<()> {
         let node_distro = self.fetch_node(version_spec)?.into_version();
+        let success_message = format!(
+            "Installation successful. Default node version set to {}",
```
chriskrycho: Can we format this to match the rest of the installation/operation output? And the same with the others below?

charlespierce: I'm not sure I follow, what installation / operation output do you mean?

chriskrycho: Sorry; I was meaning that we could match the styling for things like `Fetching` or `error:` and print it with something like:

```
*Success*: installed and set node@{} as default.
```

where `*Success*` should be green instead of wrapped in `*` characters.

charlespierce: Ahh, good call, I'll update to do that :+1:

crates/notion-core/src/session.rs
```diff
@@ -174,29 +175,23 @@ impl Session {
     /// Fetch and unpack a version of Node matching the input requirements.
     pub fn install_node(&mut self, version_spec: &VersionSpec) -> Fallible<()> {
         let node_distro = self.fetch_node(version_spec)?.into_version();
-        let success_message = format!(
-            "Installation successful. Default node version set to {}",
-            &node_distro.runtime
-        );
+        let success_message = format!("installed and set node@{} as default", &node_distro.runtime);
```
chriskrycho: We can use one of the other `style` helper functions to make sure the tool names here get formatted consistently:

```rust
use crate::style::{format_success_message, tool_version};
```

```rust
let success_message =
    format!("installed and set {} as default", tool_version("node", &node_distro.runtime));
```

charlespierce: Good catch, updated!

charlespierce:
Updated to use the `tool_version` helper and also to remove redundant output when doing `notion pin`

Issue #334, opened 2019-03-31
get rid of notion-fail


dherman:
Since we now have a single `ErrorDetails` error type, I wonder if we can completely eliminate the `notion-fail` crate and all its abstractions. If I remember right, those were all needed to deal with the fact that there were _n_ Notion error types, but now there's only 1. It would be nice if `Fallible<T>` were just a shorthand for `Result<T, ErrorDetails>` or something like that.

This is off the top of my head so I could be wrong; needs some investigation.

charlespierce:
On some further investigation into the usefulness of having separate crates for some operations (specifically the `layout` abstraction allowing us to version the layout separately from the overall project), we will likely have > 1 error types going forward, so instead of removing `volta-fail` completely, we should instead do #306 and remove the no-longer-supported `unknown()` method.

PR #403, opened 2019-05-03
Provide informative error for `notion <action> <tool> <version-like>`.

chriskrycho:
Users coming from tools like `nvm` or `nodenv` or other ecosystems may expect `notion install node 10` to install Node v10. Since it will instead install `node@lts` and `10@latest` -- and the package `10` [does indeed exist!][10] -- provide an error message when Notion detects that this scenario has occurred.

[10]: https://www.npmjs.com/package/10

### DIFF
--- a/crates/notion-core/src/error/details.rs
+++ b/crates/notion-core/src/error/details.rs
@@ -549,7 +549,7 @@ Please ensure that the command output is valid UTF-8 text.", command),
                 );
 
                 let call_to_action = format!(
-"To install '{name}' version '{version}', please run `notion {action} {formatted}`. \
+"To {action} '{name}' version '{version}', please run `notion {action} {formatted}`. \
 To {action} the packages '{name}' and '{version}', please {action} them in separate commands, or with explicit versions.",
                     action=action,
                     name=name,
--- a/crates/notion-core/src/error/details.rs
+++ b/crates/notion-core/src/error/details.rs
@@ -5,7 +5,7 @@ use textwrap::{fill, indent};
 
 use notion_fail::{ExitCode, NotionFail};
 
-use crate::style::display_width;
+use crate::style::{text_width, tool_version};
 use crate::tool::ToolSpec;
 
 const REPORT_BUG_CTA: &'static str =
@@ -16,7 +16,7 @@ an issue at https://github.com/notion-cli/notion/issues with the details!";
 const PERMISSIONS_CTA: &'static str =
     "Please ensure you have correct permissions to the Notion directory.";
 
-#[derive(Debug, Fail)]
+#[derive(Debug, Fail, PartialEq)]
 pub enum ErrorDetails {
     /// Thrown when package tries to install a binary that is already installed.
     BinaryAlreadyInstalled {
@@ -127,6 +127,14 @@ pub enum ErrorDetails {
     /// Thrown when output from a hook command could not be read
     InvalidHookOutput {
         command: String,
+    },
+
+    /// Thrown when a user does e.g. `notion install node 12` instead of
+    /// `notion install node@12`.
+    InvalidInvocation {
+        action: String,
+        name: String,
+        version: String,
     },
 
     /// Thrown when a tool name is invalid per npm's rules.
@@ -531,9 +539,30 @@ Please ensure that the correct command is specified.", command),
             ErrorDetails::InvalidHookOutput { command } => write!(f, "Could not read output from hook command: '{}'
 
 Please ensure that the command output is valid UTF-8 text.", command),
+
+            ErrorDetails::InvalidInvocation { action, name, version } => {
+                let error = format!(
+                    "`notion {action} {name} {version}` is not supported.",
+                    action=action,
+                    name=name,
+                    version=version
+                );
+
+                let call_to_action = format!(
+"To install '{name}' version '{version}', please run `notion {action} {formatted}`. \
+To {action} the packages '{name}' and '{version}', please {action} them in separate commands, or with explicit versions.",
+                    action=action,
+                    name=name,
+                    version=version,
+                    formatted=tool_version(name, version)
+                );
+
+                write!(f, "{}\n\n{}", error, fill(&call_to_action, text_width()))
+            },
+
             ErrorDetails::InvalidToolName { name, errors } => {
                 let indentation = "    ";
-                let wrapped = &fill(&errors.join("\n"), display_width() - indentation.len());
+                let wrapped = &fill(&errors.join("\n"), text_width() - indentation.len());
                 let formatted_errs = indent(&wrapped, indentation);
 
                 let call_to_action = if errors.len() > 1 {
@@ -542,11 +571,9 @@ Please ensure that the command output is valid UTF-8 text.", command),
                     "Please fix the following error:"
                 };
 
-                write!(f, "Invalid tool name `{}`
-
-{}
-{}", name, call_to_action, formatted_errs)
+                write!(f, "Invalid tool name `{}`\n\n{}\n{}", name, call_to_action, formatted_errs)
             },
+
             ErrorDetails::NoBinPlatform { binary } => {
                 write!(f, "Platform info for executable `{}` is missing
 
@@ -863,6 +890,7 @@ impl NotionFail for ErrorDetails {
             ErrorDetails::HookNoFieldsSpecified => ExitCode::ConfigurationError,
             ErrorDetails::InvalidHookCommand { .. } => ExitCode::ExecutableNotFound,
             ErrorDetails::InvalidHookOutput { .. } => ExitCode::ExecutionFailure,
+            ErrorDetails::InvalidInvocation { .. } => ExitCode::InvalidArguments,
             ErrorDetails::InvalidToolName { .. } => ExitCode::InvalidArguments,
             ErrorDetails::NoBinPlatform { .. } => ExitCode::ExecutionFailure,
             ErrorDetails::NodeVersionNotFound { .. } => ExitCode::NoVersionMatch,
--- a/crates/notion-core/src/tool/mod.rs
+++ b/crates/notion-core/src/tool/mod.rs
@@ -160,9 +160,6 @@ impl ToolSpec {
     }
 
     /// Check the args for the bad pattern of `notion install <tool> <number>`.
-    ///
-    /// Requires you to pass *all* args. Should only be called from commands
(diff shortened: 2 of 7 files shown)


crates/notion-core/src/error/details.rs
```diff
@@ -16,7 +16,7 @@ an issue at https://github.com/notion-cli/notion/issues with the details!";
 const PERMISSIONS_CTA: &'static str =
     "Please ensure you have correct permissions to the Notion directory.";
 
-#[derive(Debug, Fail)]
+#[derive(Debug, Fail, PartialEq)]
```
chriskrycho: Adding `PartialEq` lets us check against this in tests! …though not without some finagling, as you can see below. (#334 would simplify a lot of that, however.)

crates/notion-core/src/tool/mod.rs
```diff
@@ -123,6 +125,35 @@ impl ToolSpec {
             package => ToolSpec::Package(package.into(), version),
         })
     }
+
+    /// Check the args for the bad pattern of `notion install <tool> <number>`.
+    ///
+    /// Requires you to pass *all* args. Should only be called from commands
+    /// like `notion install`, `notion fetch`, or `notion pin`.
+    pub fn check_args<I>(mut args: I, action: String) -> Fallible<()>
```
charlespierce: I'm not sure this fits as part of `impl ToolSpec`, it seems like it might be a higher-level concern of the CLI, as that's where the coordination of installing multiple tools happens.

crates/notion-core/src/tool/mod.rs
```diff
@@ -123,6 +125,35 @@ impl ToolSpec {
...
+    where
+        I: Iterator<Item = String>,
+    {
+        // The case we are concerned with is where we have `<tool> <number>`.
+        // Those arguments will always follow `notion` and one of `pin`,
+        // `fetch`, or `install`. We use `nth(2)` to advance the iterator to the
+        // first item of interest, then `next` to get the other item.
+        if let (Some(name), Some(maybe_version)) = (args.nth(2), args.next()) {
```
charlespierce: Using specific positioning could pose a problem, as we have the `--verbose` flag that could ruin this logic, e.g. `notion --verbose install node 10` wouldn't cause an error because it's reading `install` and `node` instead of `node` and `10`.

It may be better to use the tools vector from the commands, since that has already been parsed by StructOpt.

Additionally, we need to make sure we only error if there are exactly 2 parameters. If someone does `notion install node 10 yarn`, I think we can let that through since at that point they know they're installing multiple different things.

src/command/fetch.rs
```diff
@@ -21,6 +21,7 @@ pub(crate) struct Fetch {
 impl Command for Fetch {
     fn run(mut self, session: &mut Session) -> Fallible<ExitCode> {
         session.add_event_start(ActivityKind::Fetch);
+        ToolSpec::check_args(std::env::args(), "fetch".into())?;
```
charlespierce: `std::env::args()` Will panic if any of the args is not valid UTF-8, so we've used `args_os()` everywhere else to deal with arguments. We should probably maintain that to prevent panics from malformed invocations.

crates/notion-core/src/error/details.rs
```diff
@@ -513,9 +521,30 @@ Please ensure that the correct command is specified.", command),
...
+            ErrorDetails::InvalidInvocation { action, name, version } => {
+                let error = format!(
+                    "`notion {action} {name} {version}` is not supported.",
+                    action=action,
+                    name=name,
+                    version=version
+                );
+
+                let call_to_action = format!(
+"To install '{name}' version '{version}', please run `notion {action} {formatted}`. \
```
benblank: I believe `To {action} '{name}'` is meant here, rather than `To install '{name}'`.

crates/notion-core/src/tool/mod.rs
```diff
@@ -123,6 +125,35 @@ impl ToolSpec {
             package => ToolSpec::Package(package.into(), version),
         })
     }
+
+    /// Check the args for the bad pattern of `notion install <tool> <number>`.
```
benblank: What about `notion <action> <tool> <number> <tool>…`?  E.g. should we also allow `notion install node 10 yarn ember-cli`?

crates/notion-core/src/tool/mod.rs
chriskrycho: Wondered about the same. I'll think about where to move it to.

crates/notion-core/src/tool/mod.rs
chriskrycho: Ah, blast, that's a good point re: flags. Unfortunately, the `tools` vector for the commands *can't* be used here: we've lost the relevant information by that point—namely, what the bare commands were. We can attach that, perhaps, but they're parsed into the relevant data structures by that point.

crates/notion-core/src/error/details.rs
chriskrycho: Good catch; I'll push a fix; thanks!

src/command/fetch.rs
chriskrycho: Ah, I had wondered about that. Will make that change!

crates/notion-core/src/tool/mod.rs
chriskrycho: Good point – @charlespierce caught the same.

crates/notion-core/src/tool/mod.rs
charlespierce: Ah, good point about not being able to use the tools, since once it’s parsed you can’t tell the difference between `node` and `node@lts`.

A couple ideas I had for working around that:

1. Change all of the StructOpt types to take a `Vec<String>`, then add a helper function (maybe in `command/mod.rs`) that accepts that vector and does this verification, the `ToolSpec` parsing, and the sorting. Then each command can call that function and use the results from there.

2. Use `.filter()` on the iterator of args, removing any args that begin with a `-`, then use the positional logic on that.

The latter is the approach we use in `notion-core/src/tool/yarn.rs` to detect global installs, since those are also positional (the first two non-flag arguments need to be `global install` in that order)

chriskrycho: I keep going back and forth. If we ever introduce an *optional argument* command specified like `--foo=true`, (2) breaks. At the same time, I *really* don't want to pull the parsing behavior out of the actual CLI parsing; it's *really* nice that if you have a command, it's already a valid set of `tools` and there's no further work to be done. 🤔

crates/notion-core/src/tool/mod.rs
chriskrycho: Ended up making this a *private* function in `ToolSpec`, which only exists to support the newly created `from_strings` function on `ToolSpec`.

crates/notion-core/src/tool/mod.rs
chriskrycho: Per our brief discussion on Discord, I ended up deciding that the best and cleanest thing was to wrap all of these bits into a `ToolSpec::from_strings` function which returns `Fallible<Vec<ToolSpec>>` – this worked out nicely and lets us just do this in the actual tool invocations:

```rust
for tool in ToolSpec::from_strings(&self.tools, "install")? {
  // ...
}
```

This includes being able to close over the need to make sure the tools are all sorted as their final operation, which is a nice win for maintainability.

crates/notion-core/src/error/details.rs
chriskrycho: Fixed in 15b9b01!

src/command/fetch.rs
chriskrycho: Ended up not needing to mess with `OsStr` at all, by way of refactoring it to do as you suggested and using `Vec<String>` for the various command structs.

crates/notion-core/src/tool/mod.rs
```diff
@@ -131,6 +133,61 @@ impl ToolSpec {
...
+    /// Returns a listed sorted so that if `node` is included in the list, it is
+    /// always first.
+    pub fn from_strings<T>(tool_strs: &mut [T], action: &str) -> Fallible<Vec<ToolSpec>>
+    where
+        T: AsRef<str>,
+    {
+        let mut iter = tool_strs.iter();
+        Self::check_args(iter.by_ref(), action)?;
+
+        let mut tools = iter
```
charlespierce: I think you actually need to use a new iterator here. Even though you're only borrowing it in the call to `check_args`, it's being advanced by the calls to `.next()`, so when you get to this point it's missing 3 elements.

crates/notion-core/src/tool/mod.rs
```diff
@@ -442,4 +499,67 @@ mod tests {
             );
         }
     }
+
+    mod from_strings {
```
charlespierce: Should also have assertions for the success cases that the number of elements in the result vector matches the number in the input vector.

chriskrycho: Agreed!

crates/notion-core/src/tool/mod.rs
chriskrycho: Ahhhh, good catch, I misunderstood what I was reading in the docs. I'll just drop the last commit and then add the test for the positive case below. Thank you!

chriskrycho:
Added those tests and dropped the bad commit. TIL re: iterators! (And I felt so clever for those few minutes. 😆)

crates/notion-core/src/tool/mod.rs
```diff
@@ -131,6 +133,62 @@ impl ToolSpec {
...
+    ///
+    /// Accounts for the following error conditions:
+    ///
+    /// - `notion install node 12`, where the user intended to install `node@12`
+    ///   but used syntax like in nodenv or nvm
+    /// - invalid version specs
+    ///
+    /// Returns a listed sorted so that if `node` is included in the list, it is
+    /// always first.
+    pub fn from_strings<T>(tool_strs: &mut [T], action: &str) -> Fallible<Vec<ToolSpec>>
```
charlespierce: It looks like we don't need the mutability for the tool list. I checked locally and opened a PR against your branch to pull it out.

chriskrycho:
Let’s merge it! (I don’t yet have permissions.)

crates/notion-core/src/tool/mod.rs
chriskrycho: Excellent catch!

PR #411, opened 2019-05-08
Add per-project configuration

### LINKED ISSUE #417: Implement per-project configuration RFC
Implement https://github.com/notion-cli/rfcs/pull/33

benblank:
This is an implementation of the as-yet unlanded [RFC](https://github.com/notion-cli/rfcs/pull/33).

(Also contains a change to the VS Code config to automatically `cargo fmt` when saving files.)

* [x] parse per-project hooks configuration and merge with per-user hooks
* [x] change format from TOML to JSON

Resolves #417.


crates/volta-core/src/hook/mod.rs
```diff
@@ -79,6 +153,48 @@ impl HookConfig {
             })?;
         src.parse()
     }
+
+    /// Creates a merged struct, with "right" having precedence over "left".
+    fn merge(left: Self, right: Self) -> Self {
+        Self {
+            node: if let Some(left_node) = left.node {
```
charlespierce: I think these blocks can be done a little cleaner with a `match` instead of nested `if let` statements:

```rust
node: match (left.node, right.node) {
    (Some(left), Some(right)) => Some(ToolHooks::merge(left, right)),
    (Some(left), None) => Some(left),
    _ => right.node,
}
```

And similarly for the other blocks.

crates/volta-core/src/hook/mod.rs
```diff
@@ -68,9 +72,79 @@ pub struct ToolHooks<D: Distro> {
...
+    fn for_current_dir() -> Fallible<Option<Self>> {
+        Self::for_dir(&env::current_dir().with_context(|_| ErrorDetails::CurrentDirError)?)
+    }
+
+    /// Returns the per-project hooks for the specified directory.  If the
+    /// specified directory is not itself a project, its ancestors will be
+    /// searched.
+    fn for_dir(base_dir: &Path) -> Fallible<Option<Self>> {
+        let mut dir = base_dir.clone();
+        while !is_project_root(dir) {
```
charlespierce: Since we're using the `is_project_root` as the differentiator, it would be nice to abstract this logic into a `find_project_root` method that takes `base_dir: &Path` in and returns an `Option<&Path>`. Then both this and the `project.rs` code can use that method, instead of duplicating the code for the ancestor searching.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
benblank: I thought it would be nice to include, as we already provide `.vscode/launch.json`.  I can drop it if you feel strongly about it.

crates/volta-core/src/hook/mod.rs
```diff
@@ -68,9 +72,79 @@ pub struct ToolHooks<D: Distro> {
     pub phantom: PhantomData<D>,
 }
 
+impl<D: Distro> ToolHooks<D> {
+    /// Creates a merged struct, with "right" having precedence over "left".
+    fn merge(left: Self, right: Self) -> Self {
+        Self {
+            distro: if let Some(_) = right.distro {
```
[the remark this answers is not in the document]
benblank: Nice!  Didn't know about that one, will change.

crates/volta-core/src/hook/mod.rs
benblank: Excellent, I had a feeling this wasn't fully idiomatic.

benblank: I had to tweak this a bit to avoid a "use of moved value" issue, but it's still much better than before.

crates/volta-core/src/hook/mod.rs
benblank: 👍, will do.

.vscode/settings.json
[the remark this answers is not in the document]
charlespierce: Fair point, it's reasonable to include and shouldn't hurt anything anyway :+1:

crates/volta-core/src/error/details.rs
```diff
@@ -816,11 +818,14 @@ Please ensure you have correct permissions.",
...
                 REPORT_BUG_CTA
             ),
-            ErrorDetails::ParseHooksError => write!(
+            ErrorDetails::ParseHooksError { file } => write!(
                 f,
-                "Could not parse hooks.toml configuration file.
+                "Could not parse configuration file.
 
-Please ensure the file is correctly formatted."
+{}
```
charlespierce: Nitpick: We've generally used `from {}` in other errors that reference a file name. See e.g. the message for `ErrorDetails::PackageReadError`.

Perhaps:

```rust
"Could not parse Volta configuration
from {}

Please ensure the file is correctly formatted."
```

crates/volta-core/src/hook/mod.rs
```diff
@@ -68,27 +68,106 @@ pub struct ToolHooks<D: Distro> {
...
+                Self::from_file(&path, false).map(|hooks| Some(hooks))
+            }
+            None => Ok(None),
+        }
     }
-}
 
-impl FromStr for HookConfig {
-    type Err = VoltaError;
+    fn from_file(file_path: &Path, create: bool) -> Fallible<Self> {
```
charlespierce: I'm a little wary of boolean parameters in methods, because it's difficult to tell from the call-site what that `true` or `false` value means. I think we can achieve this behavior by checking for an empty string instead of creating a file with `{}` as the contents:

```rust
fn from_file(file_path: &Path) -> Fallible<Self> {
    let src = touch(&file_path)
        .and_then(|mut file| file.read_into_string())
        .with_context(|_| ErrorDetails::ReadHooksError {
            file: file_path.to_string_lossy().to_string(),
        })?;

    if src.is_empty() {
        HookConfig {
            node: None,
            yarn: None,
            package: None,
            events: None,
        }
    } else {
        let serial: serial::HookConfig =
            serde_json::de::from_str(src).with_context(|_| ErrorDetails::ParseHooksError {
                file: file_path.to_string_lossy().to_string(),
            })?;
        serial.into_hook_config()
    }
}
```

crates/volta-core/src/project.rs
```diff
@@ -70,20 +54,13 @@ impl Project {
...
     fn for_dir(base_dir: &Path) -> Fallible<Option<Rc<Project>>> {
-        let mut dir = base_dir.clone();
-        while !is_project_root(dir) {
-            dir = match dir.parent() {
-                Some(parent) => parent,
-                None => {
-                    return Ok(None);
-                }
-            }
+        match path::project_for_dir(base_dir) {
```
charlespierce: 😍 This is so much cleaner now!

crates/volta-core/src/error/details.rs
benblank: Done.

crates/volta-core/src/hook/mod.rs
benblank: I'm not happy with the boolean argument, either, but `from_file` can't blindly call `touch`, because it's also used in the per-project hooks, where we **don't** want it to create a file when one doesn't already exist.

Really, I'd be happier not generating the per-user file, though that would be a departure from existing behavior.

chriskrycho: When I *have* to do something like that (it happens sometimes) I just create a standalone, two-value enum to make the invocation clear.

charlespierce: It appears from lines 110-114 above, that the per-project hook part only calls `from_file` if the file already exists on the disk, so the use of `touch` shouldn't change anything, correct?

benblank: I'd forgotten that!  Let me make these changes.

crates/volta-core/src/hook/mod.rs
```diff
@@ -68,27 +67,103 @@ pub struct ToolHooks<D: Distro> {
...
+            }
+            None => Ok(None),
+        }
     }
-}
 
-impl FromStr for HookConfig {
-    type Err = VoltaError;
+    fn from_file(file_path: &Path) -> Fallible<Self> {
+        let file = touch_with_contents(&file_path, "{}").with_context(|_| {
```
charlespierce: I keep going back and forth on whether we should do the empty file detection, as opposed to writing a default JSON object. On the one hand, the default object simplifies the logic here a bit since we don't have to do any reading. This is going to be a breaking change anyway, since people will have to migrate their `hooks.toml` to `hooks.json` format, so there isn't any major concern about porting the old empty file to a new one.

On the other hand, I feel like it would be nice if we supported an empty file (or even just no file instead of using `touch`), since that's a more obvious way of saying "No custom hooks" than a specific empty-object value.

@dherman @chriskrycho @mikrostew Thoughts?

thoov:
@benblank / @charlespierce Can we also make sure that we add documentation to the website about this change?

crates/volta-core/src/path/mod.rs
```diff
@@ -211,11 +212,51 @@ pub fn package_archive_root_dir_name(name: &str, version: &str) -> String {
...
+
+fn is_dependency(dir: &Path) -> bool {
+    dir.parent().map_or(false, |parent| is_node_modules(parent))
+}
+
+fn is_project_root(dir: &Path) -> bool {
+    is_node_root(dir) && !is_dependency(dir)
+}
+
+pub fn project_for_dir(base_dir: &Path) -> Option<&Path> {
```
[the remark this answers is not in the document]
benblank: Good change, done.

crates/volta-core/src/hook/mod.rs
benblank: I've updated this to no longer attempt to create `$VOLTA_HOME/hooks.json` if it isn't present.  It makes the code simpler and cleaner.

crates/volta-core/src/hook/mod.rs
```diff
@@ -68,27 +67,103 @@ pub struct ToolHooks<D: Distro> {
...
+
+    /// Returns the per-user hooks, loaded from the filesystem.
+    fn for_user() -> Fallible<Self> {
+        Self::from_file(&user_hooks_file()?)
+    }
+
+    /// Creates a merged struct, with "right" having precedence over "left".
+    fn merge(left: Self, right: Self) -> Self {
+        Self {
+            node: match (left.node, right.node) {
```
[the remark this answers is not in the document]
benblank: The macrofied version still isn't as terse as I'd like, but you're right, it's a lot better.

PR #414, opened 2019-05-09
Improve 'NoPlatform' error message

### LINKED ISSUE #363: Confusing error messages before installing first node or tool versions
It's fairly common to use `npm` and `yarn` to e.g. create new projects. However, if you have not yet run `notion install node` and, for yarn, also `notion install yarn`, this fails with this output:

```sh
Notion error: Could not determine Node version.

Use `notion pin node` to select a version for a project.
Use `notion install node` to select a default version.
```

The error message is technically correct… but I found it somewhat confusing (even for a fairly experienced user of Notion who has helped develop some of it!).

Two thoughts:

1. We may want to tweak this error message a little. It's *quite* terse at the moment and I think it might be helpful for the message to be *slightly* longer here, even if we keep the general terseness we've landed on to date: this is probably the most likely error for first-time users.

    ```sh
    Notion error: Could not determine Node version.

    To run any Node command, you need to set a project or default version.

      - Use `notion pin node` to select a version for a project.
      - Use `notion install node` to select a default version.
    ```

    (I'm not married to that!)

2. Perhaps our installation and setup process should prompt the user to learn *how* to start using Notion, so that if they happen to just run the installer, the installer itself tells them what to do next (without necessarily having the docs open). E.g.:

    ```sh
    $ curl -sSLf https://get.notionjs.com | bash
        Checking for existing Notion installation
        Fetching macOS installer
    ######################################################################## 100.0%
        Creating Notion directory tree (/Users/ckrycho/.notion)
       Unpacking `notion` executable and shims
         Editing user profile

        Finished installation. Open a new terminal to start using Notion!

             Tip Once you start a new terminal session, install a node version with
                 `notion install node`. For more on how to use Notion, check out
                 https://www.notionjs.com!
    ```

    (I'm not married to that, either.)

---

Also, the error report if you run `node --version` in this scenario is rather surprising:

```sh
$ node --version
Notion error: Could not determine Node version.

Use `notion pin node` to select a version for a project.
Use `notion install node` to select a default version.
```

This *looks* like there's a problem figuring out the Node version from the node command we just called; the `Notion error:` part *helps* with that but I can still see this misleading users. We may want to rephrase that first sentence a bit as well to help with that?

charlespierce:
Closes #363 

Updated error message for the `NoPlatform` case (when `node` is not set in either the project nor the user platform) to be less confusing and give a stronger CTA.

### DIFF
--- a/crates/volta-core/src/error/details.rs
+++ b/crates/volta-core/src/error/details.rs
@@ -707,10 +707,11 @@ Use `volta pin node` to pin Node first, then pin a Yarn version."
             ),
             ErrorDetails::NoPlatform => write!(
                 f,
-                "Could not determine Node version.
+                "Node is not available.
 
-Use `volta pin node` to select a version for a project.
-Use `volta install node` to select a default version."
+To run any Node command, first set a project or default version.
+  - Use `volta pin node` to select a version for a project.
+  - Use `volta install node` to select a default version."
             ),
             ErrorDetails::NoProjectYarn => write!(
                 f,
--- a/crates/volta-core/src/error/details.rs
+++ b/crates/volta-core/src/error/details.rs
@@ -709,9 +709,7 @@ Use `volta pin node` to pin Node first, then pin a Yarn version."
                 f,
                 "Node is not available.
 
-To run any Node command, first set a project or default version.
-  - Use `volta pin node` to select a version for a project.
-  - Use `volta install node` to select a default version."
+To run any Node command, first set a default version using `volta install node`"
             ),
             ErrorDetails::NoProjectYarn => write!(
                 f,


crates/volta-core/src/error/details.rs
```diff
@@ -707,10 +707,11 @@ Use `volta pin node` to pin Node first, then pin a Yarn version."
             ),
             ErrorDetails::NoPlatform => write!(
                 f,
-                "Could not determine Node version.
+                "Node is not available.
 
-Use `volta pin node` to select a version for a project.
-Use `volta install node` to select a default version."
+To run any Node command, first set a project or default version.
```
dherman: I'm finding "set a project or default version" pretty opaque. I think what you're trying to get across is that `volta pin node` will ensure the Node version is available as part of what it accomplishes. But I'm wondering if this is just far enough from what the user is trying to accomplish that it's worth eliminating the advice about `volta pin` altogether here?

IOW, if I'm trying to run Node and it's not available, probably the only real advice I need for now is to run `volta install node`. Selecting a Node version for a project is just kind of not in my critical path, even though it would happen to make a Node version available.

charlespierce: That's a good point, if the user just wants to run `node`, we can point them to how to do that, then if they want to learn more advanced stuff (like `volta pin`), we have docs and `volta help` to guide them. That'll also let us reduce this down to a single-line CTA, which is more in line with most of our other errors.

PR #419, opened 2019-05-10
which: look up user tools, including in projects.

### LINKED ISSUE #391: notion which is not aware of installed bins
`notion which` seems unaware of any `notion install`ed tools other than `node`/`yarn`/`npm`.

Steps to reproduce:

```
notion install node
notion install yarn
notion install ember-cli
```

```
❯ notion which node
/Users/rjackson/.notion/tools/image/node/12.1.0/6.9.0/bin/node

❯ notion which yarn
/Users/rjackson/.notion/tools/image/yarn/1.15.2/bin/yarn

❯ notion which ember

```

chriskrycho:
- Fixes #391.
- Includes some cleanup of the existing `which` code.

### DIFF
--- a/src/command/which.rs
+++ b/src/command/which.rs
@@ -53,7 +53,7 @@ impl Command for Which {
         // Treat any error with obtaining the current platform image as if the image doesn't exist
         // However, errors in obtaining the current working directory or the System path should
         // still be treated as errors.
-        let mut default_path = match session
+        let path = match session
             .current_platform()
             .unwrap_or(None)
             .and_then(|platform| platform.checkout(session).ok())
@@ -61,14 +61,6 @@ impl Command for Which {
         {
             Some(path) => path,
             None => System::path()?,
-        };
-
-        let path = match tool_path {
-            Some(tool_path) => {
-                default_path.push(tool_path.as_os_str());
-                default_path
-            }
-            None => default_path,
         };
 
         let cwd = env::current_dir().with_context(|_| ErrorDetails::CurrentDirError)?;
--- a/src/command/which.rs
+++ b/src/command/which.rs
@@ -23,7 +23,6 @@ impl Command for Which {
         // Treat any error with obtaining the current platform image as if the image doesn't exist
         // However, errors in obtaining the current working directory or the System path should
         // still be treated as errors.
-        let cwd = env::current_dir().with_context(|_| ErrorDetails::CurrentDirError)?;
         let path = match session
             .current_platform()
             .unwrap_or(None)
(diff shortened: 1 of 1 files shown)


src/command/which.rs
```diff
@@ -17,14 +18,42 @@ pub(crate) struct Which {
...
+        let project_bin_path =
+            session
+                .project()?
+                .and_then(|project| match project.has_direct_bin(&bin) {
+                    Ok(true) => Some(project.local_bin_dir()),
+                    _ => None,
+                });
+
+        let tool_path = match (user_tool, project_bin_path) {
+            (Some(_), Some(project_path)) => Some(project_path),
```
charlespierce: This looks like it will only show the directory that the project local bin is in (i.e. `.../node_modules/.bin`), but won't include the actual file binary file like the other returns from `which` do.

src/command/which.rs
```diff
@@ -34,20 +63,29 @@ impl Command for Which {
             None => System::path()?,
         };
 
-        match which_in(&self.binary, Some(path), cwd) {
+        let path = match tool_path {
+            Some(tool_path) => {
```
charlespierce: If `tool_path` is a `Some(_)` value, it appears this method will return on line 50, so we won't ever reach this branch of the `match`

src/command/which.rs
```diff
@@ -34,20 +63,29 @@ impl Command for Which {
...
+                default_path.push(tool_path.as_os_str());
```
charlespierce: If this is still needed, it's probably best to use the `tool_path` as the whole `path` for this case. We can't just `push` an extra path onto the end of the `PATH` string, because we need to take into account the separator character, which is platform-specific. Inside `System::path` and `Image::path`, we use the `envoy` crate to handle that for us, but in this case I think we'll know definitively that the binary will be found at `tool_path`, so we don't need to include any extra paths when calling `which_in`.

src/command/which.rs
chriskrycho: This whole path is extraneous; I totally forgot to rip it back out.

src/command/which.rs
chriskrycho: Interesting; this actually is simply failing to resolve correctly, and I'll need to dig further to figure out why.

src/command/which.rs
chriskrycho: Resolved in 12d7e6b0.

src/command/which.rs
charlespierce: It looks like it's working for me, apart from needing to `push` `&bin` to the end of the Path for the project local one (I locally updated `voyager-web` to use `volta` instead of `toolchain`):

```bash
$ volta which node
/Users/cpierce/.volta/tools/image/node/10.15.3/6.4.1/bin/node
$ volta which yarn
/Users/cpierce/.volta/tools/image/yarn/1.16.0/bin/yarn
$ volta which ember
/Users/cpierce/.volta/tools/image/packages/ember-cli/3.9.0/bin/ember
$ cd voyager-web
$ volta which node
/Users/cpierce/.volta/tools/image/node/10.6.0/6.1.0/bin/node
$ volta which yarn
/Users/cpierce/.volta/tools/image/yarn/1.13.0/bin/yarn
$ volta which ember
/Users/cpierce/code/voyager-web_trunk/node_modules/.bin
```

So all that's missing is adding `/ember` to the end of the path in the last case.

chriskrycho: Yeah, I just worked through the same; doing `.join(&bin)` resolves it nicely. Pushing that up now!

chriskrycho: Resolved in ba5cea4f!

PR #420, opened 2019-05-10
Don't parse dependencies when locating a local binary

### LINKED ISSUE #393: Avoid parsing dependency package.json's when searching for local bin to run
Currently (my understanding) is that when a binary is installed at the user/global level (e.g. `notion install ember-cli`), and then a bin script is invoked (e.g. `ember build`) within a project we do the following:

* Parse the project's own `package.json`
* Iterate the projects dependencies/devDependencies (should also do `optionalDependencies` but I don't know if we do?)
* Search each of those combined deps's `package.json` (e.g. `node_modules/package-name/package.json`) for bin scripts with the given name
* If found, use the local bin 
* If not found, use the globally installed package's bin script

Having to open and parse all of the dependencies/devDependencies's package.json's adds additional overhead to launching, and I believe it can be avoided completely.

I propose:

* Parse the project's own `package.json`
* Look for the specific installed package (e.g. `ember-cli` from my example) in the dependencies/devDependencies
* If found, use the local bin
* If not found, use the globally installed package's bin script

This relies on the fact that we keep track of which package a given globally installed bin shim is provided by, so we can directly compare/look for that instead of manually searching.

Changing to this system would have a few benefits:

* Lower overall overhead for globally installed shims
* Better resilience to malformed `package.json`'s (we no longer need to parse all of the dependencies/devDependencies package.json, only the projects itself and _possibly_ the local installed version of the global package)
* Better alignment with `PnP` (the manually parsing the `node_modules` tree will have to branch when in a PnP project vs when not in one, IMHO better to remove the branching)

This idea originally came from the discussion in discord when https://github.com/notion-cli/notion/issues/388 was reported.

charlespierce:
Closes #393 

Info
-----
As noted in the issue, we are currently parsing all of the dependency `package.json` files to find which `bins` are local, direct dependencies. However, since we have implemented `volta install <package>`, we have a bin config file that can tell us which package a binary came from.

Changes
-----
* Updated `Project::has_direct_bin` to read the `BinConfig` file and then check that the `package` for that bin is declared in the current project's `dependencies` or `devDependencies`, bypassing the need to parse all of the dependencies directly.
* Removed the `LazyDependentBins` struct and all of the code around generating that HashMap, since we don't use it anymore.

### DIFF
--- a/crates/volta-core/src/project.rs
+++ b/crates/volta-core/src/project.rs
@@ -1,7 +1,6 @@
 //! Provides the `Project` type, which represents a Node project tree in
 //! the filesystem.
 
-use std::collections::HashMap;
 use std::env;
 use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
@@ -11,8 +10,10 @@ use lazycell::LazyCell;
 use semver::Version;
 
 use crate::distro::node::{load_default_npm_version, NodeVersion};
+use crate::distro::package::BinConfig;
 use crate::error::ErrorDetails;
 use crate::manifest::{serial, Manifest};
+use crate::path;
 use crate::platform::PlatformSpec;
 use volta_fail::{throw, Fallible, ResultExt};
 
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/project.rs
```diff
@@ -133,43 +114,19 @@ impl Project {
...
-    /// Convert dependency names to the path to each project.
-    fn get_dependency_path(&self, name: &String) -> PathBuf {
-        // ISSUE(158): Add support for Yarn Plug'n'Play.
-        let mut path = PathBuf::from(&self.project_root);
-
-        path.push("node_modules");
-        path.push(name);
-
-        path
+    pub fn has_direct_dependency(&self, dependency: &str) -> bool {
```
chriskrycho: It *looks* like this is only `pub` so it can be tested. If so, you can actually drop the `pub` marker entirely: the submodule has access to its parent module's private items. You can see an example of how that works [in this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4d494347badb5a86bcbced056f00744a).

charlespierce: Good catch, updated!

PR #421, opened 2019-05-10
Tools: better message for intercepted global installs.

### LINKED ISSUE #360: Improve `yarn global add ember-cli` error message
Running `yarn global add ember-cli` produces:

```
Notion error: Global package installs are not recommended.

Use `notion install` to add a package to your toolchain (see `notion help install` for more info).
```

Couple of thoughts:
1. It seems like most errors start with just `error` and not `Notion error`. We should be consistent.
2. Should we change the wording of `error: Global package installs are not recommended.` to something stronger (since the is an error) `error: Global package installs via yarn are not allowed.`?
3. It would be great if the message about `Use 'notion install'` was changed to `Use 'notion install ember-cli'` so that users can copy and paste the correct command.

Encountered this with notion v0.3.0.

cc: @dherman @charlespierce @mikrostew

chriskrycho:
Fixes #360

<details>
<summary>
output for <code>yarn global add</code> and <code>npm -g, --global install</code> (all variations)
</summary>

```
$ yarn global add waffles
Volta error: Global package installs are not supported.

Use `volta install waffles` to add a package to your toolchain (see `volta help install` for more info).

$ yarn global add
Volta error: Global package installs are not supported.

Use `volta install` to add a package to your toolchain (see `volta help install` for more info).

$ npm i -g waffles
Volta error: Global package installs are not supported.

Use `volta install waffles` to add a package to your toolchain (see `volta help install` for more info).

$ npm add -g waffles
Volta error: Global package installs are not supported.

Use `volta install waffles` to add a package to your toolchain (see `volta help install` for more info).

$ npm add waffles --global
Volta error: Global package installs are not supported.

Use `volta install waffles` to add a package to your toolchain (see `volta help install` for more info).

$ npm isntall --global
Volta error: Global package installs are not supported.

Use `volta install` to add a package to your toolchain (see `volta help install` for more info).
```

</details>

### DIFF
--- a/crates/volta-core/src/tool/npm.rs
+++ b/crates/volta-core/src/tool/npm.rs
@@ -1,7 +1,7 @@
 use std::env::args_os;
 use std::ffi::{OsStr, OsString};
 
-use super::{intercept_global_installs, ToolCommand};
+use super::{intercept_global_installs, CommandArg, ToolCommand};
 use crate::error::ErrorDetails;
 use crate::session::{ActivityKind, Session};
 
@@ -15,8 +15,10 @@ where
 
     match session.current_platform()? {
         Some(ref platform) => {
-            if intercept_global_installs() && is_global_npm_install() {
-                throw!(ErrorDetails::NoGlobalInstalls);
+            if intercept_global_installs() {
+                if let CommandArg::GlobalAdd(package) = check_npm_install() {
+                    throw!(ErrorDetails::NoGlobalInstalls { package });
+                }
             }
             let image = platform.checkout(session)?;
             let path = image.path()?;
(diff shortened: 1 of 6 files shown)


crates/volta-core/src/tool/npm.rs
```diff
@@ -26,25 +28,33 @@ where
...
-        .skip_while(|arg| match arg.to_str() {
-            Some(arg) => arg.starts_with("-"),
-            None => false,
-        })
-        .next();
+fn check_npm_install() -> CommandArg {
+    // npm global installs will have `-g` or `--global` somewhere in the
+    // argument list
+    let is_global =
+        args_os().any(|arg| arg == OsString::from("-g") || arg == OsString::from("--global"));
```
mikrostew: It may be a premature optimization, but at this point you could check `is_global` and short-circuit the argument filtering and comparison if it's not a global command, instead of doing all of that and then checking `is_global`.

chriskrycho: Ah, good point.

chriskrycho: Resolved in 18856cdb; thanks!

PR #426, opened 2019-05-10
Improve error message when executing a project-local bin

charlespierce:
When attempting to execute a project-local tool (e.g. `ember`), if the binary doesn't exist (because, for example, the `node_modules` have not been installed), then the error message is misleading:

```
Could not execute command.

See `volta help install` and `volta help pin` for info about making tools available.
```

The problem, in this case, isn't that the tool hasn't been made available, but rather that executing it failed for some reason (most likely because the file doesn't exist). Update that to instead say:

```
Could not execute command `/path/to/project/node_modules/.bin/ember`

Please ensure that all project dependencies are installed.
```

Which gives information about what command was being run _and_ what the user can do to fix it.

Also removed the `DepPackageReadError`, as we no longer read the dependent package information, so it won't ever be thrown.

### DIFF
--- a/crates/volta-core/src/error/details.rs
+++ b/crates/volta-core/src/error/details.rs
@@ -27,6 +27,7 @@ pub enum ErrorDetails {
         new_package: String,
     },
 
+    /// Thrown when executing an external binary fails
     BinaryExecError,
 
     /// Thrown when a binary could not be found in the local inventory
@@ -84,9 +85,6 @@ pub enum ErrorDetails {
     DeleteFileError {
         file: String,
     },
-
-    /// Thrown when reading dependency package info fails
-    DepPackageReadError,
 
     DeprecatedCommandError {
         command: String,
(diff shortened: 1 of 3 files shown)


dherman:
Will this be the error even if the command runs but exits with a non zero code?

charlespierce:
No, if the command returns a non-zero exit code, then the `ExitStatus` returned in [here](https://github.com/notion-cli/notion/blob/master/crates/volta-core/src/tool/mod.rs#L262) will include the exit code. See also [the base executable](https://github.com/notion-cli/notion/blob/master/src/main.rs#L18) that unwraps it.

rwjblue:
If the file is missing, can we tell them that (instead of `could not execute command /foo/bar/node_modules/.bin/command-here`)?

Specific suggestion:

```
Could not execute `/path/to/project/node_modules/.bin/ember`, the path does not exist.

Please ensure that all project dependencies are installed (e.g. `npm install` / `yarn install`).
```

charlespierce:
@rwjblue Good idea! And we can actually short-circuit that case, since we can easily check that the file doesn't exist before attempting to shell out to it. Updated.

PR #443, opened 2019-05-17
Use owned PathBuf instead of String for file system errors

### LINKED ISSUE #381: Use owned PathBuf instead of String for File System Errors
Currently for our errors that involve a file / directory path, we are passing the path as a String to the error struct. This results in us calling `pathbuf.to_string_lossy().to_string()` in many different places. Instead, we should move the `PathBuf` object into the error, then we can do the formatting in one place and prevent a lot of code duplication.

This also opens up the potential for other manipulations of the `PathBuf` before writing the error message, if we find a use for that.

benblank:
Closes #381.

### DIFF
--- a/crates/volta-core/src/error/details.rs
+++ b/crates/volta-core/src/error/details.rs
@@ -18,6 +18,21 @@ an issue at https://github.com/volta-cli/volta/issues with the details!";
 const PERMISSIONS_CTA: &'static str =
     "Please ensure you have correct permissions to the Volta directory.";
 
+#[derive(Debug, PartialEq)]
+pub enum CreatePostscriptErrorPath {
+    Directory(PathBuf),
+    Unknown,
+}
+
+impl fmt::Display for CreatePostscriptErrorPath {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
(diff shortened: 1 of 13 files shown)


crates/volta-core/src/error/details.rs
```diff
@@ -502,7 +503,7 @@ for {}
                 "Could not create directory {}
 
 Please ensure that you have the correct permissions.",
-                dir
+                dir.to_string_lossy()
```
chriskrycho: We should probably prefer [`PathBuf::display()`](https://doc.rust-lang.org/1.34.1/std/path/struct.PathBuf.html#method.display) here (and throughout) instead, since it's explicitly designed for this:

> Returns an object that implements `Display` for safely printing paths that may contain non-Unicode data.

benblank: Done.

chriskrycho:
Note that if we did the custom enum approach, the call to create it could look like this:

```rust
write(path, self.compile_postscript(postscript).as_bytes()).with_context(|_| {
    ErrorDetails::CreatePostscriptError(
        path.parent().map_or(CreatePostscriptError::Unknown, |p| {
            CreatePostscriptError::Path(p.to_path_buf())
        }),
    )
})
```

benblank:
Great idea for CreatePostscriptError; done.

PR #446, opened 2019-05-18
Publish API docs from Azure Pipelines

### LINKED ISSUE #439: Publish docs through Azure Pipelines
Currently we're publishing docs via Travis; we need to publish via Azure so we can drop Travis and have everything run through Azure.

dherman:
Add a script to allow custom CI override directives to be specified in commit messages.

The first override is `[ci docs]`, which forces the publication of API docs in a pull request. This only works for pull requests filed from the main repo, because forked repos do not have permissions to publish docs.

Closes #439.


Cargo.toml
```diff
@@ -7,7 +7,7 @@ repository = "https://github.com/volta-cli/volta"
 edition = "2018"
 
 [features]
-universal-docs = ["volta-core/universal-docs"]
+cross-platform-docs = ["volta-core/cross-platform-docs"]
```
dherman: The `universal-docs` feature flag was kind of non-self-explanatory, so I rename it to `cross-platform-docs`, which I think makes it a little easier to guess the meaning of. TLDR it's a feature flag you enable to do a single build of the API docs that will include both Unix- and Windows-specific modules.

dherman:
@charlespierce Yeah, it's pretty well-tested: I added the `[ci docs]` functionality specifically to be able to test it out in a PR (for security reasons, it only works in a PR pushed from the main repo, not from a fork, but I made this PR from the main repo). I've actually probably rerun this through CI a couple hundred times until I got it all working 😝

PR #461, opened 2019-05-29
Implement `volta list` command

### LINKED ISSUE #349: `volta list` command
As a user who has installed a bunch of different default tools, I want to know what tools I have installed at what version, e.g:

```sh
$ volta list
Default tools:

    Tool               | Version   | Node Version   | Packager
    ================== | ========= | ============== | ===========
    ember-cli          | 3.9.0     | 11.14.0        | npm 6.7.0
    typescript         | 3.4.3     | 11.14.0        | npm 6.7.0
    ts-node            | 8.0.3     | 11.14.0        | npm 6.7.0
    yarn-deduplicate   | 1.1.1     | 11.14.0        | npm 6.7.0

```

(One can imagine that the <b>Packager</b> and <b>Node Version</b> fields will differ more in the future if packages begin adopting Volta.)

We have all the relevant data in the `tools/user/package` to do this, and we should be able to use something like [prettytable-rs](https://github.com/phsym/prettytable-rs) to do the formatting.

chriskrycho:
Implements volta-cli/rfcs#34. Resolves #349.

- [x] Implement CLI changes
    - [x] Create the `list` argument and define its implementation to support the full RFC
    - [x] Create an `ls` convenience alias for the `list` argument
    - [x] Add a deprecation notice when using `volta current`
- [x] Refactor `Inventory.packages` to support associate package names with the package versions
- [x] Actually load the user's *package* inventory into `Inventory.packages`
- [x] Load the user's inventory and project in the new list `run` method
- [x] Map each tool type into a type for representing a user's toolchain
- Create helpers to render the `human` and `plain` formats for each kind of item
    - runtimes
        - [ ] human – deferred to #523
        - [x] plain
    - packagers
        - [ ] human – deferred to #523
        - [x] plain
    - packages
        - [ ] human – deferred to #523
        - [x] plain
    - tools
        - [ ] human – deferred to #523
        - [x] plain
- [x] Actually print (the appropriate subset of) the user's inventory and project info


chriskrycho:
Follow-on work:

- Implement the rest of the `--format=human` printer. This is not large, it'll just need a bit of time: #523
- List packages which have been fetched but not installed: #524
- List *packages* which match packages set in the user's default when running `list all`: #525

src/command/list/toolchain.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        /// A convenient name for this tuple, since we have to name it in a few
+        /// spots below.
+        type Triple = (Kind, PackageConfig, Source);
+
+        let (packages_and_tools, errors): (Vec<Fallible<Triple>>, Vec<Fallible<Triple>>) =
+            inventory
+                .packages
+                .clone()
+                .into_iter()
```
charlespierce: Similar comment here about potentially providing an `iter` method that returns an Iterator of references.

src/command/list/toolchain.rs
```diff
@@ -0,0 +1,345 @@
...
+                        None
+                    }
+                })
+                .partition(|result| result.is_ok());
+
+        // If there are any errors, return the first one, since we don't have a
+        // notion of composite errors in our error handling at this point.
+        // (This should be quite unusual, as it means something failed in
+        // looking up a tool's bin directory.)
+        let first_error = errors.into_iter().map(|error| error.unwrap_err()).nth(0);
```
charlespierce: Is the goal here to eventually provide support for showing all of the errors? If not, it might make sense to short-circuit the processing on the first error, using the `collect` method to gather a `Fallible<Vec<Triple>>` and then `?` on the result:

```rust
let packages_and_tools = inventory
        .packages
        .clone()
        .into_iter()
        .filter_map(|config| {
            // Start with the package itself, since tools often match
            // the package name and we prioritize packages.
            if &config.name == name {
                let source = Package::source(name, &config.version, project);
                if source.allowed_with(filter) {
                    Some(Ok((Kind::Package, config, source)))
                } else {
                    None
                }

            // Then check if the passed name matches an installed package's
            // binaries. If it does, we have a tool.
            } else if config
                .bins
                .iter()
                .find(|bin| bin.as_str() == name)
                .is_some()
            {
                tool_source(name, &config.version, project)
                    .map(|source| {
                        if source.allowed_with(filter) {
                            Some((Kind::Tool, config, source))
                        } else {
                            None
                        }
                    })
                    .transpose()

            // Otherwise, we don't have any match all.
            } else {
                None
            }
        })
        .collect::<Fallible<Vec<Triple>>>()?;
```

src/command/list/plain.rs
```diff
@@ -0,0 +1,498 @@
...
+
+fn display_package_manager(package_manager: &PackageManager) -> String {
+    format!(
+        "package-manager {}{}",
+        tool_version(&package_manager.kind, &package_manager.version),
+        package_manager.source
+    )
+}
+
+fn package_source(package: &Package) -> String {
```
charlespierce: This seems like it duplicates the existing `impl Display for Source`, would it be possible to leverage that so that we don't have formatting code that needs to change in multiple places? Thinking it through, it's kinda tricky so if there's not a simple solution maybe something to keep in mind for a follow-on issue.

src/command/list/human.rs
```diff
@@ -0,0 +1,1016 @@
+//! Define the "plain" format style for list commands.
```
charlespierce: Should be "human" for this file :smiley:

```suggestion
//! Define the "human" format style for list commands.
```

crates/volta-core/src/inventory/mod.rs
```diff
@@ -1,15 +1,20 @@
...
 
-use crate::tool::{Node, Package, Tool, Yarn};
+use failure::ResultExt;
 use lazycell::LazyCell;
+use regex::Regex;
 use semver::Version;
 use volta_fail::Fallible;
 
-pub(crate) mod serial;
+use crate::{error::ErrorDetails, fs::read_dir_eager, version::VersionSpec};
```
mikrostew: Is this something that the formatter has started doing automatically? I see this style in a few places in this PR.

I like it visually, but I have avoided doing this in the past, because it makes rebasing harder. Separate PRs changing crate dependencies will be conflicting on this one line.

crates/volta-core/src/tool/package/install.rs
```diff
@@ -22,6 +22,9 @@ use log::debug;
 use semver::Version;
 use volta_fail::{throw, Fallible, ResultExt};
 
+// TODO: this may not really belong in the `install` module, since we now need
+//       to expose it *outside* this module for the sake of listing data about
+//       installed packages.
```
mikrostew: Sounds like a good idea - need an issue for that?

crates/volta-core/src/inventory/mod.rs
chriskrycho: Ah, that's interesting. I just started doing it myself—the formatter will leave whichever style you choose. If there's a preference by the rest of the team I'm happy to change it!

src/command/list/plain.rs
chriskrycho: It does indeed, though I don't think there's any good way to eliminate the duplication at this point, because the types are distinct. 🤔 I've opened #527 to serve as an investigation issue around it, though.

src/command/list/toolchain.rs
chriskrycho: Done!

src/command/list/toolchain.rs
chriskrycho: Great suggestion! I didn't think of just `collect`-ing into a `Fallible<Vec<Triple>>`, so this works perfectly I think. I love it—lets us get rid of a lot of code.

src/command/list/toolchain.rs
```diff
@@ -0,0 +1,345 @@
...
+        if let Some(error) = first_error {
+            return Err(error);
+        }
+
+        let packages_and_tools: Vec<Triple> = packages_and_tools
+            .into_iter()
+            .map(|ok| ok.unwrap())
+            .collect();
+
+        let has_packages = packages_and_tools
```
chriskrycho: When fixing the above, I also realized that I could simplify this; it cleans up really nicely if I use `fold` and a `(bool, bool)` instead of iterating over the `Vec` twice.

crates/volta-core/src/platform/mod.rs
```diff
@@ -16,7 +16,7 @@ pub mod sourced;
 pub use self::sourced::{Source, SourcedImage, SourcedPlatformSpec};
 
 /// A specification of tool versions needed for a platform
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
charlespierce: That's a reasonable thought, and sort of where I was leading. My one question is does the `derive`d implementation depend on the order that the fields are declared in within the struct? I would imagine it has to, and given that we should probably have a comment that makes it clear not to re-order them, since we are relying on the ordering being first by name, then by version.

src/command/list/plain.rs
charlespierce: Yeah, I spent a few minutes trying to come up with something and there wasn't any obvious solution to prevent the duplication without introducing weird structures elsewhere, so I think the follow-on issue makes sense.

src/command/list/toolchain.rs
charlespierce: :+1:

crates/volta-core/src/platform/mod.rs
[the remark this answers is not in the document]
chriskrycho: I'm actually not positive. We can dig in, for sure; it's worth an investigation.

PR #464, opened 2019-06-03
Build an rpm package

### LINKED ISSUE #15: publish rpms
We should publish Linux installers as rpms.

mikrostew:
Closes #15.

## Building the RPM

This introduces the `build-rpm.sh` script, which builds an rpm package for Volta:

```
./dev/rpm/build-rpm.sh 0.5.3
```
This follows the [RPM Packaging Guide](https://rpm-packaging-guide.github.io/), which was very helpful getting this to work. It packages up the source code, runs `cargo build --release`, and packages those compiled binaries (along with the `load.*` shell integration files and the `volta-postinstall.sh` script) into an rpm.

## How the RPM Install Works

Due to the way that rpms are packaged, it can't install binaries directly to the HOME dir. It must install to known locations in the [Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). So this installs the binaries (and other scripts) to `/usr/bin/volta/`, so they are in a known location but not directly on the PATH.

So the `volta-postinstall.sh` script takes care of what the normal installer would do:
* create the home directory layout
* create symlinks and shims
* update the user's profile.

AND it needs to do 2 extra things due to where the binaries are installed:
* create symlinks to the binaries in /usr/bin/volta/*
* copy over the shell integration files (load.*) to the HOME directory


dev/rpm/build-rpm.sh
```diff
@@ -0,0 +1,46 @@
...
+rpmdev-setuptree
+
+# create a tarball of the repo for the specified version
+# using prefix because the rpmbuild process expects a 'volta-<version>' directory
+# (https://rpm-packaging-guide.github.io/#putting-source-code-into-tarball)
+git archive --format=tar.gz --output=$archive_filename --prefix="volta-${release_version}/" "v${release_version}"
+
+# move the archive to the SOURCES dir
+# (https://rpm-packaging-guide.github.io/#working-with-spec-files)
+mv "$archive_filename" "$HOME/rpmbuild/SOURCES/"
```
charlespierce: Does `rmpdev-setuptree` clean out anything in these directories as well? If not, we might want to do that just to make sure we don't get contamination from old builds.

mikrostew: Good call, I just tested and it does not clean out those dirs. I'll add some code to clean those out.

Issue #436, opened 2019-05-15
Volta error: No Yarn version found in this project.


thoov:
Steps to reproduce:

1. `volta install yarn`
2. `volta pin node` (node is the only thing pinned)
3. `yarn -v`

Result:
```
Volta error: No Yarn version found in this project.

Use `volta pin yarn` to select a version (see `volta help pin` for more info).
```

rwjblue:
@thoov - Is there a `yarn` version setup in the system in general? 

Specifically, I'm wondering if the error reported here is:

```
# fresh install

curl -sSLf https://get.volta.sh | bash
volta pin node
yarn -v
```

Or:

```
# fresh install
curl -sSLf https://get.volta.sh | bash

volta install yarn
volta pin node
yarn -v
```

rwjblue:
Also, what was your expected output? Command not found?

rwjblue:
I just did a quick test, looks like both of those snippets results in the same error. I was surprised that the second example failed, as I'd expected merging behavior similar to what we do with `hooks.json`.

At worst, we should have more specific prose in the error message, but I think we should probably "just work" (setting up a `yarn` that uses the pinned node if needed) in this case.

thoov:
@rwjblue I updated the repro steps to include the extra `volta install yarn`

charlespierce:
This might actually lead to a slightly different question: Should we treat the platform as a merge (in the way that we do with `hooks.json` currently) between the user defaults and the project-local settings? It's not immediately obvious to me either way.

One issue with doing so is that you could have a `yarn` default set up on your machine, and build your project assuming `yarn` is there, because it always works on your machine. But then when a different user who doesn't have `yarn` set up tries to run the project, it doesn't work. That's sort of the behavior we are trying to prevent, but at the same time we've moved away from being quite so overbearing with e.g. command passthroughs.

So I lean towards merging from a UX perspective, with the philosophy that Volta _enables_ you to do things in a cleaner way, but doesn't _require_ that you do so.

rwjblue:
Yes, exactly. It might even be good to issue a warning that we've found `volta` in your `package.json` with a `node` version config, and that you are using `yarn` but it isn't configured in your project. I do agree that this type of mismatch is easy to get wrong but I think that having merging behavior (with or without a warning when we hit the fallback) is probably the right call.

PR #466, opened 2019-06-10
Add verbose messages to tool shims

charlespierce:
Info
-----
* The initial wave of verbose messages only covered the `volta` executable commands, not the behavior of the shims.
* To support diagnostic messages on shims and the `volta list` command, the `Session::current_platform` method should return the source that the platform came from as well as the platform itself.
* This will allow us to avoid duplication of logic checking for `project_platform` and then `user_platform` separately.
* Additionally, when we implement a fix for #436 and have the platforms merged, we will want to make sure that merging logic is centralized in one location and not duplicated.

Changes
-----
* Added `Source`, `SourcedPlatformSpec`, and `SourcedImage` to track the source of a platform / image (`Project` or `User`).
* Cleaned up logic in the tool `command` functions that duplicated the `current_platform` lookup to instead use the source property.
* Added `debug!` messages to all of the tool implementations to include more information about the platform and files that Volta is using.

Screenshots
-----
<details><summary>Node Outside Project</summary>

Node Outside Project
</details>

<details><summary>Node In Project</summary>

Node In Project
</details>

<details><summary>npm Outside Project</summary>

npm Outside Project
</details>

<details><summary>npm In Project</summary>

npm In Project
</details>

<details><summary>Yarn Outside Project</summary>

Yarn Outside Project
</details>

<details><summary>Yarn In Project</summary>

Yarn In Project
</details>

<details><summary>Binary Outside Project</summary>

Binary Outside Project
</details>

<details><summary>Binary In Project</summary>

Binary In Project
</details>

### DIFF
--- a/crates/volta-core/src/session.rs
+++ b/crates/volta-core/src/session.rs
@@ -109,12 +109,12 @@ impl Session {
     }
 
     pub fn current_platform(&self) -> Fallible<Option<SourcedPlatformSpec>> {
-        match self.project_platform()? {
-            Some(platform) => Ok(Some(SourcedPlatformSpec::project(platform))),
-            None => {
-                let user = self.user_platform()?;
-                Ok(user.map(SourcedPlatformSpec::user))
-            }
+        if let Some(platform) = self.project_platform()? {
+            Ok(Some(SourcedPlatformSpec::project(platform)))
+        } else if let Some(platform) = self.user_platform()? {
+            Ok(Some(SourcedPlatformSpec::user(platform)))
+        } else {
+            Ok(None)
         }
     }
 
--- a/crates/volta-core/src/session.rs
+++ b/crates/volta-core/src/session.rs
@@ -14,7 +14,7 @@ use crate::error::ErrorDetails;
 use crate::event::EventLog;
 use crate::hook::{HookConfig, LazyHookConfig, Publish};
 use crate::inventory::{FetchResolve, Inventory, LazyInventory};
-use crate::platform::PlatformSpec;
+use crate::platform::{PlatformSpec, SourcedPlatformSpec};
 use crate::project::{LazyProject, Project};
 use crate::style::{success_prefix, tool_version};
 use crate::toolchain::LazyToolchain;
@@ -108,10 +108,13 @@ impl Session {
         self.project.get()
     }
 
-    pub fn current_platform(&self) -> Fallible<Option<Rc<PlatformSpec>>> {
+    pub fn current_platform(&self) -> Fallible<Option<SourcedPlatformSpec>> {
         match self.project_platform()? {
-            Some(platform) => Ok(Some(platform)),
-            None => self.user_platform(),
+            Some(platform) => Ok(Some(SourcedPlatformSpec::project(platform))),
+            None => {
+                let user = self.user_platform()?;
+                Ok(user.map(|platform| SourcedPlatformSpec::user(platform)))
+            }
         }
     }
 
--- a/crates/volta-core/src/session.rs
+++ b/crates/volta-core/src/session.rs
@@ -113,7 +113,7 @@ impl Session {
             Some(platform) => Ok(Some(SourcedPlatformSpec::project(platform))),
             None => {
                 let user = self.user_platform()?;
-                Ok(user.map(|platform| SourcedPlatformSpec::user(platform)))
+                Ok(user.map(SourcedPlatformSpec::user))
             }
         }
     }
--- a/crates/volta-core/src/tool/binary.rs
+++ b/crates/volta-core/src/tool/binary.rs
@@ -29,16 +29,20 @@ where
                 });
             }
 
-            debug!("Found bin in project at {}", path_to_bin.display());
+            debug!(
+                "Found {} in project at '{}'",
+                exe.to_string_lossy(),
+                path_to_bin.display()
+            );
             let path_to_bin = path_to_bin.as_os_str();
 
             if let Some(platform) = session.current_platform()? {
                 match platform.source() {
                     Source::Project => {
-                        debug!("Using node@{} from project platform", platform.node())
+                        debug!("Using node@{} from project configuration", platform.node())
                     }
                     Source::User => {
-                        debug!("Using node@{} from user default platform", platform.node())
+                        debug!("Using node@{} from default configuration", platform.node())
                     }
                 };
 
@@ -48,16 +52,20 @@ where
             }
 
             // if there's no platform available, pass through to existing PATH.
-            debug!("Could not find platform, delegating to system");
+            debug!("Could not find Volta configuration, delegating to system");
             return ToolCommand::passthrough(&path_to_bin, args, ErrorDetails::NoPlatform);
         }
     }
 
     // try to use the user toolchain
     if let Some(user_tool) = session.get_user_tool(&exe)? {
-        debug!("Found user default bin in {}", user_tool.bin_path.display());
         debug!(
-            "Using node@{} from bin platform",
+            "Found default {} in '{}'",
+            exe.to_string_lossy(),
+            user_tool.bin_path.display()
+        );
+        debug!(
+            "Using node@{} from binary configuration",
             user_tool.image.node.runtime
         );
 
--- a/crates/volta-core/src/tool/binary.rs
+++ b/crates/volta-core/src/tool/binary.rs
@@ -90,7 +90,7 @@ where
     // at this point, there is no project or user toolchain
     // Pass through to the existing PATH
     debug!(
-        "Could not find '{}', delegating to system",
+        "Could not find {}, delegating to system",
         exe.to_string_lossy()
     );
     ToolCommand::passthrough(
--- a/crates/volta-core/src/tool/binary.rs
+++ b/crates/volta-core/src/tool/binary.rs
@@ -3,8 +3,8 @@ use std::iter::once;
 
 use super::ToolCommand;
 use crate::error::ErrorDetails;
+use crate::platform::Source;
 use crate::session::{ActivityKind, Session};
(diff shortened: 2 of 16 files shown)


crates/volta-core/src/tool/binary.rs
```diff
@@ -27,29 +29,38 @@ where
                 });
             }
 
+            debug!("Found bin in project at {}", path_to_bin.display());
```
rwjblue: Should we say "Found a bin for {} in project at {}" where the first substitution is the name of the bin we are looking for? 

All of the messages in this file _might_ want the same treatment?

crates/volta-core/src/tool/node.rs
```diff
@@ -13,11 +15,19 @@ where
...
+                Source::User => debug!("Using node@{} from user default platform", platform.node()),
+            };
+
             let image = platform.checkout(session)?;
             let path = image.path()?;
             Ok(ToolCommand::direct(OsStr::new("node"), args, &path))
         }
-        None => ToolCommand::passthrough(OsStr::new("node"), args, ErrorDetails::NoPlatform),
+        None => {
+            debug!("Could not find platform, delegating to system");
```
rwjblue: Probably should include what we are looking for (AFAICT it is `node`, right?).

charlespierce: That's a good point, all of the "No platform" messages could be updated to be more specific. Maybe "Could not find volta-managed node, delegating to system"?

crates/volta-core/src/tool/binary.rs
charlespierce: That's a good idea, I initially omitted it because the `path_to_bin` will have e.g. `/ember` as the end of it, but I think it probably reads better if the command is included separately.

charlespierce:
@rwjblue I used "platform", because that's the internal name for the block of Node and Optional Yarn that can be specified for running a command. It can come from the user defaults (when you run `volta install node` it sets that platform), from the project itself with the `volta` key in `package.json`, or from the tool settings when you do `volta install ember-cli` (since those are pinned to allow the tools to always run).

We probably don't need to use that language, however, we can use "configuration" or something like that:

```
Using node@10.15.0 from project configuration
Using yarn@1.16.0 from default configuration
Using node@12.1.0 from binary configuration
```

crates/volta-core/src/tool/node.rs
rwjblue: Ya, that sounds great

crates/volta-core/src/tool/binary.rs
rwjblue: Agree, also (not sure if this is true already) we'd want to quote the path and ensure it doesn't get wrapped (wasn't there a util for that already?)

rwjblue:
Ya, `configuration` definitely seems more user friendly than `platform` to me. But I'm also confused about all the different combinations of "configurations". The list above (from the messages in this PR) _seems_ to indicate that there are 4 different platform/configuration types, and AFAICT that is not true.

charlespierce:
I don't believe we have any messages about `system platform` (if we do that's definitely wrong, because that's not a concept we have. The other three are all valid, there's the user defaults, the project-local, and for global tools, we pin the version of `node` that we use to run them (bin platform). That allows the global tools to continue to work, even if you switch default versions of Node, without having to re-install the tool.

crates/volta-core/src/tool/binary.rs
charlespierce: There isn't any custom wrapping style applied to the debug output at the moment, so the only wrapping that would happen would be applied by the terminal itself. If / when we add that behavior, we'll definitely use the same logic we already have for the Warning messages that prevents it from wrapping paths on hyphen characters.

crates/volta-core/src/session.rs
```diff
@@ -108,10 +108,13 @@ impl Session {
         self.project.get()
     }
 
-    pub fn current_platform(&self) -> Fallible<Option<Rc<PlatformSpec>>> {
+    pub fn current_platform(&self) -> Fallible<Option<SourcedPlatformSpec>> {
         match self.project_platform()? {
-            Some(platform) => Ok(Some(platform)),
-            None => self.user_platform(),
+            Some(platform) => Ok(Some(SourcedPlatformSpec::project(platform))),
+            None => {
```
chriskrycho: *Certainly* not a blocker, but you can simplify this a bit:

```rust
Ok(user.map(SourcedPlatformSpec::user))
```

You can also reduce it into a single expression if you so desire:

```rust
None => self
    .user_platform()
    .map(|platform| platform.map(SourcedPlatformSpec::user)),
```

The form I actually find *clearest*, though, is reworking the whole thing to make the decision tree clearer:

```rust
match (self.project_platform()?, self.user_platform()?) {
    (Some(platform), _) => Ok(Some(SourcedPlatformSpec::project(platform))),
    (_, Some(platform)) => Ok(Some(SourcedPlatformSpec::user(platform))),
    _ => Ok(None),
}
```

crates/volta-core/src/tool/node.rs
```diff
@@ -13,11 +15,21 @@ where
     session.add_event_start(ActivityKind::Node);
 
     match session.current_platform()? {
-        Some(ref platform) => {
+        Some(platform) => {
+            match platform.source() {
+                Source::Project => {
+                    debug!("Using node@{} from project configuration", platform.node())
```
chriskrycho: Nit: it'd be nice to use the `style::tool_version` formatter here:

```rust
debug!("Using {} from project configuration", tool_version("node", platform.node()));
```

As an aside, I'd be inclined to do this to avoid having to tweak the messages separately if we reword in the future:

```rust
let source = match platform.source() {
    Source::Project => "project",
    Source::User => "default",
};
let version = tool_version("node", platform.node());
debug!("Using {} from {} configuration", version, source);
```

crates/volta-core/src/tool/npm.rs
```diff
@@ -14,17 +16,28 @@ where
...
                 if let CommandArg::GlobalAdd(package) = check_npm_install() {
                     throw!(ErrorDetails::NoGlobalInstalls { package });
                 }
             }
             let image = platform.checkout(session)?;
             let path = image.path()?;
+
+            match image.source() {
+                Source::Project => {
+                    debug!("Using npm@{} from project configuration", image.node().npm)
```
chriskrycho: Same suggestions as above here.

crates/volta-core/src/tool/npx.rs
```diff
@@ -14,22 +16,34 @@ where
...
             let image = platform.checkout(session)?;
 
             // npx was only included with npm 5.2.0 and higher. If the npm version is less than that, we
             // should include a helpful error message
             let required_npm = VersionSpec::parse_version("5.2.0")?;
-            if image.node.npm >= required_npm {
+            if image.node().npm >= required_npm {
+                match image.source() {
+                    Source::Project => {
+                        debug!("Using npx@{} from project configuration", image.node().npm)
```
chriskrycho: And again!

crates/volta-core/src/tool/yarn.rs
```diff
@@ -16,40 +16,48 @@ where
...
             if intercept_global_installs() {
                 if let CommandArg::GlobalAdd(package) = check_yarn_add() {
                     throw!(ErrorDetails::NoGlobalInstalls { package });
                 }
             }
 
+            // Note: If we've gotten this far, we know there is a yarn version set
+            match platform.source() {
+                Source::Project => debug!(
+                    "Using yarn@{} from project configuration",
```
chriskrycho: And again. 😀

crates/volta-core/src/session.rs
charlespierce: I did consider the last of those, and I agree it makes the decision tree clearer, the issue I had with it is that it means the user platform file needs to be read and parsed for every invocation of a tool, even if everything is specified in the Project settings. With the current approach as well as the old code, the user platform isn't looked at unless it's needed.

This may be an early optimization, but especially when adding the verbose messaging, keeping the file access the same seemed important.

I also find the single-expression form slightly harder to grok, with a map within a map, but I like the improvement of using point-free style for the map function.

crates/volta-core/src/tool/node.rs
charlespierce: Good call in both cases! Updated!

crates/volta-core/src/tool/npx.rs
charlespierce: Updated!

crates/volta-core/src/tool/yarn.rs
charlespierce: Updated!

crates/volta-core/src/tool/npm.rs
charlespierce: Updated!

crates/volta-core/src/session.rs
chriskrycho: Totally makes sense. The fact that it's doing a file lookup *does* make a difference here, and I'm inclined to agree with you.

Another alternative I considered here, which is the longest but may be the next-best option to preserve that *and* keep the decision tree equally clear:

```rust
if let Some(platform) = self.project_platform()? {
    Ok(Some(SourcedPlatformSpec::project(platform)))
} else if let Some(platform) = self.user_platform()? {
    Ok(Some(SourcedPlatformSpec::user(platform)))
} else {
    Ok(None)
}
```

At the end of the day, any of these are fine, but I did find the flow a little hard to follow in the original `match`, which is why I mentioned it!

charlespierce: That’s a good idea, and makes the changes in the follow-on PR (#470) a little more explicit, so I think I’ll look into that! Thanks!

PR #468, opened 2019-06-11
Support relative paths for Bin hooks

### LINKED ISSUE #459: Bin Hooks should allow relative paths
Currently, if a `bin` hook is specified in `hooks.json` (either per-user or per-project), that command will be executed exactly. That means that if you specify a relative path, it will be executed relative to the current working directory you are running in, not relative to the file it came from.

We should update it so that paths are relative to the `hooks.json` file they are specified in, rather than the directory you happen to be executing from.

charlespierce:
Closes #459 

Info
-----
* The Hooks configuration has an option `bin` that allows you to specify a command to run to resolve the hook.
* Currently, that command is run relative to the current directory, so if you specify a relative path in `hooks.json`, it will only work if you are in the correct directory.
* The easiest way to specify hooks will be with a relative path to the script, so we should support this use-case.

Changes
-----
* Updated the hooks deserialization to propagate the base directory that the `hooks.json` file is found in to any `Bin` hooks.
* If the command is detected as being a relative path (if it begins with `.`), then we execute it relative to the base directory found above.
* Added detection for non-zero exit codes, so that if the hook command returns a failure code, we throw a Volta error instead of continuing as if the command succeeded.
* Also updated `stderr` to inherit from the parent process, so that bin hooks can use it to show progress bars or waiting spinners, as well as report errors directly to the user.

Tested
-----
* Verified that a bin hook with a relative path correctly executes anywhere in a project, as opposed to only from a single directory.

### DIFF
--- a/crates/volta-core/src/hook/tool.rs
+++ b/crates/volta-core/src/hook/tool.rs
@@ -1,6 +1,7 @@
 //! Types representing Volta Tool Hooks.
 
 use std::ffi::OsString;
+use std::path::{Path, PathBuf};
 use std::process::Stdio;
 
 use crate::command::create_command;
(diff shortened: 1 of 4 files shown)


crates/volta-core/src/hook/tool.rs
```diff
@@ -43,56 +46,73 @@ impl DistroHook {
...
-    let cmd = if let Some(word) = words.next() {
-        word
-    } else {
-        throw!(ErrorDetails::InvalidHookCommand {
-            command: String::from(bin.trim()),
-        })
+    let cmd = match words.next() {
+        Some(word) => {
+            // Treat any path that starts with a '.' as a relative path
+            if word.starts_with('.') {
```
chriskrycho: I'm concerned about this implementation for the reason that, though it's not advised, you *could* have an item somewhere in your `PATH` which is *named* `.foo` and that's what you're trying to run. Should we make this check be against (the OS-equivalent of) `"./"` instead of `'.'`?

```rust
lazy_static! {
    static ref REL_PATH: String = format!(".{}", std::path::MAIN_SEPARATOR);
}

if word.starts_with(REL_PATH) {
  // ...
}
```

charlespierce: Hmm, that's true, but we also need to account for a path that starts with `../` (or the OS-equivalent). We can pretty easily check both, but now I'm curious if there are any other corner cases we need to consider.

chriskrycho: Ah, yep, good point.

charlespierce: I couldn't think of any other edge cases, so I think we're okay just checking for the two types of relative path: `./` and `../`, normalized to the OS path separator.

chriskrycho: Very good!

PR #470, opened 2019-06-12
Fall back to default Yarn if project Yarn is not set

### LINKED ISSUE #436: Volta error: No Yarn version found in this project.

charlespierce:
Closes #436 

Info
-----
* Currently, if we have a default yarn set, but the user runs `yarn` within a project that _doesn't_ have yarn specified, then we throw an error and require the user to run `volta pin yarn` to continue.
* This is a poor UX, as the user has already run `volta install yarn`, so we do have a Yarn version available, we are just not using it.
* Additionally, with the command passthrough and other UX improvements, we have taken the approach of setting Volta up to _support_ doing things the "safe" way (e.g. if you use `yarn` in your project, you should specify it there), but not _require_ that they are done that way.
* To align with that goal, we should fall back to the default yarn, even in a project that doesn't have `yarn` specified.

Changes
-----
* Added an additional `platform::Source`: `ProjectNodeUserYarn`, that represents a merged platform.
* Implemented the fall-back logic in `session::current_platform`, so that the fallback is available to any tool (as running other tools could result in running `yarn`, e.g. `ember new --yarn`).
* Updated the tool implementations to account for the new source.

Tested
-----
* Added new acceptance tests to cover the new behavior.

### DIFF
--- a/crates/volta-core/src/platform/sourced.rs
+++ b/crates/volta-core/src/platform/sourced.rs
@@ -8,8 +8,14 @@ use semver::Version;
 use volta_fail::Fallible;
 
 pub enum Source {
+    /// Represents a Platform that came from the user default
     User,
+
+    /// Represents a Platform that came from a project manifest
     Project,
+
+    /// Represents a Platform that is the result of merging the User and Project platforms
+    ProjectNodeUserYarn,
 }
 
 pub struct SourcedPlatformSpec {
@@ -34,6 +40,13 @@ impl SourcedPlatformSpec {
         SourcedPlatformSpec {
             platform,
             source: Source::User,
+        }
+    }
+
+    pub fn merged(platform: Rc<PlatformSpec>) -> Self {
+        SourcedPlatformSpec {
+            platform,
+            source: Source::ProjectNodeUserYarn,
         }
     }
 
--- a/crates/volta-core/src/platform/sourced.rs
+++ b/crates/volta-core/src/platform/sourced.rs
@@ -9,13 +9,13 @@ use volta_fail::Fallible;
 
 pub enum Source {
     /// Represents a Platform that came from the user default
-    User,
+    Default,
 
     /// Represents a Platform that came from a project manifest
     Project,
 
-    /// Represents a Platform that is the result of merging the User and Project platforms
-    ProjectNodeUserYarn,
+    /// Represents a Platform that is the result of merging the Default and Project platforms
+    ProjectNodeDefaultYarn,
 }
 
 pub struct SourcedPlatformSpec {
@@ -36,17 +36,17 @@ impl SourcedPlatformSpec {
         }
     }
 
-    pub fn user(platform: Rc<PlatformSpec>) -> Self {
+    pub fn default(platform: Rc<PlatformSpec>) -> Self {
         SourcedPlatformSpec {
             platform,
-            source: Source::User,
+            source: Source::Default,
         }
     }
 
     pub fn merged(platform: Rc<PlatformSpec>) -> Self {
         SourcedPlatformSpec {
             platform,
-            source: Source::ProjectNodeUserYarn,
+            source: Source::ProjectNodeDefaultYarn,
         }
     }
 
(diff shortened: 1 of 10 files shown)


chriskrycho:
Is this ready to go? If so, I’ll review it and we can merge it! Sorry for the delay; I missed seeing it was open!

charlespierce:
Yeah, sorry I think I might have forgotten to mention it was out of draft state. Thanks!

rwjblue:
FWIW, I think GH needs to get better about notifications here. I don't think they send another notification when going from "Draft" to "Ready to review".

chriskrycho:
CONFIRM, @rwjblue. I'm reviewing now!

crates/volta-core/src/platform/sourced.rs
```diff
@@ -8,8 +8,14 @@ use semver::Version;
...
 
 pub enum Source {
+    /// Represents a Platform that came from the user default
     User,
+
+    /// Represents a Platform that came from a project manifest
     Project,
+
+    /// Represents a Platform that is the result of merging the User and Project platforms
+    ProjectNodeUserYarn,
```
chriskrycho: Do we need to support the *inverse* of this? I.e. `ProjectYarnUserNode` seems like a thing that might exist, even if we wouldn't recommend it. In that case, it seems like we might want to restructure this slightly (perhaps in a follow-on PR; I can open an issue) to differentiate the source of *each tool* rather than the source of the whole platform. As it turns out, that's also [the tack I've ended up needing to take](https://github.com/volta-cli/volta/blob/c56fcab253cece17772a62a955f9a1a68cd1ae27/src/command/list/mod.rs#L54:L77) simply to be able to accurately *describe* the user's toolchain in `volta list`.

crates/volta-core/src/platform/sourced.rs
```diff
@@ -8,8 +8,14 @@ use semver::Version;
 use volta_fail::Fallible;
 
 pub enum Source {
+    /// Represents a Platform that came from the user default
```
chriskrycho: I'm wondering: given how our terminology has standardized, should we consider renaming `User` to `Default` throughout, since both project tools and default tools are part of the user toolchain?

charlespierce: Definitely agree here! I'll update this list and also create a follow-on issue to look at cleaning that naming up throughout the code base.

crates/volta-core/src/platform/sourced.rs
charlespierce: We don't (currently) need to support`ProjectYarnUserNode`, as we require the `node` setting to be the start of any platform. So if a project has `yarn`, then it must necessarily have `node` set as well (or we would have returned an error on the parse).

That said, I agree that having the source of each tool separately makes sense, and I actually started with the code organized that way. This PR, along with #466, were the result of thinking about the different issues (this one, the verbose logging for tools, and the work you're doing on `volta list`) and realizing that there was a need for some sort of "Source" attached to our `PlatformSpec` object. I spent an inordinate amount of time trying to work through a bunch of conceptual and technical hurdles, and while I'm definitely not thrilled with the final result, it was the least bad that I could come up with.

That said, I'm open to doing some more brainstorming on how to track / propagate the source through the data structures in a cleaner way, perhaps we can touch on that in the open call today, or chat on Discord.

chriskrycho: Least bad is something with which I sympathize deeply and it's much how I feel about this space as well. Do let's chat; but let's go ahead and land this as is!

charlespierce:
Created https://github.com/volta-cli/volta/projects/6#card-23308438 and https://github.com/volta-cli/volta/projects/6#card-23308348 to track possible future refactors brought up in PR Review.

PR #479, opened 2019-06-19
Use sort_by to preserve user-defined order of package installs

### LINKED ISSUE #478: `volta install` shouldn't change the order of packages
From @rwjblue on Discord: When installing multiple packages with `volta install`, we shouldn't change the order of the packages from what the user specified. Currently we are sorting them alphabetically:

```bash
$ volta install ember-cli yarn-deduplicate sort-package-json release-it
success: installed ember-cli@3.10.1 with executables: ember
success: installed release-it@12.3.0 with executables: release-it
success: installed sort-package-json@1.22.1 with executables: sort-package-json
success: installed yarn-deduplicate@1.1.1 with executables: yarn-deduplicate
```

We _do_ still want to pull Node & Yarn (and eventually npm) up to the front of the list, however within the packages, we should maintain the order that the user specified, to match expectations of how `npm install` / `yarn add` work.

charlespierce:
Closes #478 

Use `sort_by` and specify an ordering function to do the sorting of tools. That sorting function explicitly treats Node as the lowest, followed by Npm, Yarn, and Packages, but within each category it ignores the specifics and treats all tools as the same.

Per [the rust docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by), `Vec::sort_by` is stable, so by treating all of the Packages as equal, the original order (provided by the user) will be preserved.

Note: We use `sort_by` as opposed to a custom implementation of the `Ord` trait because the docs specify that the implementations of `Ord` and `PartialEq` must match exactly, and we use the derived `PartialEq` implementation elsewhere so it can't be changed.

### DIFF
--- a/crates/volta-core/src/tool/mod.rs
+++ b/crates/volta-core/src/tool/mod.rs
@@ -1,5 +1,6 @@
 //! Traits and types for executing command-line tools.
 
+use std::cmp::Ordering;
 use std::env::{self, args_os, ArgsOs};
 use std::ffi::{OsStr, OsString};
 use std::fmt::{self, Debug, Display, Formatter};
@@ -51,7 +52,7 @@ enum CommandArg {
 /// dealing with multiple tools.
 ///
 /// [`Ord`]: https://doc.rust-lang.org/1.34.0/core/cmp/trait.Ord.html
-#[derive(PartialEq, Eq, PartialOrd, Ord)]
+#[derive(PartialEq)]
 pub enum ToolSpec {
     Node(VersionSpec),
     Npm(VersionSpec),
@@ -157,7 +158,7 @@ impl ToolSpec {
             .map(|arg| Self::try_from_str(arg.as_ref()))
             .collect::<Fallible<Vec<ToolSpec>>>()?;
 
-        tools.sort();
+        tools.sort_by(tool_spec_sort_comparator);
         Ok(tools)
     }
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you for working on this! The current behavior confused me when I first noticed it.

crates/volta-core/src/tool/mod.rs
```diff
@@ -193,6 +194,24 @@ impl ToolSpec {
     }
 }
 
+/// Function for comparing toolspecs for use in sorting.
+/// We want to preserve the original order as much as possible, so we treat same tool categories as equal
+/// We still need to pull Node up first, followed by Npm / Yarn, and then Packages last
+fn tool_spec_sort_comparator(left: &ToolSpec, right: &ToolSpec) -> Ordering {
```
chriskrycho: I'd suggest moving this to be an implementation on the `ToolSpec`:

```rust
impl ToolSpec {
    // ... all the other impls
    fn sort_comparator(left: &ToolSpec, right: &ToolSpec) -> Ordering {
        // ... the same body you have below
    }
}
```

 The invocation would then just be `tools.sort_by(Self::sort_comparator)`.

crates/volta-core/src/tool/mod.rs
```diff
@@ -193,6 +194,24 @@ impl ToolSpec {
     }
 }
 
+/// Function for comparing toolspecs for use in sorting.
```
chriskrycho: Formatting nits:

-   We can make the first line imperative voice:

    ```rust
    /// Compare `ToolSpec`s for sorting.
    ```
  
    This drops the repetitive "Function for…" at the beginning


-   There should be a blank newline (starting a new Markdown paragraph) between the first line (the summary) and the following lines.

-   My own personal preference, which helps a lot with readability in apps which don't have some kind of autowrapping (or have it disabled, as I do), is to wrap paragraphs at 80 characters, so:

    ```rust
    /// We want to preserve the original order as much as possible, so we treat same
    /// tool categories as equal. We still need to pull Node up first, followed by
    /// Npm / Yarn, and then Packages last.
    ```

    (That one is *really* just a nit, though.)

crates/volta-core/src/tool/mod.rs
charlespierce: Good thinking, that's a lot cleaner and better encapsulated!

crates/volta-core/src/tool/mod.rs
charlespierce: Done!

PR #486, opened 2019-06-24
Separate bin manifest parse from regular manifest

### LINKED ISSUE #476: `volta install` should not warn when package has a `toolchain` property
Packages that are published with a `toolchain` property in their `package.json` are currently causing a warning (to change from `toolchain` to `volta`) even though the property is not used to actually impact what will be used.

For example:

```
❯ volta install neovim
  Fetching neovim@4.5.0  [========================================] 100%
warning: this project
         (`/Users/rjackson/.volta/tools/image/packages/neovim/4.5.0/package.json
         `) is configured with the `toolchain` key, which is deprecated and will
         be removed in a future version. Please switch to `volta` instead.
warning: this project
         (`/Users/rjackson/.volta/tools/image/packages/neovim/4.5.0/package.json
         `) is configured with the `toolchain` key, which is deprecated and will
         be removed in a future version. Please switch to `volta` instead.
```

charlespierce:
Closes #476 

Info
-----
* Currently the package install logic uses the full project Manifest parser to read the `package.json` and get the `bin` and `engines` information.
* This causes the package install to show a warning when the package being installed has a `toolchain` property.
* Additionally, we are parsing extra fields (`dependencies`, `volta`, etc.) that aren't needed in the package install context.

Changes
-----
* Split `Manifest` into two structs (`Manifest` for the project information and `BinManifest` for the specific information needed when doing a package install).
* Update the deserialization logic to reflect the split in behavior.
* Switch package install process to use the `BinManifest` instead of the `Manifest` object, which removes the code path that shows a warning about `toolchain` existing

### DIFF
--- a/crates/volta-core/src/manifest/mod.rs
+++ b/crates/volta-core/src/manifest/mod.rs
@@ -128,12 +128,10 @@ impl BinManifest {
             file: package_file.to_path_buf(),
         })?;
 
-        let serial: serial::RawBinManifest =
-            serde_json::de::from_reader(file).with_context(|_| {
-                ErrorDetails::PackageParseError {
-                    file: package_file.to_path_buf(),
-                }
-            })?;
-        Ok(serial.into_bin_manifest())
+        serde_json::de::from_reader::<File, serial::RawBinManifest>(file)
+            .with_context(|_| ErrorDetails::PackageParseError {
+                file: package_file.to_path_buf(),
+            })
+            .map(BinManifest::from)
     }
 }
--- a/crates/volta-core/src/manifest/mod.rs
+++ b/crates/volta-core/src/manifest/mod.rs
@@ -23,10 +23,6 @@ pub struct Manifest {
     pub dependencies: HashMap<String, String>,
     /// The `devDependencies` section.
     pub dev_dependencies: HashMap<String, String>,
-    /// The `bin` section, containing a map of binary names to locations.
-    pub bin: HashMap<String, String>,
-    /// The `engines` section, containing a spec of the Node versions that the package works on.
-    pub engines: Option<String>,
 }
 
 impl Manifest {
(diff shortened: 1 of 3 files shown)


crates/volta-core/src/manifest/mod.rs
```diff
@@ -123,7 +114,26 @@ impl Manifest {
...
 
-#[cfg(test)]
-pub mod tests;
+        let serial: serial::RawBinManifest =
+            serde_json::de::from_reader(file).with_context(|_| {
+                ErrorDetails::PackageParseError {
+                    file: package_file.to_path_buf(),
+                }
+            })?;
+        Ok(serial.into_bin_manifest())
```
chriskrycho: FWIW, you can make this a little shorter and IMO clearer:

```rust
serde_json::de::from_reader(file)
    .with_context(|_| ErrorDetails::PackageParseError {
        file: package_file.to_path_buf(),
    })
    .map(serial::RawBinManifest::into_bin_manifest)
```

crates/volta-core/src/manifest/serial.rs
```diff
@@ -177,6 +168,28 @@ impl Manifest {
     }
 }
 
+impl RawBinManifest {
+    pub fn into_bin_manifest(self) -> super::BinManifest {
```
chriskrycho: We could also do this with the `From` trait (which also gets us `.into()` for free), though presumably in the module with `BinManifest` rather than here:

```rust
impl From<RawBinManifest> for BinManifest {
    fn from(raw: RawBinManifest) -> Self {
        // exactly the same implementation
    }
}
```

Implementing `From` will give us `into` for free, and is also the [idiomatic] way to do this:

> Library authors should not directly implement this trait, but should prefer implementing the `From` trait, which offers greater flexibility and provides an equivalent Into implementation for free, thanks to a blanket implementation in the standard library.

[idiomatic]: https://doc.rust-lang.org/1.34.1/std/convert/trait.Into.html

The net is we'll be able to write this:

```rust
fn takes_bin_manifest(bin_manifest: BinManifest) {
    // ...
}

takes_bin_manifest(a_raw_bin_manifest.into());
```

chriskrycho: As a related note: 1.34 stabilized the `TryFrom` and `TryInto` traits, so a fair number of places where we have custom handling for `into_...` which is `Fallible`, we can convert to implementations of `TryFrom` instead. Definitely not for *this* PR; I'll open an issue tracking it, though.

charlespierce: Good call, that is simpler! I opted to leave the `From` implementation in the `serial.rs` module, however, since it's related to the serialization, so it seems like it would fit better there.

crates/volta-core/src/manifest/mod.rs
charlespierce: Done!

Issue #497, opened 2019-07-12
Fetch Node & Yarn archive Shasums and compare when downloading


charlespierce:
Currently we don't check the shasum for downloads of Node and Yarn archives, we just assume that the download was successful if the library reports success. We should fetch the hashes from their canonical sources and then verify our downloads against those values.

We should also persist the shasum in the same way that we do for packages, so that we can use it to verify the integrity of the cache in the future.

Note: We also need to be sure to consider hooks for fetching the Shasums, as when we are downloading from non-public sources we may need to support different hashes.

charlespierce:
Closing in favor of #134

PR #498, opened 2019-07-12
Download archives to temp file to prevent corrupted state

### LINKED ISSUE #200: Interrupting download of a tool leaves the cache in a bad state
I installed notion on a new computer yesterday and set it up to download Node v10, and it fell asleep midway through the install. The install failed, and when I went to try to rerun the command I kept getting errors.

I was able to reproduce this by sending a kill signal mid install (`CTRL + C`), and cleaning the cached binaries at `~/.notion` fixed the issue.

charlespierce:
Closes #200 
Closes #275 

Info
-----
When downloading Node or Yarn, if the download is interrupted (by the user pressing Ctrl+C, for example) or if the download is corrupted in some way, the Volta internal state gets corrupted. We currently write the archive directly to the cache, so if it is invalid in some way, on subsequent attempts, we continue to try to use that invalid cache.

Changes
-----
* Updated `distro::node` and `distro::yarn` to download the archives into a temp file, and only move that temp file over to the cache location once the fetching and unpacking has completed successfully.

Tested
-----
* Manual testing to verify that interrupting the download of Node or Yarn no longer leaves an incomplete cache file in the inventory directory.
* Manually confirmed that after interrupting the download, a subsequent `volta install node` will succeed, and that `node` is correctly available.
* Added acceptance tests to ensure that if the archive is corrupted, it won't be written to the inventory directory.

Note: This PR doesn't attempt to tackle matching the downloads against the shasum, as that will require additional design to make sure we cover all cases. Opened #497 to cover that functionality.

### DIFF
--- a/crates/volta-core/src/distro/node.rs
+++ b/crates/volta-core/src/distro/node.rs
@@ -4,10 +4,6 @@ use std::fs::{read_to_string, rename, write, File};
 use std::path::{Path, PathBuf};
 use std::string::ToString;
 
-use archive::{self, Archive};
-use serde::Deserialize;
-use tempfile::tempdir_in;
-
 use super::{download_tool_error, Distro, Fetched};
 use crate::error::ErrorDetails;
 use crate::fs::ensure_containing_dir_exists;
@@ -18,8 +14,11 @@ use crate::style::{progress_bar, tool_version};
 use crate::tool::ToolSpec;
 use crate::version::VersionSpec;
 
+use archive::{self, Archive};
 use log::debug;
 use semver::Version;
+use serde::Deserialize;
+use tempfile::{tempdir_in, NamedTempFile};
 use volta_fail::{Fallible, ResultExt};
 
 #[cfg(feature = "mock-network")]
@@ -41,6 +40,7 @@ cfg_if::cfg_if! {
 /// A provisioned Node distribution.
 pub struct NodeDistro {
     archive: Box<dyn Archive>,
+    staging_file: Option<NamedTempFile>,
     version: Version,
 }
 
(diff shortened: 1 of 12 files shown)


crates/volta-core/src/distro/node.rs
```diff
@@ -41,6 +40,7 @@ cfg_if::cfg_if! {
 /// A provisioned Node distribution.
 pub struct NodeDistro {
     archive: Box<dyn Archive>,
+    staging_file: Option<NamedTempFile>,
```
chriskrycho: A thought on seeing this and how it's used—I'd be inclined to have two separate types here: one for the staging version and one for the version where it has successfully downloaded. That way, it's always transparently obvious from the type signatures alone whether you're dealing with a version that is in-progress or one that is in fact ready to be used. Then we could explicitly transform from the one to the other.

I tend to lean toward that pattern over using an `Option` many times, as it makes it possible to use the types to express the legal flows through the program. However, it's possible the refactor that would do that is larger than we want to take on, or that the flow here with `Option` is clear *enough*, so this is by no means a blocker—your call!

crates/volta-core/src/distro/node.rs
```diff
@@ -239,6 +250,17 @@ impl Distro for NodeDistro {
             dir: dest.clone(),
         })?;
 
+        if let Some(staging) = self.staging_file {
```
chriskrycho: In the refactor I suggested above, the invariant we maintain simply by way of checking it ourselves in the code flow here would be much more clearly required by the types themselves: if we had a `StagingDistro` and needed to transform it to a `NodeDistro` (or some such), this implementation would presumably live there, and the types themselves *wouldn't* let us treat a `StagingDistro` as a valid `NodeDistro` elsewhere.

crates/volta-core/src/distro/node.rs
charlespierce: I agree, and I initially considered using an Enum as the different types, but I like the idea of having distinct types that we can transform (we do this already with `PlatformSpec` and `Image`). I think this fits into the larger refactor I'm doing around the entire fetching process, which will likely get rid of the `Distro` trait entirely, because its abstraction breaks down horribly for packages.

In the implementation for that, however, I'll be sure to use distinct types to represent the different cases, instead of an `Option` :+1:

chriskrycho: Sweet; I look forward to seeing it!

Issue #223, opened 2018-12-18
Naming convention: don't pun serial type names


dherman:
As @mikrostew pointed out in #221, it's confusing to have `foo::Bar` and `foo::serial::Bar` types have the same name, especially with `foo::serial::Bar::into_bar()`.

Example: `notion_core::inventory::NodeIndex` and `notion_core::inventory::serial::NodeIndex`.

Some possible naming conventions:
- `notion_core::inventory::serial::NodeIndexAst` (i.e., [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree))
- `notion_core::inventory::serial::NodeIndexAst` (i.e., [IR](https://en.wikipedia.org/wiki/Intermediate_representation))
- `notion_core::inventory::serial::NodeIndexSerial`
- `notion_core::inventory::serial::NodeIndexSerde`
- `notion_core::inventory::serial::NodeIndexSD`
- `notion_core::inventory::serial::SerialNodeIndex`
- `notion_core::inventory::serial::SerdeNodeIndex`
- `notion_core::inventory::serial::SDNodeIndex`

I don't love any of these. Other ideas?

charlespierce:
Looking at some (de)serialization stuff for the package configs, I wonder if the solution would actually be to not have "serial types" at all. From what I can tell, we should be able to serialize / deserialize directly to / from the actual data models used in Volta.

That way we avoid needing to parse into an intermediate type, then convert that type to our data model (and the reverse when writing).

PR #505, opened 2019-07-25
Refactor Tool Resolve / Fetch / Install Process

### LINKED ISSUE #87: Abstract Catalog's activate, install and uninstall methods generic to any Tool (node, yarn)
It's a nice improvement to abstract some of the methods (activate, install and uninstall) in Catalog. Instead of implementing these methods for node and yarn, we can have a generic method for Tool.

### LINKED ISSUE #140: Abstract "tool" (node | npm | yarn | package binary)
We have a lot of code duplication for analogous functions that deal with node vs yarn. We should be able to abstract a lot of this with an abstraction -- maybe an enum, maybe traits -- that lets us define just one function for each operation.

There's already a trait called `Tool` that this will need to be reconciled with. Either `Tool` is the wrong name, or somehow it should be worked into the protocol.

### LINKED ISSUE #240: Make `DistroVersion` more static
One issue that came out of https://github.com/notion-cli/notion/pull/234 is that it _should_ be possible to make `DistroVersion` more static (instead of a dynamic Enum), and verify the code correctness using the type system. This may be possible using some combination of Traits and associated types, but it was non-obvious how exactly to do that without blocking the PR for some time.

### LINKED ISSUE #244: Rename or refactor FetchResolve?
The `FetchResolve` trait isn't the most explanatory name. It's basically the interface for the `Collection` type. It seems like the name should be more evocative of "here's the stuff you can do with a collection" maybe?

charlespierce:
Closes #244
Closes #240 
Closes #140 
Closes #87 

Info
-----
* The process of resolving a requested `VersionSpec` into a `Version` and then fetching that Version of a given tool was abstracted in a way that made it difficult to follow.
* The `inventory::Collection` objects were responsible for resolving a Version, through the horribly named `FetchResolve` Trait, while the actual fetching was delegated to `Distro`.
* Additionally, the `Distro` trait was a poor abstraction that didn't fit well with installing 3rd party packages, as they require additional information which doesn't align with installing Node or Yarn.
* Finally, due to how the control flow was organized, we didn't have any way to access the `Session` when resolving or fetching a tool, which is useful for determining the user's default Node in some spots.

Changes
-----
* The core change is moving the source of the control flow into the `tool` module.
    * The `Spec` (formerly `ToolSpec`) type now has two methods:
        * `resolve` - Consumes the `Spec` and returns a `Resolved` enum type
        * `uninstall` - Consumes the `Spec` and handles uninstalling (only implemented for packages at the moment)
    * The `tool` module has a submodule for each tool type (`node`, `npm`, `package`, and `yarn`)
    * Resolution for each tool has been moved from `inventory` to the `resolve` submodule of each tool, and the `Spec::resolve` method calls into that module based on which tool is being resolved.
    * The `Resolved` type has methods for the 3 main actions that can be taken on a tool: `fetch`, `install`, and `pin` (Not all apply to all tools, however we provide clear error messages for the cases that don't apply).
    * The `Resolved` type contains an implementation of the `Tool` trait for the given tool (`Node`, `Npm`, `Yarn`, and `Package`), each of which has its own logic for the 3 actions.
    * Fetching each tool has been moved from `distro` to the `fetch` submodule for each tool.
    * The implementations of the `Tool` trait perform any additional actions needed (such as updating the inventory, editing the `platform.json`, etc.) based on the action being taken.
    * The various methods on `Session` that were responsible for fetching, pinning, etc, have all been moved into the `Tool` implementations, so they are all located in one place.
* Instead of using the `Distro` implementations as markers for `Collection` and `ToolHooks`, we are now using the `Tool` implementations, which provides a clearer indication of what is happening.
* The code that was in the `tool` module, relating to executing a tool, has been moved into the `run` module.
* Where possible, serialized types have been switched to use `RawXXX` as the name, instead of having the exact same name as the type they are shadowing.
* Where possible, bespoke `into_xxx` methods have been converted to use either `From`/`Into` or `TryFrom`/`TryInto`, to be more idiomatic Rust.

Notes
-----
* We will also need to update the [contributing guide](https://docs.volta.sh/contributing/) to reflect the requirement of Rust 1.34 instead of 1.33, as this refactor uses `TryFrom` in a few places.


crates/volta-core/src/run/binary.rs
```diff
@@ -101,3 +107,55 @@ where
         },
     )
 }
+
+pub struct DefaultBinary {
```
chriskrycho: Can we add a doc-string to explain what this is for? I *think* I follow based on how it's used, but some lightweight docs would be 😍.

crates/volta-core/src/run/binary.rs
```diff
@@ -101,3 +107,55 @@ where
...
+    pub fn from_config(bin_config: BinConfig, session: &mut Session) -> Fallible<Self> {
+        let bin_path = bin_full_path(
+            &bin_config.package,
+            &bin_config.version,
+            &bin_config.name,
+            &bin_config.path,
+        )?;
+
+        // If the user does not have yarn set in the platform for this binary, use the default
+        // This is necessary because some tools (e.g. ember-cli with the --yarn option) invoke `yarn`
```
chriskrycho: Nit: consistency in code formatting—the same way you're wrapping `yarn`, my very detail-obsessed self wants `--yarn` to be wrapped. (This is *not* important, but it distracted me so I figured I'd note it. 😆)

crates/volta-core/src/fs.rs
```diff
@@ -49,8 +51,13 @@ pub fn delete_dir_error<P: AsRef<Path>>(directory: &P) -> impl FnOnce(&io::Error
     |_| ErrorDetails::DeleteDirectoryError { directory }
 }
 
+pub fn delete_file_error<P: AsRef<Path>>(file: &P) -> impl FnOnce(&io::Error) -> ErrorDetails {
```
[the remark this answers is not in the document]
charlespierce: Yeah, `Path` is to `PathBuf` as `str` is to `String`

crates/volta-core/src/fs.rs
```diff
@@ -49,8 +51,13 @@ pub fn delete_dir_error<P: AsRef<Path>>(directory: &P) -> impl FnOnce(&io::Error
...
+    let file = file.as_ref().to_path_buf();
+    |_| ErrorDetails::DeleteFileError { file }
+}
+
 /// Reads a file, if it exists.
-pub fn read_file_opt(path: &PathBuf) -> io::Result<Option<String>> {
+pub fn read_file_opt<P: AsRef<Path>>(path: P) -> io::Result<Option<String>> {
```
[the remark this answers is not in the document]
charlespierce: Good idea, I made a follow-on code quality issue #514

crates/volta-core/src/hook/mod.rs
```diff
@@ -92,6 +88,22 @@ macro_rules! merge_hook_config_field {
 }
 
 impl HookConfig {
+    pub fn node(&self) -> Option<&ToolHooks<Node>> {
```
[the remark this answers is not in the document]
charlespierce: Good call! It's actually useful to call `as_ref` on the Option values, so updated to make the fields non-`pub` 🙂

crates/volta-core/src/hook/serial.rs
```diff
@@ -1,52 +1,50 @@
...
-use crate::distro::Distro;
 use crate::error::ErrorDetails;
+use crate::tool::{Node, Package, Tool, Yarn};
 use serde::{Deserialize, Serialize};
-use volta_fail::Fallible;
+use volta_fail::{Fallible, VoltaError};
 
 #[derive(Serialize, Deserialize)]
-pub struct ResolveHook {
+pub struct RawResolveHook {
```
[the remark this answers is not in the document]
charlespierce: We actually originally had them all named the same as the internal object (e.g. `serial::ResolveHook` vs `hook::ResolveHook`), but we have an issue #223 to avoid having the same name, as it can get confusing to figure out which one is being referred to. I've been using `Raw` as a prefix when I refactor since that most closely reflects what the types are: The raw result of using `serde`.

I could see renaming the module to `raw`, but that would probably be work for a follow-on PR.

crates/volta-core/src/hook/serial.rs
```diff
@@ -99,51 +99,49 @@ impl PublishHook {
...
-    pub latest: Option<ResolveHook>,
-    pub index: Option<ResolveHook>,
+pub struct RawToolHooks<T: Tool> {
+    pub distro: Option<RawResolveHook>,
+    pub latest: Option<RawResolveHook>,
+    pub index: Option<RawResolveHook>,
 
     #[serde(skip)]
-    phantom: PhantomData<I>,
+    pub phantom: PhantomData<T>,
```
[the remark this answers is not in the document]
charlespierce: Nothing ... nothing at all 🙂 Removed.

crates/volta-core/src/hook/serial.rs
```diff
@@ -153,8 +151,8 @@ impl HookConfig {
     }
 }
 
-impl<D: Distro> ToolHooks<D> {
-    pub fn into_tool_hooks(self, base_dir: &Path) -> Fallible<super::ToolHooks<D>> {
+impl<T: Tool> RawToolHooks<T> {
+    pub fn into_tool_hooks(self, base_dir: &Path) -> Fallible<super::ToolHooks<T>> {
```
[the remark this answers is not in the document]
charlespierce: This was actually one of the first ones I wanted to switch over, but this method takes a parameter, so it doesn't quite fit with the contract of `TryFrom`

crates/volta-core/src/run/binary.rs
charlespierce: Sure, easy change!

crates/volta-core/src/run/binary.rs
charlespierce: Definitely, I'll write something up.

crates/volta-core/src/hook/serial.rs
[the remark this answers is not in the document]
chriskrycho: Doh! You’re right; I totally missed that!

crates/volta-core/src/hook/serial.rs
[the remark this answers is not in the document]
chriskrycho: That makes sense, and I certainly see the trade off. I think I actually brought it up myself at one point. I’ll keep thinking on the names here. Agreed that it’s be for a follow up regardless!

crates/volta-core/src/fs.rs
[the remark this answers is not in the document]
chriskrycho: Yep, I knew that hit, but hadn’t thought about it specifically for `as_ref`, which I reach for less often for whatever reason!

crates/volta-core/src/tool/node/serial.rs
```diff
@@ -0,0 +1,52 @@
...
+        NodeIndex { entries }
+    }
+}
+
+fn lts_version_serde<'de, D>(deserializer: D) -> Result<bool, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    match String::deserialize(deserializer) {
+        Ok(_t) => Ok(true),
```
chriskrycho: Nit: it's slightly more idiomatic to just use `_` for these:

```rust
match String::deserialize(deserializer) {
    Ok(_) => Ok(true),
    Err(_) => Err(false),
}
```

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -0,0 +1,143 @@
...
+use semver::Version;
+use sha1::{Digest, Sha1};
+use volta_fail::{Fallible, ResultExt};
+
+pub fn fetch(name: &str, details: &PackageDetails) -> Fallible<()> {
+    let version_string = details.version.to_string();
+    let cache_file = path::package_distro_file(&name, &version_string)?;
+    let shasum_file = path::package_distro_shasum(&name, &version_string)?;
+
+    if let Some(archive) = load_cached_distro(&cache_file, &shasum_file) {
```
chriskrycho: Possible alternative here which I think might be clearer:

```rust
let (archive, cached) = match load_cached_distro(&cache_file, &shasum_file) {
    Some(archive) => (archive, true),
    None => fetch_remote_distro(
        tool::Spec::Package(name.into(), VersionSpec::exact(&details.version)),
        &details.tarball_url,
        &cache_file,
    )
    .map(|archive| (archive, false))?,
};

if cached {
    debug!(
        "Loading {} from cached archive at '{}'",
        tool_version(&name, &version_string),
        cache_file.display(),
    );
}

unpack_archive(archive, name, &details.version)?;

if cached {
    Ok(())
} else {
    write(&shasum_file, details.shasum.as_bytes()).with_context(|_| {
        ErrorDetails::WritePackageShasumError {
            package: name.into(),
            version: version_string,
            file: shasum_file,
        }
    })
}
```

As for *why* I think that's clearer: it seems to me that it more cleanly separates the concerns of the function into their own distinct blobs; it's clearer here how you'd factor it out into its own function if you wanted.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -0,0 +1,143 @@
...
+        // Save the shasum in a file
+        write(&shasum_file, details.shasum.as_bytes()).with_context(|_| {
+            ErrorDetails::WritePackageShasumError {
+                package: name.into(),
+                version: version_string,
+                file: shasum_file,
+            }
+        })
+    }
+    // Check for cache, if available then use that (unpack_archive)
```
chriskrycho: Looks like this comment should probably move to the top of the function and/or be deleted wholesale?

charlespierce: Whoops, that should have been deleted! That was one of my comments I did mapping out the flow of the code before I started writing 😓Good catch!

crates/volta-core/src/tool/package/fetch.rs
charlespierce: I like that approach. I actually implemented it slightly differently, including the `debug!` message as part of the `Some(archive)` match arm, but that cleans it up nicely and makes the flow clearer, thanks!

crates/volta-core/src/tool/node/serial.rs
charlespierce: Updated!

src/command/uninstall.rs
```diff
@@ -1,7 +1,7 @@
 use structopt::StructOpt;
 
 use volta_core::session::{ActivityKind, Session};
-use volta_core::tool::ToolSpec;
+use volta_core::tool::Spec;
```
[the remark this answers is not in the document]
charlespierce: Agreed, I thought I had done that everywhere, I guess I didn't. Updated!

crates/volta-core/src/tool/package/fetch.rs
chriskrycho: That makes sense and is probably cleaner!

crates/volta-core/src/session.rs
```diff
@@ -108,6 +102,13 @@ impl Session {
         self.project.get()
     }
 
+    /// Returns the user's currently active platform, if any
```
chriskrycho: 😍 I *love* this documentation string. 🙇

crates/volta-core/src/session.rs
```diff
@@ -108,6 +102,13 @@ impl Session {
...
+    ///
+    /// Active platform is determined by first looking at the Project Platform
+    /// - If it exists and has a Yarn version, then we use the project platform
```
chriskrycho: Nit: there should be an empty line between the previous line and this one, so that it renders correctly as a Markdown list in editors and IDEs which understand that.

Without, it'll tend to render like this:

> Active platform is determined by first looking at the Project Platform - If it exists and has a Yarn version, then we use the project platform - If it exists but doesn't have a Yarn version, then we merge the two, pulling Yarn from the user default platform, if available - If there is no Project platform, then we use the user Default Platform

With, it'll render as expected:

> Active platform is determined by first looking at the Project Platform
> 
> - If it exists and has a Yarn version, then we use the project platform
> - If it exists but doesn't have a Yarn version, then we merge the two, pulling Yarn from the user default platform, if available
> - If there is no Project platform, then we use the user Default Platform

(There's some variation on how various Markdown engines parse it, so you *may* see it work as expected as is in *some* editors as is, but you *will* see it work as expected in *all* editors with the tweak.)

crates/volta-core/src/session.rs
```diff
@@ -177,220 +186,12 @@ impl Session {
...
-            project.pin_node(&node_version)?;
-            info!(
-                "{} pinned {} (with {}) in package.json",
-                success_prefix(),
-                tool_version("node", node_version.runtime),
-                tool_version("npm", node_version.npm),
-            );
-        } else {
-            throw!(ErrorDetails::NotInPackage);
+            Yarn::new(version.clone()).fetch_internal(self)?;
```
chriskrycho: Well, this sure-as-heck simplifies a bunch of this implementation!

crates/volta-core/src/session.rs
```diff
@@ -428,10 +229,7 @@ impl Session {
 
 fn publish_plugin(hooks: &LazyHookConfig) -> Fallible<Option<&Publish>> {
     let hooks = hooks.get()?;
-    Ok(hooks
-        .events
-        .as_ref()
-        .and_then(|events| events.publish.as_ref()))
+    Ok(hooks.events().and_then(|events| events.publish.as_ref()))
```
chriskrycho: One note: this whole body can become:

```rust
hooks.get().map(|hooks| hooks.events().and_then(|events| events.publish.as_ref()))
```

This isn't a necessary refactor; I just tend to lean that direction when I see a `?` immediately followed by an `Ok` wrapping something.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -0,0 +1,108 @@
...
+            index: Some(ref hook),
+            ..
+        }) => {
+            debug!("Using yarn.index hook to determine yarn index URL");
+            hook.resolve("releases")?
+        }
+        _ => public_yarn_version_index(),
+    };
+
+    let spinner = progress_spinner(&format!("Fetching public registry: {}", url));
```
chriskrycho: One thought for a code cleanliness thing: I think we might want to extract our spinner invocations to something like this:

```rust
fn task_with_spinner<T>(description: &str, task: impl Fn() -> T) {
    let spinner = progress_spinner(description);
    let result = task();
    spinner.finish_and_clear();
    result
}
```

Then we'd write this whole sequence (and the many others like it) like this (possibly with `move ||` in some cases, I haven't pushed too far down this rabbit hole yet):

```rust
let releases = task_with_spinner(&format!("Fetching public registry: {}", url), || {
    let releases: serial::RawYarnIndex = reqwest::get(&url)
         .and_then(|mut resp| resp.json())
         .with_context(registry_fetch_error("Yarn", &url))?;
     let index: YarnIndex = releases.into();
     index.entries
});
```

Very much a task for *later*, but I think it'd eliminate a lot of boilerplate quite nicely.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -0,0 +1,108 @@
...
+    let releases: serial::RawYarnIndex = reqwest::get(&url)
+        .and_then(|mut resp| resp.json())
+        .with_context(registry_fetch_error("Yarn", &url))?;
+    let index: YarnIndex = releases.into();
```
chriskrycho: Slightly more idiomatic:

```suggestion
    let index = YarnIndex::from(releases);
```

crates/volta-core/src/run/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        _ => binary::command(exe, args, session)?,
+    };
+
+    command.exec()
+}
+
+/// Represents the command to execute a tool
+struct ToolCommand {
+    command: Command,
+    error: ErrorDetails,
```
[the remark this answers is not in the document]
charlespierce: Yeah, that's essentially what it does. It's mostly for the pass-through behavior, where we want to let the system resolve e.g. `node`, but also want to have a nice Volta error to show if that fails (instead of the generic "command not found" error that the OS would otherwise provide. I'll update the name, I like `on_failure` as it's more descriptive of what the field is for.

I like the idea of adding a doc comment in spots where it isn't trivially obvious what something is for, so I'll add one here.

crates/volta-core/src/tool/yarn/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+pub struct Yarn {
+    pub(super) version: Version,
+}
+
+impl Yarn {
+    pub fn new(version: Version) -> Self {
+        Yarn { version }
+    }
+
+    pub(crate) fn fetch_internal(&self, session: &mut Session) -> Fallible<()> {
```
chriskrycho: I'd be inclined here to pass in the required `&mut Inventory` and `&Hooks` dependencies *instead* of `Session`: it makes the contract here clearer, and it decreases our coupling to the current design of the `Session` as our bag for everything (in case we want to refactor in the future). In general I prefer to pass functions as minimal and exact a set of what they need as possible, for that reason.

crates/volta-core/src/tool/yarn/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+            .yarn
+            .versions
+            .insert(self.version.clone());
+
+        Ok(())
+    }
+}
+
+impl Tool for Yarn {
+    fn fetch(self, session: &mut Session) -> Fallible<()> {
```
chriskrycho: I note that the design of the `Tool` trait is what necessitates passing down `Session` instead, so we'd need to push that refactor back up the chain. However, I think it's likely worth it for all the reasons mentioned above. If the `Tool` trait can become entirely ignorant of `Session`, that'd be a huge win in my book.

crates/volta-core/src/run/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+impl ToolCommand {
+    fn direct<A>(exe: &OsStr, args: A, path_var: &OsStr) -> Self
```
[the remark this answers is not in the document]
charlespierce: That's true, this code didn't change, it was just copied from a different file. But a refactor like this is the perfect time to add some doc comments and make things more clear :+1:

crates/volta-core/src/tool/yarn/fetch.rs
```diff
@@ -0,0 +1,150 @@
...
+            "https://github.com/yarnpkg/yarn/releases/download".to_string()
+        }
+    }
+}
+
+pub fn fetch(version: &Version, hooks: Option<&ToolHooks<Yarn>>) -> Fallible<()> {
+    let yarn_dir = path::yarn_inventory_dir()?;
+    let cache_file = yarn_dir.join(path::yarn_distro_file_name(&version.to_string()));
+
+    if let Some(archive) = load_cached_distro(&cache_file) {
```
chriskrycho: This could probably be refactored/broken apart along similar lines to the similar functionality I identified in one of the other packages (the Node one IIRC).

crates/volta-core/src/session.rs
charlespierce: Done!

crates/volta-core/src/session.rs
charlespierce: While I like the functional syntax and the `map` / `and_then` methods, I tend to avoid nesting them, as it starts to feel like "callback hell" and I find it a little difficult to parse. Using the `?` here lets us be nearly as brief, while only having one level of lambda.

crates/volta-core/src/tool/yarn/resolve.rs
chriskrycho: Aside: I whipped up a tiny demo prototype of that [in this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=396d5c43f13ab1d0977ccd16f55bd1d9) when I wrote it up to make sure I wasn't missing something important. It works! 😅

crates/volta-core/src/tool/mod.rs
```diff
@@ -1,660 +1,188 @@
...
+    fn fetch(self, session: &mut Session) -> Fallible<()>;
+    /// Install a tool, making it the default so it is available everywhere on the user's machine
+    fn install(self, session: &mut Session) -> Fallible<()>;
+    /// Pin a tool in the local project so that it is usable within the project
+    fn pin(self, session: &mut Session) -> Fallible<()>;
 }
 
 /// Specification for a tool and its associated version.
-///
-/// Since [`Ord`] is implemented for `ToolSpec`, we can use `.sort` on any
```
chriskrycho: Oh hello, a prime example of comments getting out of date with code. 😬 Glad this is gone now.

crates/volta-core/src/tool/yarn/resolve.rs
charlespierce: I like that, and we can likely do something similar with `progress_bar` and cut out a lot of the duplicated boilerplate in the `fetch` modules. Created #518 to cover that case, linking back to this discussion.

crates/volta-core/src/tool/mod.rs
```diff
@@ -1,660 +1,188 @@
...
-            ToolSpec::Node(version) => session.install_node(&version)?,
-            // ISSUE(#292): Implement install for npm
-            ToolSpec::Npm(_version) => throw!(ErrorDetails::Unimplemented {
-                feature: "Installing npm".into()
-            }),
-            ToolSpec::Yarn(version) => session.install_yarn(&version)?,
-            ToolSpec::Package(name, version) => {
-                session.install_package(name.to_string(), &version)?;
+            Spec::Node(version) => {
+                node::resolve(version, session).map(|version| Resolved::Node(Node::new(version)))
```
chriskrycho: This is totally a take-it-or-leave-it suggestion; I like point-free style but a lot of people don't.

```rust
node::resolve(version, session)
    .map(Node::new)
    .map(Resolved::Node)
```

This works because enum variant constructors are in fact functions!

It ends up reading very nicely and cleanly if you apply this to the following functions as well:

Screen Shot 2019-08-06 at 12 52 37


As ever with taste ones, no worries if you prefer it as is, just figured I'd note it as an option.

crates/volta-core/src/tool/yarn/fetch.rs
charlespierce: Good call, I'll do that here and in `package/fetch.rs` as they all follow a roughly similar pattern.

chriskrycho:
Yeah, that’s totally fair. One thought further, then—what are your thoughts on this?

```rust
let  = hooks.get()?;
let events = hooks.events().and_then(|events| events.publish.as_ref());
Ok(events)
```

(Again, no particular pressure to change it; that’s just noting what reads more clearly to me.)

crates/volta-core/src/tool/yarn/mod.rs
charlespierce: I think it's a bit of a balance, given that the different implementations use different subsets of the available data. If we restructure it to pass in exactly what is needed, then the parameters on the Trait method need to be the superset that contains the necessary data for _any_ of the implementations. It also means if we, in the future, want to add some additional functionality that requires other data off of the `Session`, then we have to plumb that through the trait and each of the implementations.

For example, one follow-on refactor that we should be able to do is to generate the `PlatformSpec` and call `npm view` directly in the `package::resolve` module, instead of the current design which requires calling out to the `npm` shim. That will require us to pass along the result of `current_platform` to the `resolve` method, even though most of the resolvers don't actually need to know about that.

While I agree the `Session` is a "bag of everything", I think this refactor has removed a lot of the behavior, which lets it be more of a "state of the world" object, which I think is a bit more reasonable to pass around.

crates/volta-core/src/session.rs
chriskrycho: Yeah, that’s totally fair. One thought further, then—what are your thoughts on this?

```rust
let  = hooks.get()?;
let events = hooks.events().and_then(|events| events.publish.as_ref());
Ok(events)
```

(Again, no particular pressure to change it; that’s just noting what reads more clearly to me.)

crates/volta-core/src/tool/yarn/mod.rs
charlespierce: See below for some of the reasoning behind passing the `Session` object. In this case, I think there's less of an issue if we wanted to pass the needed objects directly, the only concern I have is then each of the Trait methods that call `fetch_internal` will have to independently resolve the `&mut Inventory` and `&Hooks` objects, which adds some boilerplate to the overall flow, without a significant benefit.

crates/volta-core/src/tool/yarn/mod.rs
chriskrycho: Yeah, that's also a fair take. In situations where I find that to be the case, it makes me wonder if I've actually over-abstracted. (In general, I'm starting to lean toward the conclusion that it's *very* difficult to properly abstract over *all* of runtimes, package managers, and packages.)

I'm happy to defer to you on this, and if/when we have pain in the future we can make a refactor-to-make-the-hard-thing-easy at that point.

crates/volta-core/src/session.rs
charlespierce: That's reasonable, I'll make that change :+1:

crates/volta-core/src/tool/yarn/fetch.rs
charlespierce: Hmm, actually looking at this it's slightly different behavior, because the after-unpack operation in the non-cached case refers to a variable that was only created in the non-cached branch: we only create a `NamedTempFile` to download into if we're actually downloading, if we're unpacking from the cache then it's already there and we don't have to worry about it.

We could return an `Option<NamedTempFile>` from the match, but then it starts to be a question of whether that is cleaner to read or not. Thoughts?

crates/volta-core/src/tool/mod.rs
charlespierce: Oh, I like that a lot :+1:

crates/volta-core/src/tool/yarn/mod.rs
charlespierce: Yeah, it's definitely possible that we're still over-abstracted even given how much has been pulled out, but I think waiting to see if extra pain points come up is a good approach. We definitely need to make sure we call out as soon as something doesn't feel right, and not just shoe-horn it into working, so that we can refactor quicker and smaller, and not let it grow until it's a tangled mess.

crates/volta-core/src/tool/yarn/fetch.rs
chriskrycho: Ah, so you end up with something like this, then:

```rust
let (archive, staging) = match load_cached_distro(&cache_file) {
    Some(archive) => {
        debug!(
            "Loading {} from cached archive at '{}'",
            tool_version("yarn", &version),
            cache_file.display(),
        );
        (archive, None)
    }
    None => {
        let staging = create_staging_file()?;
        let remote_url = determine_remote_url(&version, hooks)?;
        let archive = fetch_remote_distro(&version, &remote_url, staging.path())?;

        (archive, Some(staging))
    }
};

unpack_archive(archive, version)?;

if let Some(staging_file) = staging {
    ensure_containing_dir_exists(&cache_file)?;
    staging_file
        .persist(cache_file)
        .with_context(|_| ErrorDetails::PersistInventoryError { tool: "Yarn".into() })?;
}

Ok(())
```

🤔 I could go either way on whether that's clearer. It still has the benefit of separating out the disjoint operations, but at the cost of adding a few lines and possibly some lack of clarity. Totally your call.

crates/volta-core/src/tool/yarn/mod.rs
chriskrycho: Yep, I think we're agreed on approach here! 👍

crates/volta-core/src/tool/yarn/fetch.rs
charlespierce: That's actually not as unclear as I was picturing, so I'll do that. It also has the benefit of being consistent with the `package::fetch` implementation.

crates/volta-core/src/tool/package/resolve.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/// Index of versions of a specific package.
+pub struct PackageIndex {
+    pub latest: Version,
+    pub entries: Vec<PackageDetails>,
+}
+
+// use the input predicate to match a package in the index
+fn match_package_details(
```
chriskrycho: The `match_` bit here naming-wise kind of threw me in the invocations above because I wanted to parse it as the start of a `match` block. My two thoughts are: `matching_package_details` and something like `details_for_package_matching`. (Not a deal-breaker *not* to do it, either.)

crates/volta-core/src/tool/package/resolve.rs
```diff
@@ -0,0 +1,211 @@
...
+    index: PackageIndex,
+    predicate: impl Fn(&PackageDetails) -> bool,
+) -> Option<PackageDetails> {
+    let mut entries = index.entries.into_iter();
+    entries.find(predicate)
```
chriskrycho: How valuable do you find this helper? I'm kind of wondering if it isn't better simply to inline these invocations. 🤔 

```rust
let details = package_index
    .entries
    .into_iter()
    .find(|PackageDetails { version, .. }| {
         matching.matches(&version)
    });

File: crates/volta-core/src/tool/node/resolve.rs [hunk 28]
```diff
@@ -0,0 +1,272 @@
...
+}
+
+fn match_node_version(
+    url: &str,
+    predicate: impl Fn(&NodeEntry) -> bool,
+) -> Fallible<Option<Version>> {
+    let index: NodeIndex = resolve_node_versions(url)?.into();
+    let mut entries = index.entries.into_iter();
+    Ok(entries
+        .find(predicate)
```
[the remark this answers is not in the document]
charlespierce: Interesting, TIL!

crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -0,0 +1,272 @@
...
+            ensure_containing_dir_exists(&index_cache_file)?;
+            cached.persist(&index_cache_file).with_context(|_| {
+                ErrorDetails::WriteNodeIndexCacheError {
+                    file: index_cache_file,
+                }
+            })?;
+
+            let expiry = create_staging_file()?;
+
+            // Block to borrow expiry for expiry_file.
```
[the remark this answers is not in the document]
charlespierce: I hadn't checked before, but it looks like the NLL does solve this problem and everything compiles fine without the extra blocks :+1:

crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -0,0 +1,272 @@
...
+                .with_context(registry_fetch_error("Node", url))?;
+            let index: serial::RawNodeIndex = serde_json::de::from_str(&response_text)
+                .with_context(|_| ErrorDetails::ParseNodeIndexError {
+                    from_url: url.to_string(),
+                })?;
+
+            let cached = create_staging_file()?;
+
+            // Block to borrow cached for cached_file.
+            {
```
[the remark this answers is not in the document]
charlespierce: As above, removed the now-unnecessary blocks :+1:

crates/volta-core/src/tool/package/resolve.rs
charlespierce: I think that's actually the cleanest solution, especially given that it's only ever used once in any given flow, so it's not a case where the helper is actually providing all that much. That also solves the naming issue above :+1:

crates/volta-core/src/tool/package/resolve.rs
charlespierce: As below, inlining the helper removes the naming issue.

PR #512, opened 2019-07-31
Tooling: update IDEA config to reference Volta

chriskrycho:


chriskrycho:
Ah, good call. Let me update with an updated `.gitignore`!

PR #529, opened 2019-08-16
Update how RPM install works, and auto-create shims along with the home dir layout

mikrostew:
## RPM install changes

The `volta` binary is now installed to `/usr/bin`, so it will be available on the PATH after install. For the user who ran the install, everything is still setup in their home dir.

## Shim auto-creation changes

If this was installed by `root` (which seems to be typical in managed environments), then other users will not have `~/.volta/` setup for them. We already had code to create the directory layout if it does not exist, and this adds one more step to also create the base shims (`node`, `npm`, `yarn`). This allows a non-root user to bootstrap `~/.volta` after this has been installed by root.

There are also some changes to where Volta looks for the `shim` executable. I added an option to look in the install location `/usr/bin/volta-lib/shim` for the RPM special case, so non-root users can still create shims that will symlink correctly.

Overall, this is an incremental change that will help us get this in CI, as well as move closer to working with the upcoming upgrade changes.

### DIFF
--- a/crates/volta-core/src/path/mod.rs
+++ b/crates/volta-core/src/path/mod.rs
@@ -46,10 +46,15 @@ pub fn ensure_volta_dirs_exist() -> Fallible<()> {
         ensure_dir_exists(tmp_dir()?)?;
         ensure_dir_exists(log_dir()?)?;
         // also ensure the basic shims exist
-        ensure_shim_exists("node")?;
-        ensure_shim_exists("yarn")?;
-        ensure_shim_exists("npm")?;
-        ensure_shim_exists("npx")?;
+        // this is only for unix until the update process is refactored
+        // (windows stores the location in the Registry, which is not available for the tests)
+        #[cfg(unix)]
+        {
+            ensure_shim_exists("node")?;
+            ensure_shim_exists("yarn")?;
+            ensure_shim_exists("npm")?;
+            ensure_shim_exists("npx")?;
+        }
     }
 
     Ok(())
--- a/crates/volta-core/src/path/mod.rs
+++ b/crates/volta-core/src/path/mod.rs
@@ -7,6 +7,7 @@ use std::fs;
 use std::path::{Path, PathBuf};
 
 use crate::error::ErrorDetails;
+use crate::shim;
 use volta_fail::{Fallible, ResultExt};
 
 cfg_if::cfg_if! {
@@ -44,6 +45,10 @@ pub fn ensure_volta_dirs_exist() -> Fallible<()> {
         ensure_dir_exists(user_toolchain_dir()?)?;
         ensure_dir_exists(tmp_dir()?)?;
         ensure_dir_exists(log_dir()?)?;
+        // also ensure the basic shims exist
+        ensure_shim_exists("node")?;
+        ensure_shim_exists("yarn")?;
+        ensure_shim_exists("npm")?;
     }
 
     Ok(())
@@ -51,6 +56,10 @@ pub fn ensure_volta_dirs_exist() -> Fallible<()> {
 
 fn ensure_dir_exists(path: PathBuf) -> Fallible<()> {
     fs::create_dir_all(&path).with_context(|_| ErrorDetails::CreateDirError { dir: path })
+}
+
+fn ensure_shim_exists(shim_name: &str) -> Fallible<shim::ShimResult> {
+    shim::create(shim_name)
 }
 
 pub fn volta_home() -> Fallible<PathBuf> {
--- a/crates/volta-core/src/path/mod.rs
+++ b/crates/volta-core/src/path/mod.rs
@@ -49,6 +49,7 @@ pub fn ensure_volta_dirs_exist() -> Fallible<()> {
         ensure_shim_exists("node")?;
         ensure_shim_exists("yarn")?;
         ensure_shim_exists("npm")?;
+        ensure_shim_exists("npx")?;
     }
 
     Ok(())
--- a/crates/volta-core/src/path/unix.rs
+++ b/crates/volta-core/src/path/unix.rs
@@ -120,7 +120,7 @@ pub fn shim_executable() -> Fallible<PathBuf> {
     // if VOLTA_SHIM is set, try that first
     // (not documented yet, as it's currently only used for testing)
     if let Some(shim_location) = env::var_os("VOLTA_SHIM") {
-        return Ok(Path::new(&shim_location).to_path_buf());
+        return Ok(shim_location.into());
     }
 
     // default location for the shim executable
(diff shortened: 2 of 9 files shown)


crates/volta-core/src/path/mod.rs
```diff
@@ -44,6 +45,10 @@ pub fn ensure_volta_dirs_exist() -> Fallible<()> {
         ensure_dir_exists(user_toolchain_dir()?)?;
         ensure_dir_exists(tmp_dir()?)?;
         ensure_dir_exists(log_dir()?)?;
+        // also ensure the basic shims exist
+        ensure_shim_exists("node")?;
+        ensure_shim_exists("yarn")?;
+        ensure_shim_exists("npm")?;
```
charlespierce: For consistency with the installer, we should ensure the `npx` shim exists as well.

crates/volta-core/src/path/unix.rs
```diff
@@ -109,12 +110,34 @@ pub fn shim_file(toolname: &str) -> Fallible<PathBuf> {
...
     Ok(volta_home()?.join("volta"))
 }
 
+// check that it exists - if not, check some other locations
 pub fn shim_executable() -> Fallible<PathBuf> {
-    Ok(volta_home()?.join("shim"))
+    // if VOLTA_SHIM is set, try that first
+    // (not documented yet, as it's currently only used for testing)
+    if let Some(shim_location) = env::var_os("VOLTA_SHIM") {
+        return Ok(Path::new(&shim_location).to_path_buf());
```
charlespierce: `PathBuf` implements `From<OsString>`, so this can be simplified with `.into()`:

```suggestion
        return Ok(shim_location.into());
```

crates/volta-core/src/path/unix.rs
```diff
@@ -109,12 +110,34 @@ pub fn shim_file(toolname: &str) -> Fallible<PathBuf> {
...
+    // default location for the shim executable
+    // (this will be the case for the majority of installs)
+    let default_shim_executable = volta_home()?.join("shim");
+    if default_shim_executable.exists() {
+        return Ok(default_shim_executable);
+    }
+
+    // when an RPM is installed as root, the shim will be here for non-root users
+    // (this will be the case for some managed installs)
+    let rpm_shim_executable = Path::new("/usr/bin/volta-lib/shim").to_path_buf();
```
charlespierce: Similarly here, we can take advantage of the existing `From` implementation to simplify slightly.

```suggestion
    let rpm_shim_executable = PathBuf::from("/usr/bin/volta-lib/shim");
```

dev/rpm/volta.spec
```diff
@@ -32,42 +32,59 @@ cargo build --release
...
 # this runs after install, and sets up VOLTA_HOME and the shell integration
 %post
 printf '\033[1;32m%12s\033[0m %s\n' "Running" "Volta post-install setup..." 1>&2
 # run this as the user who invoked sudo (not as root, because we're writing to $HOME)
-/bin/su -c %{_bindir}/%{name}/volta-postinstall.sh - $SUDO_USER
+/bin/su -c %{volta_bin_dir}/volta-postinstall.sh - $SUDO_USER
 
 
 # this runs after uninstall
+# TODO: I don't think this is right - this happens after package upgrade?
```
charlespierce: Is this something we need to look at?

crates/volta-core/src/path/mod.rs
mikrostew: doh! forgot about that one - will add it

crates/volta-core/src/path/unix.rs
mikrostew: ah, awesome

crates/volta-core/src/path/unix.rs
mikrostew: updated

dev/rpm/volta.spec
mikrostew: I'm looking at that now - I was doing an RPM upgrade at some point, and saw it run this after I upgraded the RPM to the version I was testing, so the RPM was upgraded but `~/.volta/` was deleted

mikrostew: After some more testing and research, I found that `%postun` is called after both uninstall _and_ upgrade.

But, there is a way to fix that - when the `%postun` script is called, it is passed "the number of packages that will be left after this step is completed." So the script just needs to check that, to make sure it's an uninstall, not an upgrade.

crates/volta-core/src/path/mod.rs
```diff
@@ -44,6 +45,11 @@ pub fn ensure_volta_dirs_exist() -> Fallible<()> {
         ensure_dir_exists(user_toolchain_dir()?)?;
         ensure_dir_exists(tmp_dir()?)?;
         ensure_dir_exists(log_dir()?)?;
+        // also ensure the basic shims exist
```
charlespierce: The windows installer actually already creates the "shims" (in actuality they are copies of the `shim.exe` binary) for these in the `Program Files` directory. Additionally, the directory itself is stored in the Registry, so it's not easily available using the tests. Can we wrap this in a block that only applies to unix, at least until we refactor the update process? (which I believe will 

```rust
#[cfg(unix)]
{
    ensure_shim_exists("node")?;
    ensure_shim_exists("yarn")?;
    ensure_shim_exists("npm")?;
    ensure_shim_exists("npx")?;
}
```

tests/acceptance/autocreate_home_dir.rs
```diff
@@ -0,0 +1,74 @@
...
+    assert!(Sandbox::dir_exists(".volta/log"));
+    assert!(Sandbox::dir_exists(".volta/tmp"));
+    assert!(Sandbox::dir_exists(".volta/tools/image/node"));
+    assert!(Sandbox::dir_exists(".volta/tools/image/yarn"));
+    assert!(Sandbox::dir_exists(".volta/tools/inventory/node"));
+    assert!(Sandbox::dir_exists(".volta/tools/inventory/packages"));
+    assert!(Sandbox::dir_exists(".volta/tools/inventory/yarn"));
+    assert!(Sandbox::dir_exists(".volta/tools/user"));
+
+    // shims should all be created
```
charlespierce: This test will also need to be updated to reflect the fact that the shims are in a different location and not automatically created in Windows

crates/volta-core/src/path/mod.rs
mikrostew: Yeah, I just pushed some test code that I thought would work, but didn't actually test on windows yet - I'll cfg these out instead

tests/acceptance/autocreate_home_dir.rs
mikrostew: updated with the last commit

PR #533, opened 2019-08-19
Remove unknown() and is_user_friendly() from VoltaError

### LINKED ISSUE #306: Remove `.unknown()` helper method
Once all of the calls to `.unknown()` have been removed in #294, we should remove the helper method completely, to prevent us from accidentally using it in the future. The method is useful for rapid prototyping, but as we are more committed to helpful error messages, we should not be displaying unhelpful `an internal error occurred` messages in the future.

charlespierce:
Closes #306 

Since the error rework, there is no longer a need to support the `VoltaError::unknown()` and `VoltaError::is_user_friendly()` errors. All errors are now user-friendly, and no unknown errors are used anywhere in the code base.

Additionally, we do still want to have separate crates, which will necessitate having separate error types, so we don't want to remove `volta-fail` entirely (as was originally suggested in #334)

### DIFF
--- a/crates/volta-core/src/error/details.rs
+++ b/crates/volta-core/src/error/details.rs
@@ -1382,8 +1382,4 @@ impl VoltaFail for ErrorDetails {
             ErrorDetails::YarnVersionNotFound { .. } => ExitCode::NoVersionMatch,
         }
     }
-
-    fn is_user_friendly(&self) -> bool {
-        true
-    }
 }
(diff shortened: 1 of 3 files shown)


chriskrycho:
> which will necessitate having separate error types

@charlespierce can you elaborate this? If the error type lived in (e.g.) the `volta-core` crate, wouldn't that solve this? Do we have separate error types *today*? 🤔

charlespierce:
@chriskrycho My thought process is that for keeping track of changes to the layout (which will happen separately and much less frequently than changes to Volta as a whole), having the layout code in a separate crate with it's own distinct version will be a benefit. Since `volta-core` will depend on that crate, having the errors be only in `volta-core` would result in a circular dependency between the two crates.

charlespierce:
Oh, and to your last question, we don't have separate types today (except for one that's used in the `test-support` crate, but that can be dropped to a `Fail` instead of `VoltaFail`), but I didn't want to remove it completely, only to have to re-add it shortly as part of the updates.

PR #547, opened 2019-09-13
Refactor Path using Layout Macro

charlespierce:
Builds on and incorporates @dherman's excellent macro from #231 

Info
-----
* To support programmatic updates, we need to have a versioned layout schema.
* Much of this work was already started in #231, there were just a couple of pieces left to complete.
* We need the layout initialization to be fallible, since it relies on OS operations that can fail.
* We also want initialization to be lazy, so we aren't loading it if it isn't needed (especially the install directory lookup on Windows, which requires reading from the registry)

Changes
-----
* Added the `volta_layout_macro` from #231 
* Created a `volta_layout` crate for the actual implementation of layout-specific information.
* Added the existing layout as `v0` in the crate
* Implemented a thread-safe, fallible, lazy initialization of the layout objects, so that we only need to look them up once, instead of every time we access a `path` value.
* Updated everywhere to use the layout objects, instead of using the `path::*` functions.
* Moved some of the helpers that were in `path` to other, more accurate areas (e.g. `fs::symlink_file`, `Node::archive_filename`, etc.)
* Updated some dependencies and resolved some compiler warnings.


crates/volta-core/src/layout/unix.rs
```diff
@@ -0,0 +1,34 @@
...
+    // default location for the install directory
+    // (this will be the case for the majority of installs)
+    let home = volta_home()?.root();
+    if home.join("shim").exists() {
+        return Ok(home.to_owned());
+    }
+
+    // when an RPM is installed as root, the install_dir will be here for non-root users
+    // (this will be the case for some managed installs)
+    let rpm_home = PathBuf::from("/usr/bin/volta-lib");
```
dherman: I don't understand this part... can you explain this a bit more? Is there something about RPMs described in the updates RFC?

It also seems good to have some documentation about managed installs in the web site, and this kind of thing seems like a good candidate for those docs. Not a blocker for this PR but we should make sure to work on it.

crates/volta-core/src/layout/windows.rs
```diff
@@ -0,0 +1,47 @@
...
+pub(super) fn default_home_dir() -> Fallible<PathBuf> {
+    let mut home = dirs::data_local_dir().ok_or(ErrorDetails::NoLocalDataDir)?;
+    home.push("Volta");
+    Ok(home)
+}
+
+cfg_if! {
+    if #[cfg(test)] {
+        pub(super) fn default_install_dir() -> Fallible<PathBuf> {
+            Ok(PathBuf::from(r#"Z:\"#))
```
dherman: Worth a comment explaining how the tests use `Z:\`?

crates/volta-core/src/layout/unix.rs
charlespierce: We currently build an RPM package, using a script from #529 that puts the `volta` binary on `/usr/bin` and the shim binary in `/usr/bin/volta-lib`, so this detects that case. With the updates that will probably go away since we'll just put everything in one place and rename the shim to `volta-shim`.

I'll update the comment to mention that it's the current RPM installer and will likely change when the full managed installs behavior is implemented.

crates/volta-core/src/layout/windows.rs
charlespierce: The tests currently don't actually need a path that exists, but they need the function to not crash because there aren't valid values in the registry. I'll add a comment to that effect :+1:

PR #548, opened 2019-09-13
Improve Error Messages around 'npm view'

### LINKED ISSUE #545: Show package name on install error
I'm hoping I can use `volta` to install `elm` tooling. So I installed it, and ran the following:

```bash
λ volta install elm elm-live elm-test elm-fromat @elm-tooling/elm-language-server
success: installed elm@0.19.0-no-deps with executables: elm
success: installed elm-live@4.0.0 with executables: elm-live
success: installed elm-test@0.19.0-rev6 with executables: elm-test
error: Could not download package metadata.

Please verify that the package name is correct, and you have Node installed (using `volta install node`).
```

Looking at it, I can see that I miss-spelled `elm-format`, but that's not evident in the error message. It would be nice if the line was changed to:

```
error (elm-fromat): Could not download package metadata.
```

to make debugging this easier.

charlespierce:
Closes #545 

Info
-----
* The current implementation of `volta install <package>` call `npm view` and relies on the existing `PATH` / shims to execute `npm`.
* This means that calling the command requires calling the shims, which then have to re-build all of Volta's state, before finally calling `npm` directly.
* With the refactor in #505, it's now possible to access the `Session` object while resolving a package version, so we can create the command and call `npm` directly.
* This cuts out one layer of indirection and allows us to provide nicer error messages about the existence of `node` on the system.
* Additionally, we can now separate being unable to find the package from being unable to find `npm` / `node`, so we can improve the error messaging around the failure states.

Changes
-----
* Made the various tool command constructors in the `run` module available to the crate, so that they can be used to create subcommands where needed.
* Updated the `tool::package::resolve` module to use the above commands instead of manually creating a call to `npm` that relies on the shims.
* Added the requested package name to the `npm view` errors, so that the user is clear which package failed when installing multiple packages at once.

### DIFF
--- a/crates/volta-core/src/tool/package/resolve.rs
+++ b/crates/volta-core/src/tool/package/resolve.rs
@@ -1,13 +1,13 @@
 //! Provides resolution of 3rd-party packages into specific versions, using the npm repository
 
-use std::process::Command;
+use std::ffi::OsString;
 
-use crate::command::create_command;
 use crate::error::ErrorDetails;
 use crate::hook::ToolHooks;
+use crate::run::{self, ToolCommand};
 use crate::session::Session;
 use crate::style::{progress_spinner, tool_version};
-use crate::tool::{Package, PackageDetails};
+use crate::tool::PackageDetails;
 use crate::version::VersionSpec;
 use log::debug;
 use semver::{Version, VersionReq};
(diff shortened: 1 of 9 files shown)


crates/volta-core/src/tool/package/resolve.rs
```diff
@@ -110,17 +109,15 @@ pub struct PackageIndex {
 ///
 /// * normal package installation from the public npm repo
 /// * installing packages from alternate registries, configured via .npmrc
-fn npm_view_query(name: &str, version: &str) -> Fallible<PackageIndex> {
-    let mut command = npm_view_command_for(name, version);
+fn npm_view_query(name: &str, version: &str, session: &mut Session) -> Fallible<PackageIndex> {
```
chriskrycho: What requires `Session` to be `&mut` instead of `&` here?

chriskrycho:
Looks solid! Had one question, but this is a nice win.

crates/volta-core/src/tool/package/resolve.rs
charlespierce: The `PlatformSpec::checkout` operation requires a mutable reference because it can download new versions of Node and Yarn, which requires updating the Collections to include the new versions. I believe everywhere that `Session` is passed around, it’s a mutable reference, because in order to support the mutability anywhere, we need to support it everywhere.

chriskrycho: Ah, that makes sense. Thanks for clarifying!

PR #549, opened 2019-09-15
Add a bash specific load file `shell/unit/load.bash` that takes advan…

mcarifio:
…tage of some bash extensions to define the volta environment.

Specifically:

* Uses the path to load.bash to define VOLTA_HOME.

* Adds ${VOLTA_HOME}/bin to PATH iff it's not already there.

* Provides a hook for local additions if the script is present.

### DIFF
--- a/shell/unix/load.bash
+++ b/shell/unix/load.bash
@@ -1,0 +1,89 @@
+# Alternative source to load, uses specific bash extensions to the POSIX shell (BASH_SOURCE, <<<) hence the .bash extension
+# Usage in ~/.bashrc: source ~/path/to/volta/load.bash
+# Example: source ~/.volta/load.bash
+# Or: [[ -s ~/.volta/load.bash ]] && source ~/.volta/load.bash
+
+# TODO: test on windows, window WSL and osx. hahahaha.
+
+
+# Example usage in bash:
+# $ source .volta/load.bash # source it directly
+
+# $ echo $VOLTA_HOME # was VOLTA_HOME defined?
+# /home/mcarifio/.volta
+
+# $ tr ':' '\n' <<< $PATH|grep volta # was the volta bin added to PATH?
+# /home/mcarifio/.volta/bin
+
+# $ volta --version # does the function run?
+# 0.6.3
+# volta which node # does $VOLTA_HOME/shim ... shim?
+# /home/mcarifio/.volta/tools/image/node/12.10.0/6.10.3/bin/node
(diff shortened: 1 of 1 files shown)


shell/unix/load.bash
```diff
@@ -0,0 +1,89 @@
...
+#  and you don't have to keep track of them.
+_volta_setup() {
+    local me=$(realpath ${BASH_SOURCE})
+    local here=${me%/*}
+    export VOLTA_HOME=${VOLTA_HOME:-${here}}
+    if [[ -d ${VOLTA_HOME} ]] ; then
+        local bin=${VOLTA_HOME}/bin
+        [[ -d ${bin} ]] || _volta_error "Expecting ${bin}, not found."
+        # Add to PATH iff not already there. Next line is hack city.
+        if ! tr ':' '\n' <<< ${PATH} | grep --silent --no-messages ${bin} ; then
```
charlespierce: This line is pretty impossible to understand. If we're trying to check that the `bin` isn't already on the path, can we do that with a regex match, or even a simpler pattern match:

```
if [[ "${PATH}" != *$bin* ]]; then
```

Or something like that? I'll admit to being a novice at shell scripting, so if that's not possible, please let me know :smile:

shell/unix/load.bash
```diff
@@ -0,0 +1,89 @@
...
+    return 1  # not exit otherwise you exit the current session
+}
+
+# Expecting bash.
+[[ -z "${BASH}" ]] && _volta_error "Expecting bash."
+
+# Make _setup a function so you don't pollute the environment with working variables like `me` and `here`
+#  and you don't have to keep track of them.
+_volta_setup() {
+    local me=$(realpath ${BASH_SOURCE})
```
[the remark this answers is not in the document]
mcarifio: The function provides a scope for local variables, so you don't have to unset them explicitly. And more generally the caller/sourcer doesn't have to explicitly export VOLTA_HOME, it's created during the `source`.

shell/unix/load.bash
mcarifio: Yes. You could do this with a re match of ':\b*${bin}:\b*' on PATH. Bash regular expressions are also a little persnicky, so a reader will need to google stuff regardless.

mcarifio: tr gone, re instead

shell/unix/load.bash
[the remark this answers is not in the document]
mcarifio: Not sure what github means by "outdated" above. My comments remain the same.

PR #552, opened 2019-09-20
Fix tracking issue url of human fallback

CYBAI:


rwjblue:
Good catch, thank you!

chriskrycho:
DOH! This was my failure for sure; thank you for fixing it!

Issue #540, opened 2019-08-26
Investigate using local `.npmrc` settings for global installs


mikrostew:
Currently, if you try to install a package from a local repository which also has dependencies that are only in that local repo, you need a workaround:

```
$ NPM_CONFIG_USERCONFIG="$PWD/.npmrc" volta install some-local-pkg
```

This is because the npm config information is used for the package download, but not for installing its dependencies. Installing the dependencies uses the public npm repo, so anything that is not available there will fail.

We need a way to preserve the npm configuration and use it for the dependency install as well.

hugo-marques-m3:
I'm not sure if this is the same issue as @mikrostew is reporting but I can't seem to be able to install any package from an authenticated private repo. I've tried your the workaround and it doesn't seem to work.

charlespierce:
@hugo-marques-m3 Thanks for adding your experience. To dig in a little deeper, when you say it's an "Authenticated private repo", does that mean that in your non-Volta workflow, when you run `npm i -g <package>` it asks for your credentials and you have to type them in? Or are your credentials set in some way in your `.npmrc` file?

If it's the former, what is the behavior you see from `volta install <package>`, does it hang for a long time, or does it error out immediately?

hugo-marques-m3:
Hi @charlespierce, I have the following in a `.npmrc` where REGISTRY and TOKEN are the right values:
```
//{REGISTRY}/:_authToken="{TOKEN}"
registry=https://{REGISTRY}/
always-auth=true
```
It falls over immediately with a 403 response.

fusion2004:
@charlespierce @hugo-marques-m3 I am seeing the exact same thing. We are attempting to build and globally use a cli tool we have in a gemfury private registry, which requires a token to interact with (in the config format that @hugo-marques-m3 provided).

From what I can tell from looking at the yarn & npm source code, it looks like the registry expects an `Authorization` header sent with `Bearer {TOKEN}` (replacing `{TOKEN}` with the values in our configuration, as per @hugo-marques-m3's example) when downloading the tarballs.

I can see where that download happens in this chain:
https://github.com/volta-cli/volta/blob/b37e0889853f6884ae7e3191576529a06262dcae/crates/volta-core/src/tool/package/fetch.rs#L38
https://github.com/volta-cli/volta/blob/b37e0889853f6884ae7e3191576529a06262dcae/crates/volta-core/src/tool/package/fetch.rs#L84
[...]

PR #554, opened 2019-10-03
Use npm pack to retrieve tarballs

fusion2004:
Hello! This is my take on implementing #540. This is my first time writing rust for a non-toy app, so please give me any feedback you can on how I can improve this, especially from a code quality perspective.

The bulk of the work is in the `fetch_remote_distro` function, and it feels very large, but I'm not exactly sure the best way to split it up.

Compiling gives me a lot of warnings throughout the codebase about `help: use 'dyn': 'dyn Archive'`, should I fix those or leave them alone?


crates/volta-core/src/run/mod.rs
```diff
@@ -136,13 +141,20 @@ fn tool_name_from_file_name(file_name: &OsStr) -> OsString {
     }
 }
 
-fn command_for<A>(exe: &OsStr, args: A, path_var: &OsStr) -> Command
+fn command_for<A>(exe: &OsStr, args: A, path_var: &OsStr, current_dir: Option<&Path>) -> Command
```
fusion2004: I don't really think this is the best way to add this option, but I wasn't sure how to do better. I attempted extracting setting this into a separate function, but then I received compiler errors about the command being moved into my new function and I was not sure how to fix it.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+    let metadatas: Vec<super::serial::NpmPackData> = serde_json::de::from_str(&response_json)
+        .with_context(|_| ErrorDetails::NpmPackMetadataParseError {
+            package: name.to_string(),
+        })?;
+
+    debug!("Finding the first/only record from the output");
+    let metadata: super::serial::NpmPackData = match metadatas.iter().next() {
+        Some(data) => data.clone(),
+        // TODO: Make this be a correct error
+        None => throw!(ErrorDetails::PackageNotFound {
```
fusion2004: I'm not really sure what to do with the error handling here, this would be a case where we found the package in the registry, `npm pack` did something and exited successfully, but the json array it returns had nothing in it.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+    debug!("Moving the tarball to the expected path");
+    rename(tarball_from_npm_pack, path).with_context(|_| ErrorDetails::SetupToolImageError {
+        tool: name.into(),
+        version: details.version.to_string(),
+        dir: dir.to_path_buf(),
+    })?;
+
+    debug!("Attempting to load the now cached tarball from disk");
+    // TODO: Make this be a correct error
+    let distro = File::open(path).with_context(|_| ErrorDetails::PackageNotFound {
```
fusion2004: I wasn't sure what error to use or create for this line or the next error on the `Tarball::load` just after this.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+    version: &str,
+    session: &mut Session,
+    current_dir: Option<&Path>,
+) -> Fallible<ToolCommand> {
+    let args = vec![
+        OsString::from("pack"),
+        OsString::from("--json"),
+        OsString::from(format!("{}@{}", name, version)),
+    ];
+    run::npm::command(args, session, current_dir)
```
charlespierce: Instead of threading the `current_dir` through all of the unrelated `ToolCommand` code, what we can instead do is add a `current_dir` method to the `impl ToolCommand` block in `volta-core/src/run/mod.rs`:

```rust
pub(crate) fn current_dir<P: AsRef<Path>>(&mut self, dir: P) -> &mut ToolCommand {
    self.command.current_dir(dir);
    self
}
```

Since we aren't threading an `Option` through all of the sub-commands, we can change this function to take the `&Path` directly, instead of an option (since it's always used), and then here we can do:

```rust
let mut command = run::npm::command(args, session);
command.current_dir(current_dir);
command
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -59,15 +59,15 @@ impl Package {
         Package { name, details }
     }
 
-    fn fetch_internal(&self) -> Fallible<()> {
+    fn fetch_internal(&self, mut session: &mut Session) -> Fallible<()> {
```
charlespierce: I believe the `mut` prefix before the variable name is extraneous:

```suggestion
    fn fetch_internal(&self, session: &mut Session) -> Fallible<()> {
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -59,15 +59,15 @@ impl Package {
...
         // ISSUE(#288) - Once we have a valid Collection, we can check that in the same way as node/yarn
         // Until then, we use the existence of the image directory as the indicator that the package is
         // already fetched
         if path::package_image_dir(&self.name, &self.details.version.to_string())?.exists() {
             debug_already_fetched(self);
             Ok(())
         } else {
-            fetch::fetch(&self.name, &self.details)
+            fetch::fetch(&self.name, &self.details, &mut session)
```
charlespierce: And here, since the session is already a mutable reference, we can pass it along directly:

```suggestion
            fetch::fetch(&self.name, &self.details, session)
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -84,18 +84,18 @@ impl Package {
 }
 
 impl Tool for Package {
-    fn fetch(self, _session: &mut Session) -> Fallible<()> {
-        self.fetch_internal()?;
+    fn fetch(self, mut session: &mut Session) -> Fallible<()> {
```
charlespierce: Similarly here

```suggestion
    fn fetch(self, session: &mut Session) -> Fallible<()> {
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -84,18 +84,18 @@ impl Package {
...
+        self.fetch_internal(&mut session)?;
```
charlespierce: Here as well:

```suggestion
        self.fetch_internal(session)?;
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -84,18 +84,18 @@ impl Package {
...
 
         info_fetched(self);
         Ok(())
     }
-    fn install(self, session: &mut Session) -> Fallible<()> {
+    fn install(self, mut session: &mut Session) -> Fallible<()> {
```
charlespierce:
```suggestion
    fn install(self, session: &mut Session) -> Fallible<()> {
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -84,18 +84,18 @@ impl Package {
...
         if self.is_installed() {
             info!("Package {} is already installed", self);
             Ok(())
         } else {
-            self.fetch_internal()?;
+            self.fetch_internal(&mut session)?;
```
charlespierce:
```suggestion
            self.fetch_internal(session)?;
```

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+fn npm_pack_command_for(
+    name: &str,
+    version: &str,
+    session: &mut Session,
+    current_dir: Option<&Path>,
+) -> Fallible<ToolCommand> {
+    let args = vec![
+        OsString::from("pack"),
+        OsString::from("--json"),
+        OsString::from(format!("{}@{}", name, version)),
```
charlespierce: Since we only care about the file name, and not any of the other metadata that is output by `npm pack`, I think we can actually simplify this and use the standard, non-JSON output (which is already only the file name). We should also turn off the npm update notifier, since that's not relevant to our goal here:

```rust
let args = vec![
    OsString::from("pack"),
    OsString::from("--no-update-notifier"),
    OsString::from(format!("{}@{}", name, version)),
];
```

Then the `stdout` for the subcommand will be the filename

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+        path: path.to_path_buf(),
+    })?;
+
+    let dir = path.parent().unwrap();
+
+    let command = npm_pack_command_for(name, &details.version.to_string()[..], session, Some(dir))?;
+    debug!("Running command: `{:?}`", command);
+
+    let spinner = progress_spinner(&format!(
+        "Downloading via {} npm pack to {}",
```
charlespierce: Small wording change: On the spinner, we shouldn't show the download directory or the fact that we use `npm pack`, that's an implementation detail that doesn't matter to the user (though we should still have a `debug!` message showing where it is downloading to). Here, we should just show that we are downloading the tool version:

```suggestion
        "Downloading {}",
```

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+    spinner.finish_and_clear();
+
+    if !output.status.success() {
+        debug!(
+            "Command failed, stderr is:\n{}",
+            String::from_utf8_lossy(&output.stderr).to_string()
+        );
+        debug!("Exit code is {:?}", output.status.code());
+        throw!(ErrorDetails::NpmPackMetadataFetchError {
+            package: name.to_string(),
```
charlespierce: Here, since we have the resolved version as well, we can do `tool_version` to show the full name and version in the error message.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -79,9 +76,90 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+        });
+    }
+
+    let response_json = String::from_utf8_lossy(&output.stdout);
```
charlespierce: If we make the change below to call `npm pack` without the `--json` flag, then I think we can reduce this block of reading, parsing, and looking for the filename with:

```rust
let filename = String::from_utf8_lossy(&output.stdout);
```

We then will want to make sure that the filename isn't empty and that the file exists before continuing.

crates/volta-core/src/tool/package/fetch.rs
charlespierce: I think we probably want a new error in the `ErrorDetails` enum, that can be used for if the filename is blank, the file doesn't exist, or we can't load it. Something like `NpmPackUnpackError` that says:

```
Could not read archive for '{}' from npm pack.

{}
```

Where the CTA is `PERMISSIONS_CTA`, which we use as a generic catch-all for when something goes wrong with files on the file system.

crates/volta-core/src/tool/package/fetch.rs
charlespierce: See above, we can use the same error in all of these cases.

crates/volta-core/src/tool/package/serial.rs
```diff
@@ -285,3 +285,34 @@ impl From<BinLoader> for RawBinLoader {
...
+//       "mode": 420
+//     },
+//     (and lots more files...)
+//    ],
+//   "entryCount": 290,
+//   "bundled": []
+// }
+//
+#[derive(Deserialize, Clone, Debug)]
+pub struct NpmPackData {
```
charlespierce: If we switch to using `npm pack` without `--json`, then we don't need this data structure.

crates/volta-core/src/tool/package/fetch.rs
fusion2004: Yeah, this is what I was trying to do, and did implement once, but I had issues with command being moved and everything broke. I'm sure I just had my `&`s or my `mut`s wrong in some way, so I'll give your provided code a try!

fusion2004:
Thanks for the feedback, @charlespierce! I will try to make those updates soon.

charlespierce:
@fusion2004 It looks like I created a merge conflict for you. The main things that need to change are:
```rust
use crate::layout::volta_home;
```
instead of
```rust
use crate::path;
```

And the `dyn Archive` change that you noted, for the `fetch_remote_distro` function.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -80,9 +77,95 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+    path: &Path,
+    name: &str,
+    details: &PackageDetails,
+    session: &mut Session,
+) -> Fallible<Box<dyn Archive>> {
+    ensure_containing_dir_exists(&path).with_context(|_| ErrorDetails::ContainingDirError {
+        path: path.to_path_buf(),
+    })?;
+
+    let dir = path.parent().unwrap();
```
charlespierce: Can we add a comment here noting that the `parent()` will _always_ be `Some` because `ensure_containing_dir_exists` will error if there isn't a containing dir? We generally try to avoid using `unwrap` since it can potentially panic, but in this case, we know for sure that if we got here then it won't panic.

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -80,9 +77,95 @@ fn load_cached_distro(file: &Path, shasum_file: &Path) -> Option<Box<dyn Archive
...
+        });
+    }
+
+    debug!(
+        "Moving the tarball from {:?} to the expected path {:?}",
+        tarball_from_npm_pack, path
+    );
+    rename(tarball_from_npm_pack, path).with_context(|_| ErrorDetails::NpmPackUnpackError {
+        package: tool_version(name, details.version.to_string()),
+    })?;
```
charlespierce: We probably won't need to do this most of the time, since generally the file name we get from `npm pack` should match the one we generate internally. Given that, can we do a check before the `rename` to make sure that the two file paths are different?

`if tarball_from_npm_pack != path {` should work, it looks like `Path` implements `PartialEq`

fusion2004: Ah, that makes sense. My use case is a scoped package, where the naming is different. Will do! 👍

fusion2004:
@charlespierce Done and done!

Issue #222, opened 2018-12-18
Shell agnosticism


dherman:
We should make Notion work by default in all shells by having a reasonable install process and functionality set that works no matter what.

We've reduced the set of Notion's features that need to mutate the console environment to just `activate` and `deactivate`, which are really just conveniences rather than core functionality. We can make those commands simply fail with a reasonable error message by default, in not-fully-supported shells.

What should the installation logic be? Should it check for startup scripts for all supported shells and add the `PATH` and `NOTION_HOME` lines? Should it only do it for the current running shell? Should it print out instructions on manually adding those env vars?

Maybe the installer should print a list of all shells it detected and attempted to update, and then follow it with instructions for manually installing it in case of other shells.

PR #556, opened 2019-10-07
Remove 'volta activate' and 'volta deactivate'

### LINKED ISSUE #99: escape `PATH` when writing postscript

### LINKED ISSUE #104: Don't show an active node version after deactivating
## Reproduction

If I run `notion deactivate`, it still shows the current node version as being active:

```
$ notion deactivate

$ notion current
global: v6.14.3 (active)
```

But because notion has removed the shims from the path, my terminal session is actually using whatever is on the path (system node in my case):

```
$ which node
/usr/local/bin/node
$ node --version
v10.1.0
```

## Expectation

I expect that `notion current` will not show an active version of node if notion has been deactivated.

### LINKED ISSUE #165: deactivate prompts for file removal (depending on user profile settings)
I needed to deactivate `notion` for a particular shell session (due to some missing features that are still being worked on), and was somewhat surprised by this prompt to delete a file:

```
❯ notion deactivate                                                                                                                     
remove /Users/rjackson/.notion/tmp/notion_tmp_1280132002.sh?
```

I am not quite certain how things are structured (and why we are deleting a file here), but it seems like we should probably try to avoid prompting the user to delete here.

### LINKED ISSUE #216: env privilege escalation through postscript protocol
The "postscript" protocol conveys read/write access to the user's console environment by executing an arbitrary script provided by the delegated `notion` executable in the user's console environment. If there are any places where this access is granted to untrusted code (such as via `notion exec`), that access could potentially be leaked to the untrusted code.

I'm not sure how much damage can be done with full access to the console's environment, but we shouldn't ever give more privileges than necessary.

I believe we should be able to lock this down by only using the protocol when the wrapper script detects that the command is `activate` or `deactivate`, which are the only commands that require this access. We should also document in those commands that they must never invoke untrusted code since they have increased env privileges.

This should have the additional benefit of shaving some boot time latency off, since generating the random bits for the postscript variable takes time, which we can limit to only the (off-the-beaten-path) `activate`/`deactivate` commands.

If we want a bit more paranoia, we could add extra logic to any commands that call untrusted code to censor the `NOTION_POSTSCRIPT` environment variable first. (I doubt that should affect measurable performance.)

charlespierce:
Related to #222 
Closes #216 
Closes #165 
Closes #104 
Closes #99

Info
-----
* As part of the push towards automatic updates, we want to enable Volta to work directly on the command line, instead of through a shell function wrapper.
* The purpose of the function wrapper was to allow the `volta activate` and `volta deactivate` commands to work as expected, modifying the PATH environment variable for the current shell.
* However, those commands were intended as a release valve for when we were in early testing phases, and are no longer needed since Volta is much more stable now.

Changes
-----
* Removed the `activate` and `deactivate` commands from the Volta CLI and the associated `ActivityType`
* Moved the `VOLTA_UNSAFE_GLOBAL` environment variable definition to the `run` modules, where it is used.
* Removed the `shell` and `env` modules, which previous existed primarily to support the Postscript behavior that is no longer needed without `activate` / `deactivate`
* Removed `System::enabled_path`, which was only used by `volta activate` before.
* Deleted tests that referenced no-longer-available blocks / commands.

Tested
-----
* Verified that `volta activate` and `volta deactivate` are no longer allowed commands
* All automated tests pass

Notes
-----
* This PR targets a feature branch (`volta_update`), as many of the changes to support automatic updates are interconnected, but can be completed / reviewed separately. The feature branch will allow that work to be reviewed in smaller chunks, as opposed to all at once.


charlespierce:
Closing in favor of using a feature flag for updates, instead of a feature branch

charlespierce:
Reopening as on further consideration I prefer the feature branch approach at the moment, since it allows for more parallel development and easier diffs.

PR #558, opened 2019-10-08
[volta updates] Add volta-updates feature and CI override

charlespierce:
Info
-----
* To support working on Volta updates in parallel with other changes, we want to hide the updates work behind a feature flag, which is well supported by `cargo`
* We also want PRs involving the update work to test that work as it goes in, but don't want the update tests to break non-updates CI.

Changes
-----
* Added a `volta-updates` feature flag to `Cargo.toml` for the root project and for `volta-core`
* Added a CI override `[volta updates]` in the commit message that will run the automated tests with the feature flag enabled.
* Added new jobs to `azure-pipelines.yml` to enable the feature flag tests only if the CI override is detected.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ cross-platform-docs = ["volta-core/cross-platform-docs"]
 mock-network = ["mockito", "volta-core/mock-network"]
 volta-dev = []
 smoke-tests = []
+volta-updates = ["volta-core/volta-updates"]
 
 [[bin]]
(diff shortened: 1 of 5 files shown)


charlespierce:
While this works, closing in favor of the feature branch approach I was originally going to use. That will have some overhead in keeping things correctly rebased, but allows the work to start now and the PRs will be easier to understand as they will be simple additions or deletions, and not wrapped with feature-flag checks

PR #561, opened 2019-10-10
[Updates] Refactor layout to support having the binaries on the PATH

### LINKED ISSUE #563: [Updates] Change layout to support installing Volta on the PATH
We will need to make a couple of fundamental changes to the `layout` and logic of Volta to support having the binaries directly on the path:

* Improving the detection of the install directory
* Renaming `shim` to avoid conflicts with other installed tools

charlespierce:
Closes #563 

- [x] Use the currently running executable to find the `install_dir`
- [x] Rename `shim` to `volta-shim`
- [x] Add custom error for when `volta-shim` is called directly

Info
-----
* To support automatic updates and managed environments, we want to allow Volta to be installed directly on the users' PATH, instead of using a shell function wrapper to find it.
* Along with putting the binaries on the PATH, we want to rename `shim` to `volta-shim`, so that it doesn't conflict with any non-Volta binaries that may already exist on the system.

Changes
-----
* Updated the logic for finding the Volta Install directory to use the directory of the currently running executable, as opposed to the bespoke logic for each OS. Since the binaries will be installed side-by-side and can theoretically be anywhere on the PATH, we can use this approach to find them no matter where they are installed.
* Renamed `shim` to `volta-shim`, including in a new `v1` layout for the project.
* Added a custom error to show the user if they ever call `volta-shim` directly, letting them know that it should only be called indirectly through one of the shim symlinks.

Notes
-----
* All changes made behind the `volta-updates` feature flag
* No changes were made to any installers yet, as we need to build the `volta setup` command before we can redo the installer.
* This explicitly isn't handling the case of upgrading from a legacy Volta installation, that will be added in a different set of changes


crates/volta-core/src/platform/mod.rs
```diff
@@ -220,6 +220,9 @@ mod test {
         let mut pathbufs: Vec<PathBuf> = Vec::new();
         pathbufs.push(volta_home().unwrap().shim_dir().to_owned());
         pathbufs.push(PathBuf::from("C:\\\\somebin"));
+        #[cfg(feature = "volta-updates")]
```
chriskrycho: Does `cfg_if!` work here? (I don't know, to be clear, it's an actual question!)

```rust
cfg_if! {
    if #[cfg(feature = "volta-updates")] {
        pathbufs.push(volta_install().unwrap().root().to_owned());
    } else {
        pathbufs.push(volta_install().unwrap().bin_dir());
    }
}
```

*If* that works, I think it makes the intent a lot clearer (for as long as this is in here).

(Same question/suggestion for the similar bits below.)

crates/volta-layout/src/v1.rs
```diff
@@ -0,0 +1,10 @@
+use volta_layout_macro::layout;
+
+pub use super::v0::VoltaHome;
```
chriskrycho: This may just be my not following how the macro works (because I haven't dug into it at all), but what is this `pub use` here for?

src/execute.rs
```diff
@@ -0,0 +1,35 @@
+use volta_core::error::report_error;
+use volta_core::log::{LogContext, LogVerbosity, Logger};
+use volta_core::run::execute_tool;
+use volta_core::session::{ActivityKind, Session};
+use volta_core::signal::setup_signal_handler;
+use volta_fail::ExitCode;
+
+#[inline]
```
chriskrycho: Why do we need this to be explicitly inlined? Can we add an explanatory comment so it's obvious to other later readers of the code? ❤️

src/execute.rs
```diff
@@ -0,0 +1,35 @@
...
+pub(super) fn run_shim() {
+    Logger::init(LogContext::Shim, LogVerbosity::Default)
+        .expect("Only a single Logger should be initialized");
```
chriskrycho: Is there any reason besides extra loggers being set up why this would fail? My concern is that dying with `"Only a single Logger should be initialized"` might be misleading if there *are* other reasons it would fail.

chriskrycho: (I note that this is just copied over to support the change – so no worries about making any change here *for this PR* even if we do think we should.)

crates/volta-core/src/platform/mod.rs
charlespierce: It looks like it does! I'll update them to use that, good catch!

crates/volta-layout/src/v1.rs
charlespierce: Since at this stage there aren't any changes to the `VoltaHome` part of the layout, this is re-exporting the `v0` Layout to be part of `v1`. That lets us use `v1::VoltaHome` and `v1::VoltaInstall` as the two layout structs everywhere, instead of having a mixed state of `v0::VoltaHome` and `v1::VoltaInstall`

This also changes in one of the later PRs, once changes are actually made to the layout, so this is replaced by the actual `VoltaHome` struct.

src/execute.rs
charlespierce: We probably don't, and it's likely optimized by the compiler anyway since the two binaries that call this are single function calls with nothing else anyway. I'll remove it :+1:

src/execute.rs
charlespierce: Per the [docs for 'log'](https://docs.rs/log/0.4.8/log/fn.set_boxed_logger.html), that is the only failure path for the log initialization, yes.

crates/volta-core/src/platform/mod.rs
charlespierce: I lied, that doesn't actually work. All of those blocks were Windows-only, so _of course_ there weren't any compiler errors on Mac 😅

src/execute.rs
chriskrycho: Awesome! Thank you!

crates/volta-layout/src/v1.rs
chriskrycho: Ahhhh, got it. Thanks for clarifying – I was reading that very PR so it makes sense!

crates/volta-core/src/platform/mod.rs
chriskrycho: Hahahahaha 😭 Worth a try – thanks for checking!

Issue #562, opened 2019-10-11
[Updates] Remove 'volta activate' and 'volta deactivate'


charlespierce:
`volta deactivate` (and later `volta activate`) are commands that were added early on as an escape hatch for when Volta wasn't working right and we needed to give a quick escape for our beta users. As Volta has matured and stabilized, they have become less important.

Making these two commands work requires additional setup and for `volta` to be implemented as a shell function wrapper that calls out to the Volta binary. This implementation is major roadblock in supporting #222, as well as resulting in a number of other issues: #216, #165, #104, and #99.

Beyond the issues that result from the function wrappers, they also make it more difficult to install Volta in a "traditional" way by putting the binary directly on the `PATH`. That is an approach we want to take to support [automatic updates](https://github.com/volta-cli/rfcs/pull/37), so we also have pressure from that direction to remove the function wrapper (and as a result, the `activate` and `deactivate` commands).

For all these reasons, the commands `volta activate` and `volta deactivate` should be deprecated and then removed.

rwjblue:
I'd like to see this considered blocked on https://github.com/volta-cli/volta/issues/292. Without a way to `volta pin npm` _or_ `volta deactivate`, I'd have to modify my shell config in order to debug issues with specific `npm` versions (e.g. https://github.com/emberjs/data/issues/6589).

PR #571, opened 2019-10-15
Add deprecation warnings to activate and deactivate

charlespierce:
Suggestion from @thoov in #559: Add deprecation warning to `volta activate` and `volta deactivate` for the `0.6.x` releases, so that users get a warning that they are going away.

Then when we remove them in `0.7.0` with the update work, it won't be completely out of nowhere to the users.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -96,7 +96,7 @@ pub(crate) enum Subcommand {
     )]
     Current(command::Current),
 
-    /// Disables Volta in the current shell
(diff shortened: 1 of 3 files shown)


charlespierce:
Updated the description of #562 to more thoroughly explain why we are removing `volta activate` and `volta deactivate`, and included a link to that issue in the deprecation warning.

rwjblue:
Awesome, thank you!

PR #580, opened 2019-10-22
[Updates] Create new installers that use 'volta setup'

### LINKED ISSUE #565: [Updates] Simplify installers to provide the binaries and call `volta setup`
To avoid duplication of work, and take advantage of the migration / update features that are included in this overall project, we should simplify all of the installers (Unix shell, RPM, and Windows) to simply unpack the binaries and then call `volta setup` directly, allowing `volta setup` to handle the majority of the work.

charlespierce:
Closes #565 

Info
-----
* With `volta setup` created in #578, we can simplify the installers that we currently support to do 2 things:
    * Install the binary files (`volta` and `volta-shim`) onto the PATH
    * Call `volta setup` to handle creating the user directory and adding the Shim directory to the PATH

Changes
-----
* Added new installers for Windows (`updates.wxs`), Unix (`volta-install-updates.sh`), and RPM (`volta-updates.spec`).
* Copied and updated helper scripts (`release.sh` and `build-rpm.sh`) to support compiling with the `volta-updates` feature so that `volta setup` is available.

Notes
-----
* This PR builds on #578, so will remain a draft until that is merged

### DIFF
--- a/dev/rpm/build-rpm-updates.sh
+++ b/dev/rpm/build-rpm-updates.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# Build an RPM package for Volta
+
+# using the directions from https://rpm-packaging-guide.github.io/
+
+# exit on error
+set -e
+
+# only argument is the version number
+release_version="${1:?Must specify the release version, like \`build-rpm 1.2.3\`}"
+archive_filename="v${release_version}.tar.gz"
+
+# make sure these packages are installed
+# (https://rpm-packaging-guide.github.io/#prerequisites)
+sudo yum install gcc rpm-build rpm-devel rpmlint make python bash coreutils diffutils patch rpmdevtools
+
+# set up the directory layout for the RPM packaging workspace
+# (https://rpm-packaging-guide.github.io/#rpm-packaging-workspace)
+rpmdev-setuptree
+
+# create a tarball of the repo for the specified version
+# using prefix because the rpmbuild process expects a 'volta-<version>' directory
+# (https://rpm-packaging-guide.github.io/#putting-source-code-into-tarball)
+git archive --format=tar.gz --output=$archive_filename --prefix="volta-${release_version}/" HEAD
+
+# move the archive to the SOURCES dir, after cleaning it up
+# (https://rpm-packaging-guide.github.io/#working-with-spec-files)
+rm -rf "$HOME/rmpbuild/SOURCES/"*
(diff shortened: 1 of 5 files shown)


dev/rpm/build-rpm-updates.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
chriskrycho: @charlespierce do we have someone familiar with RPM who could give this some attention? I'm looking through it, but I have literally never done *anything* with RPM (and almost nothing with RHEL), so I'd very much prefer not to be the only reviewer on this!

dev/rpm/volta-updates.spec
```diff
@@ -0,0 +1,70 @@
...
+Version:        0.6.3
+Release:        1%{?dist}
+Summary:        The JavaScript Launcher ⚡
+
+License:        BSD 2-CLAUSE
+URL:            https://%{name}.sh
+Source0:        https://github.com/volta-cli/volta/archive/v%{version}.tar.gz
+
+# cargo is required, but installing from RPM is failing with libcrypto dep error
+# so you will have to install cargo manually to build this
```
chriskrycho: Are we documenting this in install docs as well?

dev/rpm/build-rpm-updates.sh
charlespierce: Most of this (and the below `volta-updates.spec`) were copied from `build-rpm.sh` and `volta.spec`, then adapted for the necessary changes (since we can't apply Rust feature flags to shell scripts). So the easiest way to review them is probably to diff with those files.

That said, @mikrostew wrote the originals, so he probably has the best context for review.

dev/rpm/volta-updates.spec
charlespierce: As above, @mikrostew probably has the most context on this, but I believe this comment is about an end user trying to use the RPM to build from source, not an install dep per se.

chriskrycho: Ahhhh, got it, makes sense. <3

dev/rpm/build-rpm-updates.sh
chriskrycho: OHHHH. That's good to know! Thanks!

charlespierce:
To facilitate the next (and last in this block!) review, I'll merge this, but we'll make sure @mikrostew takes a look at the diffs for the new installers before we unwrap the feature flag and make this the default behavior :+1:

PR #581, opened 2019-10-25
Write 'volta list' output to stdout directly

charlespierce:
We are currently using `info!` to write the output of `volta list`. However, that goes to `stderr` since it is part of the logging interface. Since the expected effect of running `volta list` is output to the terminal, we should directly write to stdout using `println!` instead.

This aligns with `volta which` also writing to stdout, since it is expected output instead of a purely informational note about what happened.


chriskrycho:
TIL! Nice fix.

PR #590, opened 2019-11-05
[Updates] Create 'volta-migrate' binary to handle layout migrations

### LINKED ISSUE #566: [Updates] Create `volta-migrate` crate / binary
Create a separate crate / binary to handle migrating the Volta layout from one version (or no version at all, in the case of a new install) to the current version. These can be kept separate so that the main binaries don't grow unbounded as we make changes.

The only migration this should support at first is the `v0` -> `v1` migration, which may also include removing some files that aren't referenced from the `v0` layout (the old `volta` binary and the `load.*` files).

charlespierce:
Closes #566 

Info
-----
* To support background updates (either automatic or managed) when the layout of the Volta directory may change between versions, we need a way to migrate from one version of a layout to the most recent.
* This migration should be checked on every execution, to make sure that the Volta system is up-to-date with the executables, however that check should be fast so as not to adversely affect the hot-path of shims.
* The migration itself should be handled by a separate executable, to prevent the primary executables (`volta` and `volta-shim`) from growing unbounded in size as we add more migrations.

Changes
-----
* Added a new executable `volta-migrate` that is responsible for performing the actual migration. This executable is called by `volta` and the shims if a discrepancy is detected between the binaries and the current layout.
* Added a new crate `volta-migrate` that contains the migration logic, including detecting the specific current layout and executing any migrations that are needed.
* Updated the installers to include `volta-migrate` along side the other binaries.
* Added acceptance tests for the different migration cases (currently they are very similar because the migration is primarily only creating the layout file).

Notes
-----
* This PR builds on #580, so it will remain a draft until that is merged.


rwjblue:
> This migration should be checked on every execution, to make sure that the Volta system is up-to-date with the executables, however that check should be fast so as not to adversely affect the hot-path of shims.

This particular point is super important to me, can you explain (at a high level) how the check can be done without doing a bunch of extra I/O during the "hot path" of a specific shim/tool execution?

charlespierce:
@rwjblue Absolutely! We're using a marker file that indicates the current layout (at the moment, it's `v1`, so the file is `layout.v1`) in the root of `~/.volta`. The hot paths check if that file exists. If it does, we're on the latest layout and we continue as normal. If it doesn't, then we launch `volta-migrate`, which does more involved logic to determine whether this is an empty layout or a previous version, and then does the appropriate migration.

rwjblue:
> We're using a marker file that indicates the current layout (at the moment, it's v1, so the file is layout.v1) in the root of ~/.volta. The hot paths check if that file exists.

Gotcha, thank you! So I'm still confused on the exact I/O operations that we expect in various cases. Instead of guessing, I thought maybe we could chat through various scenarios, do you mind?

Assuming that is fine, here are a few scenarios that I had in mind:

* The current `~/.volta` installation is on v1, but the the installed binaries are from v2. What I/O will be done? (from your reply it seems like a single `fs.stat` to see if `layout.v1` exists, and update?)
* The current `~/.volta` installation is on v2, and the binaries are on v2; what I/O operations do we do? `fs.stat('layout.v1')` and `fs.stat('layout.v2')`?
* The current `~/.volta` installation is on v1, and the binaries are on v5; what I/O operations do we do?
* The current `~/.volta` is "empty" (or missing?), and the binaries are on v2; what does it do?

rwjblue:
Hmm, as I re-read what you said in your reply:

> We're using a marker file that indicates the current layout (at the moment, it's v1, so the file is layout.v1) in the root of ~/.volta. The hot paths check if that file exists. If it does, we're on the latest layout and we continue as normal.

It seems that I misunderstood/misread. Is it correct to say that in the hot path (e.g. `node` or other tool invocation) that we would only check for the **current** layout version file (e.g. in `v2` binary, we only check for `~/.volta/layout.v2`) and if not found we launch the more expensive tool (presumably only one time)?

charlespierce:
@rwjblue That’s it exactly. To ensure we’re on the correct layout version, we check for the _current_ layout file. If that isn’t there, then we launch `volta-migrate`, which has more complicated and expensive logic to determine what migrations are needed and carry them out.

charlespierce:
Closing and reopening to debug CI not starting

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,134 @@
+#![cfg(feature = "volta-updates")]
+use std::collections::HashSet;
+use std::convert::TryInto;
+use std::fs::{DirEntry, Metadata};
+use std::path::Path;
+
+mod empty;
+mod v0;
+mod v1;
```
chriskrycho: Style suggestion for here and below: if we follow this up by `use v0::V0` and `use v1::V1`, it makes a *lot* of the rest of the module read a *lot* more cleanly. Thoughts?

crates/volta-migrate/src/v1.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        debug!("Writing layout marker file");
+        File::create(self.home.layout_file()).with_context(|_| {
+            ErrorDetails::CreateLayoutFileError {
+                file: self.home.layout_file().to_owned(),
+            }
+        })?;
+
+        Ok(())
+    }
```
chriskrycho: The name `finalize` here makes me think of this in terms of a builder-style item; that in turn makes me wonder if this should be `Fallible<&Self>`? Probably a premature refactoring given I can't actually think of what you'd do with it *afterward*, unless maybe something like `some_v1.finalize().home` or something.

A related, but less trivial, question I have here: is it valid to use a `V1` *without* its being finalized? If it's *not*, what's your thought on *actually* having a builder so that it's impossible to end up with a type that *seems* like everything is good to go, but where it's out of sync with the state of the file system?

crates/volta-migrate/src/v1.rs
```diff
@@ -0,0 +1,105 @@
...
+
+            let old_shim_bin = new_home.root().join("shim");
+            if old_shim_bin.exists() {
+                remove_file(&old_shim_bin)
+                    .with_context(|_| ErrorDetails::DeleteFileError { file: old_shim_bin })?;
+            }
+        }
+
+        Ok(V1 { home: new_home })
+    }
```
chriskrycho: Thinking about this entire flow—does it account for the pernicious failure condition of something happening that kills the operation mid-process? For example: if the user <kbd>Ctrl</kbd><kbd>C</kbd>s for some reason, right in the middle of this happening, does it end up in a state that lets us migrate successfully afterward, or is the user just out of luck?

src/common.rs
```diff
@@ -0,0 +1,46 @@
...
+            .with_context(|_| ErrorDetails::CouldNotStartMigration)
+            .into_result()?;
+    }
+
+    Ok(())
+}
+
+pub trait IntoResult<T> {
+    fn into_result(self) -> Result<T, Error>;
+}
```
chriskrycho: Could this just be a `From` impl? (I imagine no, given this exists, but I'm curious why not!)

crates/volta-core/src/layout/mod.rs
```diff
@@ -39,6 +46,7 @@ pub fn volta_home<'a>() -> Fallible<&'a VoltaHome> {
 // NOTE: This initialization will, on some code paths, call volta_home()
 // We need to make sure that volta_home does not in turn call this method
 // or we will run into problems with deadlocks
+// TODO: Remove this comment when we enable Volta updates - No longer true
```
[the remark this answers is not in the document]
charlespierce: Good idea. I'll make an issue around removing the `volta-updates` feature flag, and tag that issue in the TODO comment.

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+use volta_core::layout::volta_home;
+#[cfg(unix)]
+use volta_core::layout::volta_install;
+use volta_core::shim;
+use volta_fail::{Fallible, ResultExt};
+
+enum MigrationState {
+    Empty(empty::Empty),
+    V0(Box<v0::V0>),
+    V1(Box<v1::V1>),
```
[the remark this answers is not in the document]
charlespierce: I'll add a comment, but Clippy complains about differently-sized enums without the Boxes :smile:

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+                - If yes, then V0
+                - If no, then check if $VOLTA_HOME/load.sh exists? If yes then V0
+        - Else Empty
+
+        The extra logic on Unix is necessary because Unix installs can be either inside or outside $VOLTA_HOME
+        If it is inside, then the directory necessarily must exist, so we can't use that as a determination.
+        If it is outside (and for Windows which is always outside), then if $VOLTA_HOME exists, it must be from a
+        previous, V0 installation.
+
+        Going forward, each new version will have an associated layout file, so we can use that as a discriminant
```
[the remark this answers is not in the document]
charlespierce: For sure! It took a lot of thought and false-starts to come up with the logic, wanted to make sure it was fully documented going forward.

crates/volta-migrate/src/empty.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub struct Empty {
+    pub home: PathBuf,
+}
+
+impl Empty {
+    pub fn new(home: PathBuf) -> Self {
+        Empty { home }
+    }
+}
```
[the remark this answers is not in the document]
charlespierce: Will do!

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,134 @@
...
+        */
+
+        let home = volta_home()?;
```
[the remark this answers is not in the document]
charlespierce: The `volta_home()` function will always be fallible, because determining the root involves fallible operations like determining the user's home directory from the OS (see https://github.com/volta-cli/volta/blob/master/crates/volta-core/src/layout/unix.rs#L8). Once we have the value, however, all the other operations are infallible, since they are only `Path` manipulations.

crates/volta-migrate/src/lib.rs
charlespierce: Good call, I'm not sure why I hadn't done that! I think I might have had competing names at one iteration of this code, but that's not the case any more, so no reason to use qualified names when `use` works!

crates/volta-migrate/src/v1.rs
charlespierce: `finalize` is probably a poor name, maybe `commit()` would be better? The operation here is the one that essentially marks the migration as "completed" by writing the layout file and updating the shims to their new target. Once this completes, any future migrations will see the layout file and work under the assumption that the file system is laid out as V1 expects. That's why it's the last thing to happen, to make sure that a failure anywhere along the migration process will allow it to re-start without issue.

crates/volta-migrate/src/v1.rs
charlespierce: See above, writing the layout file that marks the migration complete is always the last thing that is done, so that if something fails we can try again. It does mean that the migrations will likely need to be tolerant of restarting from a partial failure, however. This migration is because it is removing old files only if they exist, but when building future migrations we'll need to be aware of the potential for partial application. I'll add a comment to the top of the main file indicating that necessity.

src/common.rs
charlespierce: It can't due to orphaning rules, because we're converting from a `Result<T, E>` to a `Result<U, F>`, so both the source and destination are external types.

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,147 @@
+//! Provides types for modeling the current state of the Volta directory and for migrating between versions
+//!
+//! A new layout should be represented by it's own struct (as in the existing v0 or v1 modules)
+//! Migrations between types should be represented by `TryFrom` implementations between the layout types
+//! (see v1.rs for examples)
+//!
+//! NOTE: Since the layout file is written once the migration is complete, all migration implementations
```
dherman: So IIUC you're saying migrations are supposed to be transactional? I feel like that would mean they should have to write into a temp directory before completing, in case of failure partway through? But I don't see that happening. Am I misunderstanding?

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+use log::debug;
+use volta_core::error::ErrorDetails;
+use volta_core::fs::read_dir_eager;
+use volta_core::layout::volta_home;
+#[cfg(unix)]
+use volta_core::layout::volta_install;
+use volta_core::shim;
+use volta_fail::{Fallible, ResultExt};
+
+enum MigrationState {
```
dherman: It'd be helpful to have a comment explaining what a `MigrationState` represents. The state of the system when a migration was requested? The state of a migration at each step throughout the migration? ...?

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+                - If yes, then V0
+                - If no, then check if $VOLTA_HOME/load.sh exists? If yes then V0
+        - Else Empty
+
+        The extra logic on Unix is necessary because Unix installs can be either inside or outside $VOLTA_HOME
+        If it is inside, then the directory necessarily must exist, so we can't use that as a determination.
+        If it is outside (and for Windows which is always outside), then if $VOLTA_HOME exists, it must be from a
+        previous, V0 installation.
+
+        Going forward, each new version will have an associated layout file, so we can use that as a discriminant
```
dherman: Given the point made in this comment ("going forward, ..."), it feels like it would be more readable to split this top-level algorithm into two parts that delegate to two separate helper functions:

- infer legacy state
- detect branded state

The more complex logic is inferring the legacy state, so it's nice to cordon that off from the simple part.

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        Ok(MigrationState::Empty(empty::Empty::new(volta_home)))
+    }
+}
+
+pub fn run_migration() -> Fallible<()> {
+    let mut state = MigrationState::current()?;
+
+    loop {
```
dherman: If I'm reading this right, you don't need a loop at all. You could just have:

```rust
let v1 = match state {
    MigrationState::Empty(e) => e.try_into()?,
    MigrationState::V0(zero) => zero.try_into()?,
    MigrationState::V1(one) => one
};
v1.commit()?;
```

crates/volta-migrate/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        };
+    }
+
+    Ok(())
+}
+
+fn regenerate_shims_for_dir(dir: &Path) -> Fallible<()> {
+    debug!("Rebuilding shims");
+    for shim_name in get_shim_list_deduped(dir)?.iter() {
+        shim::delete(shim_name)?;
```
dherman: I don't quite see how deleting and recreating makes it possible to gracefully restart. Is the idea that anything you delete is recreate-able? I'm just a little nervous you could delete some data somewhere along the line that you can't recreate.

Would it be safer to recreate/copy everything fresh to a temp directory and then have an atomic (or almost-atomic, anyway) directory swap at the end?

src/volta-shim.rs
```diff
@@ -1,5 +1,38 @@
-mod execute;
+#![cfg(feature = "volta-updates")]
```
dherman: So, once we un-flag `volta-updates`, we will delete `shim.rs`?

src/main.rs
```diff
@@ -4,31 +4,75 @@ mod cli;
...
-    session.add_event_start(ActivityKind::Volta);
-    let exit_code = volta.run(&mut session).unwrap_or_else(|err| {
-        report_error(env!("CARGO_PKG_VERSION"), &err);
-        session.add_event_error(ActivityKind::Volta, &err);
-        err.exit_code()
-    });
-
-    session.add_event_end(ActivityKind::Volta, exit_code);
-    session.exit(exit_code);
+cfg_if! {
```
dherman: Is it possible to narrow the flagged differences here? It looks like there's a fair amount of duplication in the two branches of the `cfg_if`.

crates/volta-core/src/error/details.rs
```diff
@@ -590,13 +595,28 @@ Please remove the file or pass `-f` or `--force` to override.",
 {}",
                 REPORT_BUG_CTA
             ),
+            #[cfg(feature = "volta-updates")]
+            ErrorDetails::CouldNotStartMigration => write!(
+                f,
+                "Volta is out-of-date and could not start migration process.
```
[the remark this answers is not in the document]
charlespierce: Good catch, it's actually the config directory that's out of date. Unfortunately, due to how the happy path works (we only check for the most recent layout file), we don't have a clean story for determining which version the user's directory supports _without_ starting the Migration app (which is what this error is talking about being unable to do).

crates/volta-core/src/error/details.rs
```diff
@@ -590,13 +595,28 @@ Please remove the file or pass `-f` or `--force` to override.",
...
+
+Please ensure Volta was installed correctly."
```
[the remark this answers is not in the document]
charlespierce: I like the idea of asking them to directly run `volta-migrate`. That actually is probably the highest likelihood of success, since the issue is that `volta-migrate` wasn't able to start, which most often means it's either not there or it's there and not executable. Both would need to be fixed in order to run it, so that's a good CTA!

Cargo.toml
```diff
@@ -45,6 +50,7 @@ test-support = { path = "crates/test-support" }
 textwrap = "0.11.0"
 which = "2.0.1"
 dirs = "1.0.4"
+volta-migrate = { path = "crates/volta-migrate", optional = true }
```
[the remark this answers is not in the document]
charlespierce: `volta-migrate` is the name of both the executable in the main app (in `src/volta-migrate.rs`) and the library crate that holds the bulk of the migration code (in `crates/volta-migrate/`). This is referring the latter, which we need to have as a dependency so that the former can call into it.

crates/volta-core/src/layout/mod.rs
```diff
@@ -74,6 +75,7 @@ fn default_install_dir() -> Fallible<PathBuf> {
         .with_context(|_| ErrorDetails::NoInstallDir)
 }
 
+#[cfg(not(feature = "volta-updates"))]
```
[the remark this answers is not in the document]
charlespierce: Can likely be a separate issue, and I think the "new" migration behavior fully encompasses the old behavior, since the old one only regenerated the config directory if the `~/.volta` folder didn't exist at all. That will still happen in this case, _and_ we can migrate from older to newer layouts.

The tricky part about any more involved "self-healing" behavior is that it requires a deeper inspection of the config directory, which we want to avoid on the hot path if possible.

crates/volta-core/src/shim.rs
```diff
@@ -44,11 +37,6 @@ pub fn create(shim_name: &str) -> Fallible<ShimResult> {
 }
 
 pub fn delete(shim_name: &str) -> Fallible<ShimResult> {
-    if !is_3p_shim(shim_name) {
```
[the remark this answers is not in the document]
charlespierce: Mentioned this in discord, but for completeness: Removing the error here is useful for removing and regenerating the shims during a migration. It's also the case that it wasn't possible to reach this error, as trying to run `volta uninstall node` or the like would error well before it got to any sort of shim removal.

crates/volta-migrate/src/lib.rs
charlespierce: My thought was less that they need to be transactional, and more that the migration method should be idempotent, so if it's interrupted we can try again without it erroring out due to being partially applied.

I'm happy to have a deeper discussion about the specific mechanics of migrations going forward, as this was definitely something I struggled with even while making this PR. My biggest concern with something like a temp directory is it would mean we need to copy the entire config directory, apply the changes to it, and then finally remove and replace the existing directory, which is a _lot_ more filesystem operations (especially on Windows / NTFS) than the current model.

crates/volta-migrate/src/lib.rs
charlespierce: Definitely will do! :+1:

crates/volta-migrate/src/lib.rs
charlespierce: I like that separation a lot, I'll add that!

crates/volta-migrate/src/lib.rs
charlespierce: We don't strictly need a loop at the moment, that's true. The looping behavior was put in thinking forward to future migrations. My thinking (which I realize wasn't well explained on this point) was that each layout version would only need to have migration code from the _previous_ layout version (and nothing, but that's usually a trivial implementation). Then the migration process would loop through one step at a time, migrating from V0 -> V1 -> V2 -> V3 -> ... -> VX.

This is slightly less efficient when doing the migration, but prevents us from having to write a migration from each previous version to the newest one going forward. That would result in each migration being incrementally more difficult to write than the previous ones.

src/volta-shim.rs
charlespierce: Yeah, that's the plan.

crates/volta-migrate/src/lib.rs
charlespierce: That's a good point that this isn't going support graceful restarting. Since the shims are symlinks, the only way (I believe) to re-target them is to remove and recreate them. For these specifically, I think we can probably do a temp directory version, since we can read through the existing directory, create whatever contents / shims we need in the temp, and then swap them over.

I initially wanted this (and the writing of the layout file) to be separated from individual migrations since it would potentially be duplicating work to have them run with each "step". However, thinking more deeply, we actually do want these operations to happen at each migration, because that lets the graceful restart actually apply happen: Any migrations that have completed will be skipped on subsequent restarts, leaving only the remaining work. It also prevents us from missing a regeneration of the shims or having to include it in every migration.

src/main.rs
charlespierce: Yeah, I'll look into it. I don't think `cfg_if` works well with random statements, but the differences are small enough that it may work fine with just regular `#[cfg]` conditionals

crates/volta-migrate/src/v1.rs
charlespierce: Based on feedback from @dherman and thinking through the flow a little more, I removed this method entirely, so the naming isn't a problem!

crates/volta-migrate/src/lib.rs
charlespierce: Letting this percolate overnight and looking at the code this morning, I'm actually not sure there is a more "restartable" solution. The issue with a temporary directory is that if we get a failure during removing the existing directory or while `rename`ing the temp directory, we lose _all_ the contents of the shim directory (which could also include the volta binaries themselves). While this approach of removing and re-creating the shims has more potential points of failure, each failure will only affect a single shim, as opposed to the entire directory. This means at worst (if we successfully delete a shim but then fail to re-create it), we lose information about a single shim, and the error should give information about which shim failed.

It won't be an easy thing to recover from, but it _will_ be easier than if we failed the operation in a way that lost all of the information about the shim directory at once.

charlespierce:
@dherman There was a lot of discussion on Discord about the concurrency problems and corruption, but are there any outstanding concerns about this PR? I'm happy to address if we want to make changes before merging, I'm just not 100% sure where we landed, as there were a lot of hard problems discussed :smile:

crates/volta-migrate/src/lib.rs
```diff
@@ -38,10 +43,33 @@ enum MigrationState {
...
+            None
+        }
+    }
+
+    fn detect_legacy_state(home: &VoltaHome) -> Fallible<Self> {
         /*
-        Triage for determining the current layout version:
-        - Does the V1 layout file exist? If yes then V1
-        - Does Volta Home exist? If yes (Windows) then V0
+        Triage for determining the lagacy layout version:
```
dherman: lagacy -> legacy

Issue #503, opened 2019-07-16
`volta pin ...` removes a newline from the end of package.json


bmakuh:
I noticed that when I run `volta pin yarn` or `volta pin node`, the newline at the end of my package.json is removed and I have to manually add it back in. It's not a big deal, but I figured I'd report it.

## Current Behavior
it removes a newline from the end of package.json

## Expected Behavior
it does not remove a newline from the end of package.json

chriskrycho:
Looks like this is a function of how serde_json's `PrettyPrinter` is implemented; see [here](https://github.com/serde-rs/json/issues/315). Our simplest solution is probably just checking if the original ends in a newline and adding it if so, given serde_json doesn't want to implement it (which seems reasonable, given that it's *effectively* a file-level parsing/writing issue here).

rwjblue:
>  Our simplest solution is probably just checking if the original ends in a newline and adding it

Ya, definitely agree here.

charlespierce:
Closed via #603

PR #603, opened 2019-11-15
Fix for #503

haochuan:
This is a attempt for fixing #503 

One of my concern is that, I'm only checking ONE `\n` at the end of `package.json`, and appending ONE `\n` if there is one after updating `package.json`. What if there are multiple `\n` at the bottom of the original `package.json`? Should we also take care of that case?

From another perspective, based my personal experience, no `\n` or one `\n` at the end of `package.json` seems fine, but people should not leave more than one `\n` there. From this I think just checking a single `\n` should be good.

Please advise thanks :)


crates/volta-core/src/manifest/mod.rs
```diff
@@ -96,13 +100,18 @@ impl Manifest {
...
                 .with_context(|_| ErrorDetails::PackageWriteError { file: get_file() })?;
             let formatter =
                 serde_json::ser::PrettyFormatter::with_indent(indent.indent().as_bytes());
-            let mut ser = serde_json::Serializer::with_formatter(file, formatter);
+            let mut ser = serde_json::Serializer::with_formatter(&file, formatter);
             map.serialize(&mut ser)
                 .with_context(|_| ErrorDetails::PackageWriteError { file: get_file() })?;
+            // append the empty line if the original package.json has one
+            if is_end_with_newline {
+                file.write_all("\n".as_bytes())
```
charlespierce: Instead of using `write_all`, this is a good spot to use the `write!` macro:

```
if is_end_with_newline {
    write!(file, "\n").with_context(...)?;
}
```

charlespierce:
Also, don't worry about the one failing test, it's a network failure (that annoyingly seems to be happening more and more lately), not anything wrong with your code.

crates/volta-core/src/manifest/mod.rs
haochuan: cool updated the PR with `write!`, thanks for the suggestion 👍

PR #604, opened 2019-11-16
Package Install Improvements

### LINKED ISSUE #480: `volta install <package>` should use latest LTS allowed by engines
Currently, running `volta install some-package` Volta will use `some-package`'s `engines` field to decide what version of Node to pin the tool too in future invocations. Conversely, `volta install node` will **automatically** use the most recent Node LTS (not `latest`).

I believe that we should change the logic such that we default to using the most recent LTS that is allowed within the engines configuration for the package. Specifically, it is **very** common (a best practice even) that packages do not limit the upper bounds of engines because most packages work just fine across many many Node versions. This would make `volta install some-package` have matching behavior to `volta install node`.

tldr; `engines` usage is most commonly `>= 8` or something like that, `volta install` choosing the `latest` (instead of latest LTS) means that its not uncommon (once every 6 months or so as new Node major versions are released) that folks would be installing with a Node that is fundamentally less tested/bulletproof.

### LINKED ISSUE #491: Support version tags as well as semver in `volta install`
Hi, thanks for making Volta! I've hit a minor issue, I'm trying to use volta to install [`reason-cli`](https://github.com/reasonml/reason-cli) tools globally. But in order to do so, I have to install a specific version for my platform (`latest-linux` in this case). This seems to run afoul of volta's version parsing:

```
error: Could not parse version "latest-linux"
```

Thanks!

charlespierce:
Closes #491 
Closes #480 

Info
-----
Our current `volta install <package>` process has a few slightly-related issues that have come up multiple times:
* We currently select the latest version of Node that matches the `engines` field (or the latest overall if there are no `engines`), which can result in broken CLI tools when there is a major version bump.
* We don't currently support tag versions other than `latest` (e.g. `beta`).

Changes
-----
* Reworked the `VersionSpec` field to support tags, with `latest` and `lts` special-cased since they require additional handling.
* Updated the package resolve logic to support resolving version tags as well as pure versions or version ranges.
* Added an internal-only tag type (that can't be created using the `FromStr` implementation): `LtsRequirement`, which represents the latest LTS version that matches a given SemVer requirement.
* Updated the node resolution algorithm to support the `LtsRequirement` tag, finding the latest LTS version that matches the requirement (or the latest version overall that matches, if no LTS versions match).

Tested
-----
* Ran `volta install elm@elm0.19.0` to confirm that the tag version is correctly resolved into an actual version and installed.
* Ran `volta install ember-cli` followed by `ember --version` to confirm that the latest LTS version of Node was selected, instead of the overall latest.
* Added a smoke test to cover installing a tag version

### DIFF
--- a/crates/volta-core/src/tool/node/resolve.rs
+++ b/crates/volta-core/src/tool/node/resolve.rs
@@ -15,7 +15,7 @@ use crate::layout::volta_home;
 use crate::session::Session;
 use crate::style::progress_spinner;
 use crate::tool::Node;
-use crate::version::VersionSpec;
+use crate::version::{VersionSpec, VersionTag};
 use cfg_if::cfg_if;
 use fs_utils::ensure_containing_dir_exists;
 use headers_011::Headers011;
@@ -42,10 +42,13 @@ cfg_if! {
 pub fn resolve(matching: VersionSpec, session: &mut Session) -> Fallible<Version> {
     let hooks = session.hooks()?.node();
     match matching {
-        VersionSpec::Latest => resolve_latest(hooks),
-        VersionSpec::Lts => resolve_lts(hooks),
         VersionSpec::Semver(requirement) => resolve_semver(requirement, hooks),
         VersionSpec::Exact(version) => Ok(version),
+        VersionSpec::None | VersionSpec::Tag(VersionTag::Lts) => resolve_lts(hooks),
+        VersionSpec::Tag(VersionTag::Latest) => resolve_latest(hooks),
+        VersionSpec::Tag(VersionTag::Custom(tag)) => {
+            Err(ErrorDetails::NodeVersionNotFound { matching: tag }.into())
+        }
     }
 }
 
--- a/crates/volta-core/src/tool/node/resolve.rs
+++ b/crates/volta-core/src/tool/node/resolve.rs
@@ -46,6 +46,7 @@ pub fn resolve(matching: VersionSpec, session: &mut Session) -> Fallible<Version
         VersionSpec::Exact(version) => Ok(version),
         VersionSpec::None | VersionSpec::Tag(VersionTag::Lts) => resolve_lts(hooks),
         VersionSpec::Tag(VersionTag::Latest) => resolve_latest(hooks),
+        VersionSpec::Tag(VersionTag::LtsRequirement(req)) => resolve_lts_semver(req, hooks),
         VersionSpec::Tag(VersionTag::Custom(tag)) => {
             Err(ErrorDetails::NodeVersionNotFound { matching: tag }.into())
         }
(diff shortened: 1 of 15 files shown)


rwjblue:
Awesome!! Thank you!

crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -42,10 +42,14 @@ cfg_if! {
...
-        VersionSpec::Latest => resolve_latest(hooks),
-        VersionSpec::Lts => resolve_lts(hooks),
         VersionSpec::Semver(requirement) => resolve_semver(requirement, hooks),
         VersionSpec::Exact(version) => Ok(version),
+        VersionSpec::None | VersionSpec::Tag(VersionTag::Lts) => resolve_lts(hooks),
+        VersionSpec::Tag(VersionTag::Latest) => resolve_latest(hooks),
+        VersionSpec::Tag(VersionTag::LtsRequirement(req)) => resolve_lts_semver(req, hooks),
+        VersionSpec::Tag(VersionTag::Custom(tag)) => {
+            Err(ErrorDetails::NodeVersionNotFound { matching: tag }.into())
+        }
```
chriskrycho: Can we add a comment here? It wasn't initially clear to me why this was always an error—I only figured it out when I realized this was Node-specific logic, and that Node doesn't have tagged releases in the same way as Node packages do.

crates/volta-core/src/version/mod.rs
```diff
@@ -1,84 +1,110 @@
...
+pub fn parse_requirements(s: impl AsRef<str>) -> Fallible<VersionReq> {
+    let s = s.as_ref();
+    serial::parse_requirements(s)
+        .with_context(|_| ErrorDetails::VersionParseError { version: s.into() })
+}
+
+pub fn parse_version(s: impl AsRef<str>) -> Fallible<Version> {
+    let s = s.as_ref();
+    s.parse()
+        .with_context(|_| ErrorDetails::VersionParseError { version: s.into() })
```
chriskrycho: Is this one of those weird cases where you need a temporary for lifetime stuff, such that this won't work?

```rust
s.as_ref()
    .parse()
    .with_context(|_| ErrorDetails::VersionParseError { version: s.into() })
```

crates/volta-core/src/tool/node/resolve.rs
charlespierce: Absolutely

crates/volta-core/src/version/mod.rs
charlespierce: This was actually trying to avoid the code duplication of calling `as_ref()` in both places, since we can't do `s.into()` directly and need to do `s.as_ref().into()`. I'm happy with it either way, it was just easier to work with a realized `&str` value than an `impl AsRef<str>` when it was used multiple times.

chriskrycho: Ahhhh, got it. I don't have a strong opinion here, was just curious!

lolmaus:
Thanks! 

Looking forward to user documentation on this feature.

Dmitry-N-Medvedev:
I might been mistaken, but

1. LTS does not give me the latest features (e.g. class instance private methods)
2. I personally do not care if the latest node's version is not yet battle tested.

charlespierce:
Hi @Dmitry-N-Medvedev, this is actually out-of-date now. As of Volta 0.9.0, released last week, we now pin each package to your default Node at the time it's installed. So if you want to use bleeding-edge features, you can set your default with `volta install node@latest`, and then any packages will use that latest version. If you then want to update in the future, you can do the same: Update your default Node with `volta install`, then re-install the packages against the latest and greatest Node.

Issue #576, opened 2019-10-17
Adding suffix to version in `volta -v` to signify if you're using a dev build


fusion2004:
What do y'all think about making the version command add a suffix like `-dev` to the outputted version if you're running a development build of volta?

E.g.:
```
$ volta -v
0.6.3-dev
```

Or is there maybe some other easy way to figure this out that I missed?

charlespierce:
@fusion2004 That's a great idea! There isn't currently an easy way to tell that you are using a dev build instead of the latest production build.

mikaoelitiana:
would checking the existance of `.git` be a way to determine we are in dev build?

charlespierce:
@mikaoelitiana I don't think that would work, because the production builds are also done in the same directory, so there will always be a `.git` folder around. However, it looks like Rust has a conditional compilation flag you can use to detect a debug build (as opposed to a `--release` build):
```rust
#[cfg(debug_assertions)]
```
So we should be able to use that to conditionally add the `-dev` suffix to the output in [`src/cli.rs`](https://github.com/volta-cli/volta/blob/master/src/cli.rs#L51)

charlespierce:
Closed via #607

PR #607, opened 2019-11-18
#576: Adding suffix "dev" for dev build

haochuan:
This is a PR for #576, not sure if we need a test for this :)


src/cli.rs
```diff
@@ -49,7 +49,11 @@ impl Volta {
     pub(crate) fn run(self, session: &mut Session) -> Fallible<ExitCode> {
         layout::ensure_volta_dirs_exist()?;
         if self.version {
-            println!("{}", env!("CARGO_PKG_VERSION"));
+            // suffix indicator for dev build
+            match cfg!(debug_assertions) {
+                true => println!("{}-dev", env!("CARGO_PKG_VERSION")),
+                false => println!("{}", env!("CARGO_PKG_VERSION")),
+            }
```
charlespierce: While not a hard error, this is a warning in Clippy, which suggests using an `if` statement instead of matching on `true` and `false`, for readability:
```
if cfg!(debug_assertions) {
    println!("{}-dev", env!("CARGO_PKG_VERSION"));
} else {
    println!("{}", env!("CARGO_PKG_VERSION"));
```

haochuan: Yep that makes sense, thanks for the feedback, already updated in this PR.

PR #618, opened 2019-12-03
[volta updates] Ensure layout file is written on new installation

### LINKED ISSUE #617: [Updates] Layout File not Written on new Installation
Testing Volta with the `volta-updates` flag enabled, the `layout.v1` file isn't being written for a new installation, thus resulting in it constantly being detected as a new installation on every invocation.

The root cause is a refactor done in #590, to move creation of the layout file into the `From<V0>` impl, that didn't get copied into the `From<Empty>` impl.

charlespierce:
Closes #617 

The layout file was being written on a migration from a legacy V0 layout, but not on creation of an entirely new layout. This was causing the migration to run on every invocation, detecting the system as being in a "new" state, and then not write the layout file.

Updated the logic so both types of migrations result in writing the layout file before finishing.

### DIFF
--- a/crates/volta-migrate/src/v1.rs
+++ b/crates/volta-migrate/src/v1.rs
@@ -26,6 +26,19 @@ impl V1 {
             home: v1::VoltaHome::new(home),
         }
     }
+
+    /// Write the layout file to mark migration to V1 as complete
+    ///
+    /// Should only be called once all other migration steps are finished, so that we don't
+    /// accidentally mark an incomplete migration as completed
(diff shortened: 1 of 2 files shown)


charlespierce:
Good call! I actually had testing of the empty migration, but somehow that _also_ ignored creation of the layout file on an empty migration 🤦‍♂

Updated to check that as well!

charlespierce:
The CI failure appears to be an unrelated network hiccup in the smoke tests (I've seen some unreliability in fetching release info from Github for old node versions, we might want to update the versions we use to test to be more recent so there's less of a chance of that)

PR #619, opened 2019-12-05
Enable updates by removing feature flag

### LINKED ISSUE #611: Remove "volta-updates" feature flag to make updates the default
Once the core updates work is complete (#590), we can remove the feature flag and make the flagged behavior the default. This will involve:

- [ ] Remove all code currently flagged `#[cfg(not(feature = "volta-updates"))]`

- [ ] Remove the now-unnecessary conditional flag `#[cfg(feature = "volta-updates")]`

- [ ] Remove the Feature from all `Cargo.toml` files

- [ ] Remove no-longer-true comment about the behavior of `volta_install()` (see comments of #590)

charlespierce:
Closes #611 

Info
-----
* Now that all of the incremental work on updates is complete, we can enable the feature and remove the feature flag.

Changes
-----
* Removed all blocks that were tagged with `#[cfg(not(feature = "volta-updates"))]`
* Removed the `#[cfg(feature = "volta-updates")]` attribute, now that the behavior is the default
* Removed unnecessary modules (where the `mod` statement was removed above)
* Removed features from `Cargo.toml` files and made all optional dependencies that depended on the feature now required dependencies.
* Removed a comment that is no longer accurate
* Made the custom "updates" installers the default, keeping around the legacy Unix installer for potential downgrade use-case.


crates/volta-core/src/session.rs
```diff
@@ -25,10 +25,6 @@ pub enum ActivityKind {
     Uninstall,
     List,
     Current,
-    #[cfg(not(feature = "volta-updates"))]
-    Deactivate,
-    #[cfg(not(feature = "volta-updates"))]
-    Activate,
```
mikrostew: 😂

wix/main.wxs
```diff
@@ -60,103 +60,79 @@
...
+                    Source='target\release\volta-shim.exe'
+                    KeyPath='yes'/>
+            </Component>
+            <Component Id='migrateBinary' Guid='*' Win64='$(var.Win64)'>
+                <File
+                    Id='voltamigrateEXE'
+                    Name='volta-migrate.exe'
+                    DiskId='1'
+                    Source='target\release\volta-migrate.exe'
+                    KeyPath='yes'/>
```
chriskrycho: How have we tested this? Just want to make sure it's solid before we ship it!

chriskrycho: (Note that it looks *right* to me, but I figured I'd double-check. Same goes for the \*n\*x installers.)

charlespierce: I tested the installers before this in a couple of beta releases, see: https://github.com/volta-cli/volta/releases/tag/v0.7.0-beta2

I definitely agree we need to make sure they're solid, I think we can cut a 0.7.0-beta3 after merging this and do some more testing (that will also allow me to do some internal tests on distribution).

charlespierce:
Most of the installer testing has been manual, so I'm definitely open to having a few users test it out. I'll cut a beta release and then ping the Discord channel to see about some additional user testing.

chriskrycho:
Seems good to me. Given I know they’re among our Windows base, cc. also @pjking07 @dustinsoftware

dustinsoftware:
Yeah! Would be happy to test a prerelease build.

charlespierce:
@dustinsoftware The pre-release build is available at https://github.com/volta-cli/volta/releases/tag/v0.7.0-beta3. Please let us know if you run into any issues (there shouldn't be much, if any, functional change from 0.6.8)

dustinsoftware:
Installed 0.70-beta3 on Mac and Windows, all my projects still build fine 👍

PR #625, opened 2019-12-14
[WIP] Support Sources for each individual Version

charlespierce:
As part of support for installing / pinning `npm`, we need to handle all combinations of the Versions in `PlatformSpec` coming from different sources. Currently we handle sources with a single enum that only supports 3 values: `Project`, `Default`, and `ProjectNodeDefaultYarn`. This was already unwieldy and adding support for new Package Managers will result in an exponential expansion in the variants of the enum.

Instead, we want each Version to have an associated `Source`, independent from the other versions, so that each tool can specify where it came from easily. This should also simplify the merging logic by removing the need to set a top-level "Source" value.


chriskrycho:
Ping me when you're ready for a review on this! As you know, it's a thing I've been interested in for a while!

charlespierce:
@chriskrycho I think this is ready for review now. It touches a lot of places, and I'm not _thrilled_ with the ergonomics of working with `SourcedVersion`, so I'd definitely be open to suggestions on improvements or better ways to model the relationship.

The core frustration I have is that when you are _constructing_ a `PlatformSpec` object, you pretty much always have a single source, so it's redundant to have to specify it multiple times (or specify it at all, depending on the use case). The only exception is the `merge` operation, which is already part of `impl PlatformSpec` and so sort of internal.

However, when you are _reading_ a `PlatformSpec`, you need the source to be associated with each tool, because the `merge` could cause them to come from different places.

It'd also be nice to have the type system enforce some of the Sources, as it is it's entirely possible for the `Project` object to have a `PlatformSpec` that specifies `Default` as the source, which seems like a missed opportunity.

One idea I have, that I might investigate, would be to create a `trait PlatformSpec` and then have different implementations of that trait for the various objects, e.g. `struct DefaultPlatformSpec`. That would allow you to construct it without needing to specify the source (just the object type), while still supporting more general sources. Then the places that just need a generic one can accept a trait Object (of some kind, still need to work out the specifics), while places where you want more type safety can use one of the concrete implementations.

chriskrycho:
Excellent. I will take a look in the morning and see if I come up with any alternatives while I’m at it!

crates/volta-core/src/toolchain/mod.rs
```diff
@@ -52,37 +52,38 @@ impl Toolchain {
...
-        if let Some(ref mut platform) = self.platform {
-            if platform.node_runtime != node_version.runtime {
-                platform.node_runtime = node_version.runtime.clone();
-                dirty = true;
-            }
-
-            if platform.npm.as_ref() != Some(&node_version.npm) {
-                platform.npm = Some(node_version.npm.clone());
+        if let Some(platform) = &self.platform {
+            if platform.node.version != *node_version {
```
chriskrycho: As a point of curiosity: any reason to prefer `platform.node.version != *node_version` over `&platform.node.version != node_version`?

crates/volta-core/src/toolchain/mod.rs
```diff
@@ -52,37 +52,38 @@ impl Toolchain {
...
+                self.platform = Some(Rc::new(PlatformSpec {
+                    node: SourcedVersion::default(node_version.clone()),
+                    npm: platform.npm.clone(),
+                    yarn: platform.yarn.clone(),
+                }));
                 dirty = true;
```
chriskrycho: Is dropping this handling intentional?

crates/volta-core/src/toolchain/mod.rs
```diff
@@ -95,39 +96,32 @@ impl Toolchain {
...
-                platform.yarn = Some(yarn_version.clone());
-                dirty = true;
+        if let Some(platform) = &self.platform {
+            if platform.yarn.as_ref().map(|y| &y.version) != Some(yarn_version) {
+                self.platform = Some(Rc::new(PlatformSpec {
+                    node: platform.node.clone(),
+                    npm: platform.npm.clone(),
+                    yarn: Some(SourcedVersion::default(yarn_version.clone())),
+                }));
+                self.save()?;
```
chriskrycho: Reading these implementations, I see what you mean about repetition. I wonder if the ergonomic improvement might be to have methods on `PlatformSpec` like this:

```rust
impl PlatformSpec {
    fn with_new_default_yarn(&self, yarn_version: &Version) -> Rc<PlatformSpec> {
        Rc::new(PlatformSpec {
            node: self.node.clone(),
            npm: self.npm.clone(),
            yarn: Some(SourcedVersion::default(yarn_version.clone()),
        })
    }
}
```

Then in each of these places, you'd use it like:

```rust
if platform.yarn.as_ref().map(|y| &y.version) != Some(yarn_version) {
    self.platform = self.platform.with_new_default_yarn(yarn_version);
}
```

If you wanted, you could even go so far as to do something like this:

```rust
impl PlatformSpec {
    fn with_updated_yarn(&self, yarn_version: &VersionSpec, source: Fn(Version) -> Source) {
        Rc::new(PlatformSpec {
            node: self.node.clone(),
            npm: self.npm.clone(),
            yarn: Some(source(yarn_version.clone()),
        })
    }
}
```

That in turn would let you just do:

```rust
if platform.yarn.as_ref().map(|y| &y.version) != Some(yarn_version) {
    self.platform = self.platform.with_updated_yarn(yarn_version, SourcedVersion::default);
}
```

etc.

crates/volta-core/src/toolchain/mod.rs
charlespierce: None really, other than that conceptually it makes more sense in my head to compare the values (by dereferencing the borrow) than the references. I know that both result in the same comparison, so if the borrow approach is more idiomatic Rust then I'm happy to swap.

crates/volta-core/src/toolchain/mod.rs
charlespierce: Yes, as we're now only passing the `Version`, instead of the `NodeVersion`, so we aren't passing the version of `npm`. Also, looking forward to the support of `volta install npm`, we don't want to write the `npm` version unless the user explicitly calls that command, otherwise we want to leave it alone.

This might be premature, however, so it may be better to pass both along and keep the behavior the same for now.

crates/volta-core/src/toolchain/mod.rs
charlespierce: I like that, and I was thinking that adding some convenience methods would help clean up a lot of the issues I felt when converting. I'll look at adding those in addition to the Trait idea I had and see if I can come up with something that feels great to use.

crates/volta-core/src/toolchain/mod.rs
chriskrycho: I don't think there's any particular preference one way or the other – I personally tend to reach for borrow over deref, but I think it's totally fine either way! Was just curious what *your* feeling was, mostly!

charlespierce:
After letting this sit for a while over the break, I ultimately think the trait approach that lets us get better guarantees from the type system in #626 is the better path forward.

PR #626, opened 2019-12-17
Support Sources for individual Tool Versions

charlespierce:
Info
-----
* To support installing / pinning `npm`, we need to handle Sources in a more granular way.
* Currently we have a single enum for the source of the `PlatformSpec` as a whole: `Project`, `Default`, or `ProjectNodeDefaultYarn`.
* As we expand to supporting `npm` and eventually other package managers, a single enum would scale exponentially, so we need a better approach.
* We want each `Version` to have an associated `Source`, independent from the other versions.
* Additionally, we want to make ensure as much as possible that the code is ergonomic and take advantage of the Type system wherever we can.

Changes
-----
* `PlatformSpec` is now a trait, which exposes methods that return the various sourced versions.
* Concrete structs now exist for `DefaultPlatformSpec`, `ProjectPlatformSpec`, `BinaryPlatformSpec`, and `MergedPlatformSpec`, each of which has simple creation semantics since the sources are known ahead of time.
* This makes creating a `PlatformSpec` more ergonomic, since you create the specific concrete type.
* It also allows for more type-safety, since we can ensure that `Project` only ever contains a `ProjectPlatformSpec`.
* When reading, you can now access the source for each version as needed.


crates/volta-core/src/session.rs
```diff
@@ -126,36 +124,31 @@ impl Session {
...
     }
 
     /// Returns the user's default platform, if any
-    pub fn default_platform(&self) -> Fallible<Option<Rc<PlatformSpec>>> {
+    pub fn default_platform(&self) -> Fallible<Option<Rc<DefaultPlatformSpec>>> {
         let toolchain = self.toolchain.get()?;
-        Ok(toolchain
-            .platform_ref()
-            .map(|platform| Rc::new(platform.clone())))
+        Ok(toolchain.platform())
```
chriskrycho: A triviality: this could just be:

```rust
self.toolchain.get().map(|toolchain| toolchain.platform())
```

crates/volta-core/src/toolchain/mod.rs
```diff
@@ -40,7 +38,7 @@ impl LazyToolchain {
 }
 
 pub struct Toolchain {
-    platform: Option<PlatformSpec>,
+    platform: Option<Rc<DefaultPlatformSpec>>,
```
chriskrycho: In this approach, why is the `Toolchain.platform` *always* `DefaultPlatformSpec`? Is there no scenario where the toolchain platform spec is the *project* value? (I'm trying to reload the `Toolchain` context into my head so this may be a stupid question!)

crates/volta-core/src/session.rs
charlespierce: Good call!

crates/volta-core/src/toolchain/mod.rs
charlespierce: It's not a stupid question, it's actually a consequence of poor naming :smile: The `Toolchain` object really represents the "Default Toolchain", it serializes and deserializes the `platform.json` that we write into the user's `~/.volta` directory. This is in contrast with the `Project`, which represents the information we pull from `package.json`.

So, for example, when a user does `volta install node`, we call `Toolchain::set_active_node`, while when a user does `volta pin node`, we call `Project::pin_node`

crates/volta-core/src/session.rs
charlespierce: Actually simplified even one more step:

```
self.toolchain.get().map(Toolchain::platform)
```

crates/volta-core/src/run/mod.rs
```diff
@@ -57,6 +59,17 @@ pub fn execute_tool(session: &mut Session) -> Fallible<ExitStatus> {
     command.status()
 }
 
+fn debug_tool_message<T>(tool: &str, version: Sourced<T>)
```
chriskrycho: A tiny suggestion that doesn't much matter, but might be a nice little win for maintainability: could we make this `version: &Sourced<T>`? I just plumbed it through and it all seems to work nicely – mostly doesn't matter where we're using it, but lets us drop one `.as_ref()` in favor of just a `&`.

crates/volta-core/src/session.rs
```diff
@@ -116,36 +114,30 @@ impl Session {
...
+                Ok(Some(platform))
             }
-        } else if let Some(platform) = self.default_platform()? {
-            Ok(Some(SourcedPlatformSpec::default(platform)))
-        } else {
-            Ok(None)
+            None => match self.default_platform()? {
+                Some(platform) => Ok(Some(platform)),
+                None => Ok(None),
+            },
```
chriskrycho: Huh. There's an interesting quirk here where the match triggers implicit behavior I would not expect—one of the bits of fallout from the `match` ergonomics improvements I 98% really appreciate. I would normally expect that this would be identical to just doing `None => self.default_platform()`, but no.

charlespierce: Yeah, I think I tried that as well. I believe it's a result of the fact that the Trait object is nested within other types. So it's not necessarily valid that `Fallible<Option<T>>` is assignable to `Fallible<Option<U>>` if `T: U`, but if you unwrap and rewrap it, it works out, since you are then explicit about how it's contained.

crates/volta-core/src/run/mod.rs
charlespierce: Good call :+1:

PR #638, opened 2020-01-10
Fix #636: Improve volta uninstall messaging for non-installed packages

### LINKED ISSUE #636: Improve volta uninstall messaging for non-installed packages
`0.7.0-beta3`
image

haochuan:
- Move the logic of showing success/warn into `package::uninstall`
- Show `uninstall success` when package config exists or orphaned binary exists
- Matched unpacked and initialized package won't trigger `uninstall success`

Let me know if there is anything missing or anything I need to change. Thanks!

### DIFF
--- a/crates/volta-core/src/tool/package/mod.rs
+++ b/crates/volta-core/src/tool/package/mod.rs
@@ -137,12 +137,10 @@ impl Display for Package {
 /// * the shims
 /// * the unpacked and initialized package
 pub fn uninstall(name: &str) -> Fallible<()> {
-    let mut package_found = false;
     let home = volta_home()?;
     // if the package config file exists, use that to remove any installed bins and shims
     let package_config_file = home.default_package_config_file(name);
-    if package_config_file.exists() {
-        package_found = true;
+    let package_found = if package_config_file.exists() {
         let package_config = PackageConfig::from_file(&package_config_file)?;
 
         for bin_name in package_config.bins {
@@ -151,16 +149,19 @@ pub fn uninstall(name: &str) -> Fallible<()> {
 
         fs::remove_file(&package_config_file)
             .with_context(delete_file_error(&package_config_file))?;
+        true
     } else {
         // there is no package config - check for orphaned binaries
         let package_binary_list = binaries_from_package(name)?;
         if !package_binary_list.is_empty() {
-            package_found = true;
             for bin_name in package_binary_list {
                 remove_config_and_shim(&bin_name, name)?;
             }
+            true
+        } else {
+            false
         }
-    }
+    };
 
     // if any unpacked and initialized packages exists, remove them
     let package_image_dir = home.package_image_root_dir().join(name);
--- a/crates/volta-core/src/tool/package/mod.rs
+++ b/crates/volta-core/src/tool/package/mod.rs
@@ -137,38 +137,43 @@ impl Display for Package {
 /// * the shims
 /// * the unpacked and initialized package
 pub fn uninstall(name: &str) -> Fallible<()> {
+    let mut package_found = false;
     let home = volta_home()?;
+    // if the package config file exists, use that to remove any installed bins and shims
     let package_config_file = home.default_package_config_file(name);
-    let package_binary_file = binaries_from_package(name)?;
+    if package_config_file.exists() {
+        package_found = true;
+        let package_config = PackageConfig::from_file(&package_config_file)?;
 
-    if package_config_file.exists() || !package_binary_file.is_empty() {
-        // if the package config file exists, use that to remove any installed bins and shims
-        if package_config_file.exists() {
-            let package_config = PackageConfig::from_file(&package_config_file)?;
+        for bin_name in package_config.bins {
+            remove_config_and_shim(&bin_name, name)?;
+        }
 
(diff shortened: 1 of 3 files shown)


crates/volta-core/src/tool/package/mod.rs
```diff
@@ -138,31 +138,37 @@ impl Display for Package {
...
+
+            info!("{} package '{}' uninstalled", success_prefix(), name);
+        } else {
+            // there is no package config - check for orphaned binaries
+            for bin_name in package_binary_file {
+                remove_config_and_shim(&bin_name, name)?;
+            }
         }
+    } else {
+        warn!("No package '{}' found to uninstall.", name);
```
chriskrycho: A suggestion for restructuring this a bit – since you end up repeating the checks inside the outer `if` expression, I'd just push this up to a single, three-scenario `if` expression:

```rust
if package_config_file.exists() {
    // if the package config file exists, use that to remove any installed bins and shims
    let package_config = PackageConfig::from_file(&package_config_file)?;

    for bin_name in package_config.bins {
        remove_config_and_shim(&bin_name, name)?;
    }

    fs::remove_file(&package_config_file)
        .with_context(delete_file_error(&package_config_file))?;

    info!("{} package '{}' uninstalled", success_prefix(), name);
} else if !package_binary_file.is_empty() {
    // there is no package config - check for orphaned binaries
    for bin_name in package_binary_file {
        remove_config_and_shim(&bin_name, name)?;
    }
} else {
    warn!("No package '{}' found to uninstall.", name);
}
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -138,31 +138,37 @@ impl Display for Package {
...
+
+            fs::remove_file(&package_config_file)
+                .with_context(delete_file_error(&package_config_file))?;
+
+            info!("{} package '{}' uninstalled", success_prefix(), name);
+        } else {
+            // there is no package config - check for orphaned binaries
+            for bin_name in package_binary_file {
+                remove_config_and_shim(&bin_name, name)?;
+            }
```
charlespierce: It appears we're no longer writing the "success: " message on this branch. I think it makes sense to still write it if we find _anything_ to uninstall, even if it was orphaned. Right now for this branch we don't write any sort of "done" message, just comments within the `remove_config_and_shim` method about individual binaries being uninstalled.

charlespierce:
One other note: You can run the tests locally, as noted in the [Contributing](https://docs.volta.sh/contributing/#tests) docs, by running:

```
cargo test --all --features mock-network
```

We could probably make that more prominent in the documentation.

crates/volta-core/src/tool/package/mod.rs
haochuan: Thanks for the feedback. One question is that if we go with `if .. else if... else` as you suggested, if the code goes into the `else if` then we are not printing anything to the users right?

haochuan: Since I'm not quite sure if `!package_binary_file.is_empty()` part should be considered as a valid uninstall or just a clean up.

chriskrycho: I *think* Chuck covered this in his comments below, but if it's still unclear after working through that, I'll follow up!

haochuan:
Thanks for all the great feedbacks!

Updates:

- It should show __success__ message if config file exists or orphaned binaries exist, otherwise show __warning__ message.
- Since `binaries_from_package` is expensive, I simply introduced the `package_found` flag to prevent calling it at the beginning to check what message to show.
- Updated test `uninstall_nonexistent_pkg` to check `stderr`.
- `cargo fmt --all`

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -163,6 +169,11 @@ pub fn uninstall(name: &str) -> Fallible<()> {
             .with_context(delete_dir_error(&package_image_dir))?;
     }
 
+    match package_found {
+        true => info!("{} package '{}' uninstalled", success_prefix(), name),
+        false => warn!("No package '{}' found to uninstall", name),
+    };
```
charlespierce: For `true/false`, matching looks a bit odd imo (I also think there might be a clippy lint warning around it). Instead we should try to use `if package_found {} else {}`, as it has the same effect and is slightly more readable.

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -137,10 +137,12 @@ impl Display for Package {
 /// * the shims
 /// * the unpacked and initialized package
 pub fn uninstall(name: &str) -> Fallible<()> {
+    let mut package_found = false;
```
charlespierce: I think we can actually avoid a mutable boolean value and instead assign the value of the if statement directly. That is, we can do something like:

```rust
let package_found = if package_config_file.exists() {
    ...
    true
} else {
    let package_binary_list = binaries_from_package(name)?;
    if !package_binary_list.is_empty() {
        ...
        true
    } else {
        false
    }
};
```

And then follow it up with the `if package_found` check below to determine what output to write.

haochuan: cool love that 👍

PR #644, opened 2020-01-21
Trim loader and store on all platforms

### LINKED ISSUE #643: Volta should automatically fix incorrect CRLF in shebang lines
Upon installing [git-mob](https://github.com/findmypast-oss/git-mob) with volta (`volta install git-mob`), I noticed that it didn't work. When I invoke any of the commands (e.g. `git mob`), I get the following error:

```
$ git mob
env: node\r: No such file or directory
```

The reason, it turns out, is that the package was published to NPM incorrectly on a Windows machine, and the shebang line ended up being `#! /usr/bin/env node\r`, so `"node\r"` is passed to `env` and it's complaining about that.

You can inspect the published package using `npm view git-mob latest dist.tarball` and see the issue for yourself (the files are inside `bin/*` of the tarball).

Since this effectively means this tool is broken for everyone on unix (which is, presumably, what that shebang line is for in the first place), I figured someone must have already reported the issue. To my surprise, everyone else seems to be running this tool just fine without issues. I tried following the instructions and install it via `npm install -g` without volta, and surprisingly, it worked.

When I inspected the files installed by `npm`,  I noticed that all the CRLF in the files are still present – except in the shebang line. This leads me to believe that this must be a common-enough problem that npm (and probably yarn) already corrects for (specifically, shebang lines that ends in CRLF in "bin" files). Because I am using volta to install the tool, it is also bypassing the correction that npm/yarn would have performed.

So, I think `volta install` should also do the same fix automatically, even though it is ultimately a mistake by the package maintainer.

charlespierce:
Closes #643 

Info
-----
* Packages are sometimes published from Windows with CRLF (`\r\n`) as the line-ending on the scripts.
* This causes issues with shebang loaders, since Unix interprets `\r` as part of the command, instead of part of the EOL marker.
* Volta already supports parsing and storing information about the loader to use when running the command, we can take advantage of that to remove unexpected whitespace before calling the appropriate program.

Changes
-----
* Moved the `determine_script_loader` function out of OS-specific code and set it up to run on all platforms.
* Added a call to `.trim()` to the loader to remove any extraneous `\r` (and other whitespace) characters from the arguments before storing them.

Tested
-----
* Confirmed that the package listed in the original issue (`git-mob@1.0.1`) works correctly with the new loader information.

### DIFF
--- a/crates/volta-core/src/tool/package/install.rs
+++ b/crates/volta-core/src/tool/package/install.rs
@@ -290,6 +290,8 @@ fn determine_script_loader(bin_name: &str, full_path: &Path) -> Fallible<Option<
         })?;
     if let Some(Ok(first_line)) = BufReader::new(script).lines().next() {
         if let Some(caps) = SHEBANG.captures(&first_line) {
+            // Note: `caps["args"]` will never panic, since "args" is a non-optional part of the match
+            // So if there is a Regex match, then it will necessarily include the "args" group.
             let args = caps["args"]
                 .trim()
                 .to_string()
--- a/crates/volta-core/src/tool/package/install.rs
+++ b/crates/volta-core/src/tool/package/install.rs
@@ -3,6 +3,8 @@
 
 use std::collections::HashMap;
 use std::ffi::OsStr;
+use std::fs::File;
+use std::io::{BufRead, BufReader};
 use std::path::Path;
 use std::process::Command;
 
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/tool/package/install.rs
```diff
@@ -266,55 +278,45 @@ fn read_bins(name: &str, version: &Version) -> Fallible<HashMap<String, String>>
...
+/// On Unix, we need to do this to remove any potential erroneous \r characters that may
+/// have accidentally been published in the script.
+fn determine_script_loader(bin_name: &str, full_path: &Path) -> Fallible<Option<BinLoader>> {
+    let script =
+        File::open(full_path).with_context(|_| ErrorDetails::DetermineBinaryLoaderError {
+            bin: bin_name.into(),
+        })?;
+    if let Some(Ok(first_line)) = BufReader::new(script).lines().next() {
+        if let Some(caps) = SHEBANG.captures(&first_line) {
+            let args = caps["args"]
```
chriskrycho: If I'm remembering/understanding the docs right, this will always succeed because we only get here if we have captures, and the regex is constructed such that it *cannot* succeed without the presence of `args` in the string being targeted, right? (Just wanted to make sure I understood the invariant which guarantees this won't ever panic!)

chriskrycho: Oh, and now I see it's been around for a while, too!

charlespierce: Right, the `Index` impl will only panic ["if there is no group named by the given value"](https://docs.rs/regex/1.3.3/regex/struct.Captures.html#panics-1). In this case, the named group is always part of the match (not optional or part of an "or" condition), so if there is a match, it will include the group.

I'll add a comment about that invariant, however, in case something changes in the future and we can't figure out why we did it that way :smiley:

PR #647, opened 2020-01-21
Support creating completions in new directories

### LINKED ISSUE #645: Wrong error message for `completions` when directory is missing
When a directory in the path supplied to `-o`/`--output` is missing, we report:

```
error: Completions file `/Users/ckrycho/.config/fish/completions/volta.fish` already exists.

Please remove the file or pass `-f` or `--force` to override.
Error details written to /Users/ckrycho/.volta/log/volta-error-2020-01-21_13_24_49.899.log
```

This is the wrong error; it should (a) report the correct error or – better, if possible – just create the directory.

chriskrycho:
Previously, we were failing when any part of the path did not exist, and reporting the wrong error to boot. Here, we attempt to create the directory instead, informing the user that we are changing their file system. Adds a new error scenario to cover a failure (e.g. a permissions error on creation), and a success message after writing the completion file.

Fixes #645

Screen Shot 2020-01-21 at 16 23 05

### DIFF
--- a/crates/volta-core/src/error/details.rs
+++ b/crates/volta-core/src/error/details.rs
@@ -68,6 +68,10 @@ pub enum ErrorDetails {
         path: PathBuf,
     },
 
+    CompletionsDirCreationError {
+        path: PathBuf,
+    },
+
     /// Thrown when the containing directory could not be determined
     ContainingDirError {
         path: PathBuf,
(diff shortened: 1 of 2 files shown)


crates/volta-core/src/error/details.rs
```diff
@@ -572,6 +576,11 @@ Use `npm install` or `yarn add` to select a version of {} for this project.",
 Please remove the file or pass `-f` or `--force` to override.",
                 path.display()
             ),
+            ErrorDetails::CompletionsDirCreationError { path } => write!(
+                f,
+                "Could not create directory {}.",
```
charlespierce: Given that the message doesn't need anything specific to completions, we can probably re-use `CreateDirError`, which has essentially the same error message along with a permissions CTA.

chriskrycho: Ah, perfect – I went looking for that but had a search-fu-failure!

chriskrycho: Fixed in 452394c!

PR #659, opened 2020-02-07
Add JSON support for `list` command

musikid:
This commit add JSON output for the `list` command. I'm not really sure if the structure of the output is really clear or need to be changed.

### DIFF
--- a/src/command/list/json.rs
+++ b/src/command/list/json.rs
@@ -1,0 +1,420 @@
+use serde_json::{to_string, to_string_pretty};
+
+use super::{Node, Package, PackageManager, Toolchain};
+
+pub(super) fn format(toolchain: &Toolchain) -> Option<String> {
+    let (runtimes, package_managers, packages) = match toolchain {
+        Toolchain::Node(runtimes) => (describe_runtimes(&runtimes), None, None),
+        Toolchain::PackageManagers(package_managers) => {
+            (None, describe_package_managers(&package_managers), None)
+        }
+        Toolchain::Packages(packages) => (None, None, describe_packages(&packages)),
+        Toolchain::Tool {
+            name,
+            host_packages,
+        } => (None, None, Some(describe_tool_set(name, host_packages))),
+        Toolchain::Active {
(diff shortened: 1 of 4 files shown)


charlespierce:
Thanks for this @musikid ! I’ll take a look later today or early tomorrow!

src/command/list/json.rs
```diff
@@ -0,0 +1,420 @@
...
+
+fn describe_runtimes(runtimes: &[Node]) -> Option<String> {
+    #[derive(serde::Serialize)]
+    struct Runtimes<'a> {
+        runtimes: &'a [Node],
+    };
+    if runtimes.is_empty() {
+        None
+    } else {
+        Some(to_string_pretty(&Runtimes { runtimes }).unwrap())
```
chriskrycho: One suggestion as I'm looking at this: I think the *default* value for JSON printing should be a compact representation, and we should add a `--pretty` flag when printing JSON that uses `to_string_pretty` instead of `to_string`. What do you think @MusiKid and @charlespierce?

MusiKid: I think that it's indeed a better idea to provide an option for the pretty output. How we should implement it however? We can create two different `format` function, one for the pretty output and another other for the compact one, or provide a new argument for the function like this:
```rust
pub(super) fn format(toolchain: &Toolchain, pretty: bool) -> Option<String>
```
What do you think?

chriskrycho: Sorry for the long delay in response! I'll try to get back to this in the next week or so!

MusiKid: OK no problem!

chriskrycho:
I apologize again! The last few months were really slammed for me. This is on my radar to circle back to next week. @MusiKid if you're still interested, would you mind rebasing this sometime between now and then? If not (and no worries; it's totally my fault here!), I'll probably see if I can pick this up and finish it myself.

MusiKid:
@chriskrycho Yes of course!

charlespierce:

PR #662, opened 2020-02-12
Regenerate shims on 'volta setup'

### LINKED ISSUE #661: Ensure Valid Shims on `volta setup`
In some situations, especially with managed installs, it's possible for the shim symlinks to become invalid by the binaries changing locations. In those situations, we should have a way to force regeneration of the shims to ensure that they are valid.

We already have a command, `volta setup`, which is designed to ensure that your system is correctly set up for Volta. That command should also regenerate the shims in addition to the modifications it makes to the profile scripts.

charlespierce:
Closes #661 

Info
-----
* Since the `volta setup` command is designed to make sure the user's system is correctly set up for Volta, it should also make sure to enforce that the shims are pointing to the correct location.
* Currently `volta` commands will only regenerate the shims if a migration is run.

Changes
-----
* Moved `regenerate_shims_for_dir` function into `volta_core::shim` module, instead of the `volta-migrate` module.
* Ensured that `regenerate_shims_for_dir` is always called after a migration runs, to maintain existing functionality.
* Added a call to `regenerate_shims_for_dir` into `volta setup` command, so that even if a migration doesn't run, `volta setup` will still cause the shims to be rebuilt.

Tested
-----
* Confirmed that running `volta setup` on an otherwise fully migrated system would correctly regenerate the shims and allow them to point at the correct binaries.

Notes
-----
* Regenerating the shims on `volta setup` does mean that if `volta setup` causes a migration, the shims will be regenerated twice, but it's a quick operation and not on the hot path, so it shouldn't be a major concern.

### DIFF
--- a/crates/volta-core/src/shim.rs
+++ b/crates/volta-core/src/shim.rs
@@ -1,11 +1,59 @@
 //! Provides utilities for modifying shims for 3rd-party executables
 
-use std::{fs, io};
+use std::collections::HashSet;
+use std::fs::{self, DirEntry, Metadata};
+use std::io;
+use std::path::Path;
 
 use crate::error::ErrorDetails;
-use crate::fs::symlink_file;
+use crate::fs::{read_dir_eager, symlink_file};
 use crate::layout::{volta_home, volta_install};
-use volta_fail::{throw, FailExt, Fallible};
+use log::debug;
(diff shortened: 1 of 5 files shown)


haochuan:
👍👍👍

PR #666, opened 2020-02-13
Always add Volta shims to PATH

### LINKED ISSUE #665: VS Code Terminal not using correct node version
I could possibly be misunderstanding this, but I ran the command `volta pin node@12.13.0`.

This added ```"volta": {
        "node": "12.13.0"
    }``` to my package.json

But when I open up a new terminal in vscode, and run `node --version` its 10.X.X. When using an actual Mac terminal, Volta correctly changes node to the right version

charlespierce:
Closes #665 

Info
-----
* Due to how VS Code launches a shell using the existing environment, shells launched with VS Code already have the Volta shim directory on the PATH, before the run any profile / startup scripts.
* As a result, we detect that Volta is on the PATH and don't update it, however it is so far back on the PATH that it is superseded by any existing `node` / `npm` / `yarn` installation already on the machine.
* This causes `node --version` to output the wrong thing when using the VS Code terminal, while using a normal terminal directly works as expected.

Changes
-----
* Updated the lines that we write to profile scripts to always append `$VOLTA_HOME/bin` to the PATH, instead of first checking if it already exists and then updating.

Tested
-----
* Tested both `zsh` and `fish` through VS Code, to confirm that with this change, the Volta shim directory is always prepended to the PATH, not left at the back by the idiosyncrasies of how VS Code launches a shell.

Note
-----
* If you receive Volta updates automatically through a managed install, you may need to run `volta setup` once to update your profile script. Alternatively, you can manually edit your scripts to _always_ prepend `$VOLTA_HOME/bin` to your `PATH`, instead of conditionally updating.

### DIFF
--- a/src/command/setup.rs
+++ b/src/command/setup.rs
@@ -119,7 +119,7 @@ mod os {
         let mut file = File::create(path)?;
         write!(
             file,
-            "{}\nexport VOLTA_HOME=\"{}\"\ngrep --silent \"$VOLTA_HOME/bin\" <<< $PATH || export PATH=\"$VOLTA_HOME/bin:$PATH\"\n",
+            "{}\nexport VOLTA_HOME=\"{}\"\nexport PATH=\"$VOLTA_HOME/bin:$PATH\"\n",
             contents,
             volta_home.display(),
         )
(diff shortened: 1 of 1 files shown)


chriskrycho:
One thing to check before we ship this – make sure it doesn't prepend more and more and more instances of this to the path in fish. I don't remember all the details (it's been a year!) but I do recall that I had to do something specific in the way I wrote the original bits there to make sure that worked correctly and you didn't end up with with `/users/chris/.volta/bin /users/chris/.volta/bin /users/chris/.volta/bin /usr/local/bin` etc. I *believe* that was independent of checking whether it was already present and was a function of the specific *way* we add it to the path, but should confirm!

PR #667, opened 2020-03-02
Prevent ~/.volta/bin from being added to PATH after uninstalling Volta

rstacruz:
They there - this is more of a proposal for improving the shell script :-)

I noticed that Volta adds some lines into ~/.config/fish/config.fish or the bash/zsh profile scripts. Given that Volta doesn't have an uninstaller, some may be tempted to just remove ~/.volta when uninstalling it. This can cause some invalid paths to be added to $PATH.

This change will suppress that if Volta isn't present.


charlespierce:

PR #670, opened 2020-03-07
Implement Migration to support custom npm versions

charlespierce:
Info
-----
* To support custom npm versions, there are a few changes that need to be migrated:
    * First, we should remove the `npm` version from the PATH to the Node image directory, since that Node will be used regardless of the npm version.
    * Second, we should make sure that if the default npm version is set in `platform.json` (our internal state), we remove it, since in the new paradigm a value being set represents a custom npm version, and until now we have only had bundled versions.
* Additionally, there was some friction when adding a new migration, so I created a macro to deduplicate some of the detection work.

Changes
-----
* Added a macro for creating methods to handle detecting the existing Volta layout version (checking each existing layout file to see if it exists.
* Added a new `V2` layout that has the `node_image_dir` depend only on the Node version, not the npm version.
* Added a migration from V1 to V2 of the layout, shifting all of the node images to their new location and correctly unsetting the `npm` value in `platform.json`, if necessary.
* Updated all of the code to only use Node version when locating the `node_image_dir`, instead of needing the `npm` version as well.

Tested
-----
* Manually tested that `platform.json` and Node images are updated as expected.
* Updated the existing unit tests to match the new paradigm and confirmed that all passed.
* Added acceptance tests for happy-path Migration as well as already-migrated cases.

Notes
-----
* I'm not thrilled with the migration itself, Migrations are messy and necessarily ad-hoc, so it's difficult to have a "clean" abstraction around the work, but I'm open to any changes that might make it easier to understand going forward.

### DIFF
--- a/crates/volta-migrate/src/lib.rs
+++ b/crates/volta-migrate/src/lib.rs
@@ -38,13 +38,15 @@ enum MigrationState {
 
 /// Macro to simplify the boilerplate associated with detecting a tagged state.
 ///
-/// Should be passed a series of tuples, each of which contains:
+/// Should be passed a series of tuples, each of which contains (in this order):
 ///
 /// * The layout version (module name from `volta_layout` crate, e.g. `v1`)
 /// * The `MigrationState` variant name (e.g. `V1`)
-/// * The migration object itself (e.g. `V1` from the v1 module)
+/// * The migration object itself (e.g. `V1` from the v1 module in _this_ crate)
 ///
-/// The tuples should be in reverse chronological order, so that the newest is first
+/// The tuples should be in reverse chronological order, so that the newest is first, e.g.:
+///
+/// detect_tagged!((v3, V3, V3), (v2, V2, V2), (v1, V1, V1));
 macro_rules! detect_tagged {
     ($(($layout:ident, $variant:ident, $migration:ident)),*) => {
         impl MigrationState {
--- a/crates/volta-migrate/src/lib.rs
+++ b/crates/volta-migrate/src/lib.rs
@@ -9,19 +9,21 @@
 //! able to re-start gracefully from an interrupted migration
 
 use std::convert::TryInto;
+use std::path::Path;
 
 mod empty;
 mod v0;
 mod v1;
+mod v2;
 
 use v0::V0;
 use v1::V1;
+use v2::V2;
 
 use volta_core::layout::volta_home;
 #[cfg(unix)]
 use volta_core::layout::volta_install;
 use volta_fail::Fallible;
-use volta_layout::v1::VoltaHome;
 
 /// Represents the state of the Volta directory at every point in the migration process
 ///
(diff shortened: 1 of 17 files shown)


rwjblue:
Seems logical to me, definitely interested in moving this forward...

crates/volta-migrate/src/v2.rs
```diff
@@ -0,0 +1,158 @@
...
+use tempfile::tempdir_in;
+use volta_core::error::ErrorDetails;
+use volta_core::fs::{ensure_dir_does_not_exist, read_dir_eager};
+use volta_core::tool::load_default_npm_version;
+use volta_core::toolchain::serial::Platform;
+use volta_core::version::parse_version;
+use volta_fail::{Fallible, ResultExt, VoltaError};
+use volta_layout::v2;
+
+/// Represents a V2 Volta Layout (from v0.7.3)
```
chriskrycho: Two notes:

1. Can you clarify (both in the PR and in this comment) what "from v0.7.3" means?
2. If it means "available from 0.7.3 forward", should it actually be v0.8.0? Technically the layout internally is private API, so I can see an argument either way, but it's a change that seems breaking-ish, in that anyone managing things manually has to run `volta setup` again to migrate?

crates/volta-migrate/src/v2.rs
```diff
@@ -0,0 +1,158 @@
...
+                })?;
+                ensure_dir_does_not_exist(&new_install)?;
+                rename(&temp_image, &new_install).with_context(|_| {
+                    ErrorDetails::SetupToolImageError {
+                        tool: "Node".to_string(),
+                        version: node_version.to_string(),
+                        dir: new_install.clone(),
+                    }
+                })?;
+            }
```
chriskrycho: There's enough going on here that I'd suggest extracting this into a standalone function. I think it'd be a lot clearer to read something like:

```rust
for node_version in node_installs {
  replace_default_npm_installation(&node_version, &npm_version)?;
}
```

—or something like that, anyway, not sure about the best name.

crates/volta-migrate/src/v2.rs
```diff
@@ -0,0 +1,158 @@
...
+                        node_version.npm = None;
+                        write(platform_file, existing_platform.into_json()?).with_context(
+                            |_| ErrorDetails::WritePlatformError {
+                                file: platform_file.to_owned(),
+                            },
+                        )?;
+                    }
+                }
+            }
+        }
```
chriskrycho: Similar to my comment below, I wonder if it makes sense to extract this as its own function. I tend to instinctively aim to keep my functions shorter—not quite to the extent of [Sandi Metz's "5-line-methods" rule](https://thoughtbot.com/blog/sandi-metz-rules-for-developers#five-lines-per-method), but _inspired_ by that at a minimum.

crates/volta-migrate/src/lib.rs
```diff
@@ -31,33 +33,60 @@ enum MigrationState {
...
 
+/// Macro to simplify the boilerplate associated with detecting a tagged state.
+///
+/// Should be passed a series of tuples, each of which contains:
+///
+/// * The layout version (module name from `volta_layout` crate, e.g. `v1`)
+/// * The `MigrationState` variant name (e.g. `V1`)
+/// * The migration object itself (e.g. `V1` from the v1 module)
+///
+/// The tuples should be in reverse chronological order, so that the newest is first
```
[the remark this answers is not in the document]
charlespierce: Definitely, good call-out!

crates/volta-migrate/src/v2.rs
charlespierce: Yes, it's intended to mean "available from 0.7.3 forward", I'll update the comment to be more clear about that.

To the other question, the migrations are checked on any run of `volta` or a shim, so there isn't a need to run specifically `volta setup` again. Once the user has the new binaries, the migration will run the next time they interact with Volta.

We also don't support "managing things manually" when it comes to the data directory, things have to be laid out exactly how Volta expects them or it won't be able to find what it's looking for. The migrations allow us to make those changes without requiring a "breaking" change the way we used to (where we required the user to completely delete their installation and re-install), instead we make any updates in-place.

crates/volta-migrate/src/v2.rs
charlespierce: Good thinking, that'll definitely make it more grokkable.

crates/volta-migrate/src/v2.rs
charlespierce: Agreed, will do. Thanks!

charlespierce:
Updated to split the migration itself into separate functions and improve the comments in a few places.

crates/volta-migrate/src/v2.rs
chriskrycho: Thanks for clarifying!

PR #687, opened 2020-03-25
volta-install.sh: make curl write out downloaded file name

### LINKED ISSUE #686: Install script failure: installer tried to extract archive with bad name
Trying to install `volta` today, I run into this:

```console
$ curl https://get.volta.sh | bash -s -- --skip-setup
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 11925    0 11925    0     0  15406      0 --:--:-- --:--:-- --:--:-- 15406
  Installing latest version of Volta (0.7.2)
    Checking for existing Volta installation
    Fetching archive for macOS, version 0.7.2
######################################################################## 100.0%#=#=-#  #  ######################################################################## 100.0%
    Creating directory layout
  Extracting Volta binaries and launchers
tar: Error opening archive: Failed to open '
/var/folders/5f/7dv67jv51d55gzm_723cp1d80000gn/T/tmp.lXjtBvvw/volta-0.7.2-macos.tar.gz'
```

```console
$  file /var/folders/5f/7dv67jv51d55gzm_723cp1d80000gn/T/tmp.lXjtBvvw/volta-0.7.2-macos.tar.gz
/var/folders/5f/7dv67jv51d55gzm_723cp1d80000gn/T/tmp.lXjtBvvw/volta-0.7.2-macos.tar.gz: gzip compressed data, last modified: Thu Feb 13 01:20:12 2020, from Unix, original size modulo 2^32 16739840
```

Taking a closer look, notice that the archive exists but the installer script tried to decompress `\n/var/.../volta...tar.gz`, the newline character should not be there.

---

For what it's worth, I downloaded and patched the installer and installed successfully, PR will follow.

rhcarvalho:
Depending on environment configuration, `download_release_from_repo`
could write to stdout both through `curl` and `echo`.

The install script becomes more robust when we explicitly tell `curl` to
write out the effective name where the download was saved, and nothing
else.

Fixes #686.

### DIFF
--- a/dev/unix/volta-install.sh
+++ b/dev/unix/volta-install.sh
@@ -23,7 +23,7 @@ download_release_from_repo() {
   local download_file="$tmpdir/$filename"
   local archive_url="$(release_url)/download/v$version/$filename"
 
-  curl --progress-bar --show-error --location --fail "$archive_url" --output "$download_file" && echo "$download_file"
+  curl --progress-bar --show-error --location --fail "$archive_url" --output "$download_file" --write-out '%{filename_effective}'
 }
 
 usage() {


dev/unix/volta-install.sh
```diff
@@ -23,7 +23,7 @@ download_release_from_repo() {
   local download_file="$tmpdir/$filename"
   local archive_url="$(release_url)/download/v$version/$filename"
 
-  curl --progress-bar --show-error --location --fail "$archive_url" --output "$download_file" && echo "$download_file"
+  curl --progress-bar --show-error --location --fail "$archive_url" --output "$download_file" --write-out '%{filename_effective}'
```
rhcarvalho, line 26: Alternatively, we can call `curl --disable ...` to disable reading the default config file. The downside is that users may have good reasons to change settings like proxy, etc.

charlespierce: Agreed, I don't think we want to disable the user's config, so using the `--write-out` parameter makes sense.

PR #691, opened 2020-03-27
[npm] [Large] Support pinning and installing 'npm'

charlespierce:
Closes #95 
Closes #292 

Info
-----
* We want to support users selecting a custom version of `npm`, separate from the version installed with Node.
* The functionality and UX are described / discussed in https://github.com/volta-cli/rfcs/pull/40

Changes
-----
* Added autodownload of custom npm version in `PlatformSpec::checkout`
* Implemented `Tool` methods for resolving, fetching, installing, and pinning `npm`
* Added info message on `volta install node` and `volta pin node` if the bundled `npm` version is higher than the pinned / default version, including a CTA to use `npm@bundled` to switch.
* Added support for `volta install npm@bundled` and `volta pin npm@bundled` to switch back to the version bundled with Node.

Tested
-----
* Manual testing 
* Added new acceptance and smoke tests to cover installing / pinning npm behavior
* Added new acceptance tests to cover the `npm@bundled` behavior.

PR #693, opened 2020-03-30
Use dynamic dispatch for tool objects

charlespierce:
Info
-----
* Currently we have an enum `tool::Resolved`, whose only purpose is to store a `Tool` implementation and dynamically dispatch to the contained `Tool` when one of the methods is called (`fetch`, `install`, or `pin`).
* This is effectively the same as dynamic dispatch with trait objects, so we can implement that and save on some boilerplate.
* Additionally, the switch to trait objects makes sense in the context of #691, which introduces a new `Tool` implementation: `BundledNpm`

Changes
-----
* Removed `Resolved` enum and it's implementation.
* Updated `Spec::resolve` to return a `Box<dyn Tool>` instead of a `Resolved`
* Updated `Tool` trait to use `self: Box<Self>` as the self-type for all methods.
* Fixed the implementations to all take `self: Box<Self>`

Tested
-----
* All tests pass (no functionality change expected in this refactor)

Note
-----
* This is a small PR to help simplify #691

### DIFF
--- a/crates/volta-core/src/tool/mod.rs
+++ b/crates/volta-core/src/tool/mod.rs
@@ -5,7 +5,6 @@ use crate::session::Session;
 use crate::style::{note_prefix, success_prefix, tool_version};
 use crate::version::{parse_version, VersionSpec};
 use log::{debug, info};
-use semver::Version;
 use volta_fail::Fallible;
 
 mod node;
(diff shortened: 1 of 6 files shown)


crates/volta-core/src/tool/mod.rs
```diff
@@ -53,11 +52,11 @@ fn info_project_version<T: Display + Sized>(tool: T) {
 /// Trait representing all of the actions that can be taken with a tool
 pub trait Tool: Display {
     /// Fetch a Tool into the local inventory
-    fn fetch(self, session: &mut Session) -> Fallible<()>;
+    fn fetch(self: Box<Self>, session: &mut Session) -> Fallible<()>;
```
chriskrycho: Question for my own edification: does Clippy not suggest `Box<dyn Self>` here? It seems vaguely redundant but I’m not positive on the implications off the top of my head (or even whether that’s syntactically valid!).

charlespierce: It doesn't, and a quick test shows that apparently `Box<dyn Self>` is a syntax error because `Self` is a type, not a trait (or something like that).

I actually got to `Box<Self>` by first doing `Box<dyn Tool>`, which was itself an error that suggested using `Box<Self>` 😄

PR #694, opened 2020-03-31
[npm] [Medium] Add support for querying and list 'npm' inventory

charlespierce:
Info
-----
* As the first step of adding support for `volta install npm` and `volta pin npm`, we need to make sure that we can manage and query the inventory for `npm`.
* In line with querying the inventory, `volta list` should be updated to include custom `npm` versions as well.

Changes
-----
* Added `npm_available` and `npm_versions` function to `inventory`, analogous with the already-existing `node` and `yarn` functions.
* Updated `volta list` behavior to query and interpret fetched `npm` versions.

Tested
-----
* Added new tests to `plain.rs` that cover outputting `npm`
* All tests pass

### DIFF
--- a/src/command/list/human.rs
+++ b/src/command/list/human.rs
@@ -1,12 +1,11 @@
 //! Define the "human" format style for list commands.
 
-use textwrap::{HyphenSplitter, Wrapper};
+use std::collections::BTreeMap;
 
-use volta_core::style::{text_width, tool_version, MAX_WIDTH};
-
-use super::{Node, Package, PackageManager, Toolchain};
-
+use super::{Node, Package, PackageManager, PackageManagerKind, Toolchain};
 use lazy_static::lazy_static;
+use textwrap::{HyphenSplitter, Wrapper};
+use volta_core::style::{text_width, tool_version, MAX_WIDTH};
 
 static INDENTATION: &str = "    ";
 static NO_RUNTIME: &str = "⚡️ No Node runtimes installed!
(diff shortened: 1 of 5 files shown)


charlespierce:
Note: With the landing of the human formatter for `volta list` (#697), this doesn't quite work, as the human formatter also needs to be updated (in progress).

mikrostew:
This looks good so far - I will hold off on approval until the human formatter changes are added here

charlespierce:
Updated to support the human formatter, bumped the effort estimate up to medium because there's a fair few changes in that module (also a whole bunch of new tests).

src/command/list/human.rs
```diff
@@ -661,17 +898,51 @@ See options for more detailed reports by running `volta list --help`.";
         use crate::command::list::{PackageManager, PackageManagerKind, Source};
 
         #[test]
-        fn none_installed() {
+        fn none_installed_npm() {
+            let expected = "⚡️ No npm versions installed.
```
chriskrycho, line 902: Should we change this message? It's not really true. Maybe we want something like "No custom npm versions installed" – it's just slightly strange to say *none* are given that you can totally type `npm <whatever>` and it uses the version from Node.

Note that changing it will definitely mean treating npm distinctly from Yarn—they won't be able to just share their whole implementation directly—but I think that's *probably* the right thing to do.

charlespierce: That's a good call-out, I think we should change it to be more explicit about the difference between custom and bundled npm versions.

charlespierce:
Updated to special-case `npm` Package Managers (with a call-out about `npm` still being available through Node).

src/command/list/human.rs
```diff
@@ -125,32 +125,57 @@ fn display_node(runtimes: &[Node]) -> String {
...
-        )
+        format!("⚡️ Custom npm versions in your toolchain:\n\n{}", versions)
+    }
+}
+
+/// Format a set of `Toolchain::PackageManager`s.
+fn display_package_managers(kind: PackageManagerKind, managers: &[PackageManager]) -> String {
+    match kind {
+        PackageManagerKind::Npm => display_npms(managers),
+        _ => {
```
chriskrycho, line 152: 👏 to making this easily forward-compatible for adding support for e.g. pnpm.

Issue #524, opened 2019-08-14
volta list should list packages which have been fetched but not installed


chriskrycho:
As a user introspecting my toolchain via `volta list`, I'd like to know what all versions of a package I have downloaded, including those which are cached but not unpacked or installed. (Follow-on work from #461.)

charlespierce:
With the rework of package installations to use `npm i -g` internally, this is no longer a possible case - We don't have the ability to fetch but not install a package.

PR #697, opened 2020-03-31
Implement `volta list` for human

haochuan:
This PR is the implementation of @chriskrycho 's RFC https://github.com/volta-cli/rfcs/blob/info-commands/text/0000-informational-commands.md

Since the basic code structure and tests has been completed, I tried my best to follow the original design as much as possible. However, I noticed a few places that I have to change:

### volta list <package>

From the RFC, command `volta list package` should print out the header message as `<package> versions in toolchain:`. However, since we don't have a way to get the package name from CLI argument, I'm just showing a generic message here: `⚡️ Package versions in your toolchain:`.

---

### volta list all

From the RFC, when listing packages in `list all`, the packages should be grouped by package name and then version, like:

```
Packages:
        <package name>
            [<version> (default | current @ <project path>)
                binaries: [<binary name>]...
                platform:
                    runtime: node@<version>
                    package manager: built-in npm|<npm|yarn>@<version>]
            [<version> (fetched)]
```

In this PR I'm not doing the grouping and simply just print out every `Toolchain::Package`, since we won't get more than one default `Package` so that the list won't bee too long, also having another group with indentations looks wired somehow in my screen.

On the other hand, I also like the way to group different version by package name, which makes the output more clear. I'm here asking opinions about this, if we still want to keep the grouping, I'm looking for suggestions about some good ways to do the mapping implementation.

---

### Package availability

As far as I understand, `volta list <package>` should give me all the version installed for that package. For example, if I have ember-cli@3.12.0 and ember-cli@3.13.0, I will have this via `volta list ember-cli`:

```
ember-cli@3.13.0 (default)
        binary tools: ember
        platform:
            runtime: node@12.2.0
            package manager: npm@built-in
    ember-cli@3.12.0
        binary tools: ember
        platform:
            runtime: node@12.2.0
            package manager: npm@built-in";
```

However, I noticed that for a specific `User` package, I always only get the current default one, not sure if this is by design, or if I missed something. I tried using `volta list <package> --format=plain` and seems have the same result.

---

### Others

I changed some minor text/formatting in the codebase, especially in the test section.

---

This PR definitely needs some cleanup work, since I'm fairly new to the codebase, I think it's a good idea to post the draft first, then we can start more discussion here.

Also I'm still in the process of learn Rust, if there is anything wrong please let me know. Appreciate!

### DIFF
--- a/src/command/list/human.rs
+++ b/src/command/list/human.rs
@@ -57,9 +57,8 @@ fn display_active(
     match (runtime, package_manager) {
         (None, _) => NO_RUNTIME.to_string(),
         (Some(runtime), Some(package_manager)) => {
-            let runtime_version: String =
-                WRAPPER.fill(&format!("Node: {}", format_runtime(runtime)));
-            let package_manager_version: String = WRAPPER.fill(&format!(
+            let runtime_version = WRAPPER.fill(&format!("Node: {}", format_runtime(runtime)));
+            let package_manager_version = WRAPPER.fill(&format!(
                 "Yarn: {}",
                 format_package_manager(package_manager)
             ));
@@ -88,7 +87,7 @@ fn display_active(
             } else {
                 WRAPPER.fill(&format!(
                     "Tool binaries available:\n{}",
-                    display_packages(packages)
+                    format_tool_list(packages)
                 ))
             };
 
@@ -144,8 +143,9 @@ fn display_node(runtimes: &[Node]) -> String {
 /// Format a set of `Toolchain::PackageManager`s.
 fn display_package_managers(package_managers: &[PackageManager]) -> String {
     if package_managers.is_empty() {
+        //TODO: adding npm support https://github.com/volta-cli/volta/pull/694
         String::from(
-            "⚡️ No <npm|Yarn> versions installed.
+            "⚡️ No Yarn versions installed.
 
 You can install a Yarn version by running `volta install yarn`.
 See `volta help install` for details and more options.",
@@ -229,7 +229,7 @@ fn format_runtime_list(runtimes: &[Node]) -> String {
     WRAPPER.fill(
         &runtimes
             .iter()
-            .map(|runtime| format_runtime(&runtime))
+            .map(format_runtime)
             .collect::<Vec<String>>()
             .join("\n"),
     )
@@ -662,7 +662,7 @@ See options for more detailed reports by running `volta list --help`.";
 
         #[test]
         fn none_installed() {
-            let expected = "⚡️ No <npm|Yarn> versions installed.
+            let expected = "⚡️ No Yarn versions installed.
 
 You can install a Yarn version by running `volta install yarn`.
 See `volta help install` for details and more options.";
--- a/src/command/list/human.rs
+++ b/src/command/list/human.rs
@@ -2,7 +2,7 @@
 
 use textwrap::Wrapper;
 
-use volta_core::style::text_width;
+use volta_core::style::{text_width, tool_version};
 
 use super::{Node, Package, PackageManager, Source, Toolchain};
 
@@ -16,7 +16,10 @@ pub(super) fn format(toolchain: &Toolchain) -> Option<String> {
     // Formatting here depends on the toolchain: we do different degrees of
     // indentation
     Some(match toolchain {
-        Toolchain::Node(runtimes) => display_node(&runtimes),
+        Toolchain::Node(runtimes) => format!(
+            "⚡️ Node runtimes in your toolchain:\n\n{}",
+            display_runtimes(&runtimes)
+        ),
         Toolchain::Active {
             runtime,
             package_manager,
@@ -28,11 +31,11 @@ pub(super) fn format(toolchain: &Toolchain) -> Option<String> {
             packages,
         } => display_all(runtimes, package_managers, packages),
         Toolchain::PackageManagers(package_managers) => display_package_managers(package_managers),
-        Toolchain::Packages(packages) => display_packages(packages),
+        Toolchain::Packages(packages) => display_packages(packages, true),
         Toolchain::Tool {
             name,
             host_packages,
-        } => display_tool(name, host_packages),
+        } => display_tools(name, host_packages),
     })
 }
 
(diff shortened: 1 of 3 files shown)


chriskrycho:
Very first things first: thank you so much for picking this up and running with it! I'll have plenty of comments, but that warrants getting out there first!

charlespierce:
To echo what @chriskrycho said, thanks for taking this on! I also want to call out your great description, making it very clear what questions you have and where we can focus our review efforts!

To answer your question about "Package availability", that isn't exactly _by design_, but it is currently the case. We already have an issue #524 that covers reading additional packages and getting information about them, so that's an outstanding issue and not something you need to fix in this PR. I think that also informs your decision to not group Packages by version, since there is only ever a single version for each package (currently). I think that the formatting of `human` can be re-evaluated when we implement a fix for that issue.

Could you add a comment to that issue, reminding us when we get to it to re-evaluate the formatting of `volta list all`?

chriskrycho:
> From the RFC, command `volta list package` should print out the header message as `<package> versions in toolchain:`. However, since we don't have a way to get the package name from CLI argument, I'm just showing a generic message here: `⚡️ Package versions in your toolchain:`.

I *think* (and I haven't read the code yet) that this is just a misunderstanding, likely because of a lack of clarity in the original RFC. I didn't intend `volta list package` to be a thing, only `volta list <some specific package>`, e.g. `volta list typescript`. But I think you actually *did* cover this—you ask about a related design issue in the **Package availability** section! I'll add explicit comments on that path in the code when I review it (hopefully tomorrow).

> ### volta list all
> From the RFC, when listing packages in `list all`, the packages should be grouped by package name and then version, like:
> 
> ```
> Packages:
>         <package name>
>             [<version> (default | current @ <project path>)
>                 binaries: [<binary name>]...
>                 platform:
>                     runtime: node@<version>
>                     package manager: built-in npm|<npm|yarn>@<version>]
>             [<version> (fetched)]
> ```
> 
> In this PR I'm not doing the grouping and simply just print out every `Toolchain::Package`, since we won't get more than one default `Package` so that the list won't bee too long, also having another group with indentations looks wired somehow in my screen.
> 
> On the other hand, I also like the way to group different version by package name, which makes the output more clear. I'm here asking opinions about this, if we still want to keep the grouping, I'm looking for suggestions about some good ways to do the mapping implementation.

I grouped it per the design in `list all --format=plain`, I believe, and it's definitely an important part of the design to maintain. As you note, the output is *much* clearer that way. I’ll point you to the relevant implementation bits and walk you through it as I’m able—hopefully tomorrow!

> ### Package availability
> As far as I understand, `volta list <package>` should give me all the version installed for that package. For example, if I have ember-cli@3.12.0 and ember-cli@3.13.0, I will have this via `volta list ember-cli`:
> 
> ```
> ember-cli@3.13.0 (default)
>         binary tools: ember
>         platform:
>             runtime: node@12.2.0
>             package manager: npm@built-in
>     ember-cli@3.12.0
>         binary tools: ember
>         platform:
>             runtime: node@12.2.0
>             package manager: npm@built-in";
> ```
> 
> However, I noticed that for a specific `User` package, I always only get the current default one, not sure if this is by design, or if I missed something. I tried using `volta list <package> --format=plain` and seems have the same result.

This is likely just a gap in the current implementation; I think I forgot to handle this case when building the version fo the inventory list we use for this! We can make a note to add that capability in a follow-on: just getting to parity with the `--format=plain` is probably the right move here.

> I changed some minor text/formatting in the codebase, especially in the test section.

I haven't looked at the details yet, but I’ll note that in *general* it’s preferable to keep those kinds of changes to separate PRs, just to keep the size of the PR more manageable for review! (I’m definitely guilty of this myself a lot of the time, though, so I sympathize!)

> This PR definitely needs some cleanup work, since I'm fairly new to the codebase, I think it's a good idea to post the draft first, then we can start more discussion here.
> 
> Also I'm still in the process of learn Rust, if there is anything wrong please let me know. Appreciate!

We’ll all be very happy to help here!

src/command/list/human.rs
```diff
@@ -41,14 +43,68 @@ pub(super) fn format(toolchain: &Toolchain) -> Option<String> {
...
+                    .fill(&format!("Tool binaries available: NONE"))
+            } else {
+                Wrapper::new(width)
+                    .initial_indent(INDENTATION)
+                    .subsequent_indent(INDENTATION)
+                    .fill(&format!(
+                        "Tool binaries available:\n{}",
+                        display_packages(packages)
+                    ))
+            };
```
charlespierce, line 99: `display_packages` looks like it already has internal logic for detecting an empty package list, is it possible to leverage that instead of duplicating this block?

src/command/list/human.rs
```diff
@@ -58,46 +114,272 @@ fn display_all(
...
+/// Format a sets binaries inside `Toolchain::Package`s without detail information
+fn display_tool_sets(packages: &[Package]) -> String {
+    format!(
+        "{}",
+        &packages
+            .iter()
+            .map(display_tool_set)
+            .collect::<Vec<String>>()
+            .join("\n")
+    )
```
charlespierce, line 234: Since `join("\n")` returns a `String`, I believe this is a redundant format.

src/command/list/human.rs
```diff
@@ -58,46 +114,272 @@ fn display_all(
...
+}
+/// Format a set binaries inside a single `Toolchain::Package` without detail information
+fn display_tool_set(package: &Package) -> String {
+    match package {
+        Package::Default { tools, .. } | Package::Project { tools, .. } => {
+            let tools = match tools.len() {
+                0 => String::from(""),
+                _ => format!("{}", tools.join(", ")),
```
charlespierce, line 242: Another spot where I believe we can use `tools.join(", ")` directly, instead of using the format

src/command/list/human.rs
```diff
@@ -58,46 +114,272 @@ fn display_all(
...
+                .subsequent_indent(INDENTATION)
+                .fill(&format!("{}{}", tools, list_package_source(package)))
+        }
+        Package::Fetched(..) => String::new(),
+    }
+}
+
+/// List a set versions of `Toolchain::Node`s.
+fn list_runtimes(runtimes: &[Node]) -> String {
+    if runtimes.is_empty() {
```
charlespierce, line 257: Very minor, but there seem to be a lot of potentially redundant checks for empty lists. For example, `display_node` makes a check for an empty slice, and then if it isn't empty it calls this function, which checks again. If possible, can we map out the flow in each of the various branches of output and figure out where they are needed and where we can skip the check?

src/command/list/human.rs
```diff
@@ -58,46 +114,272 @@ fn display_all(
...
+                    .collect::<Vec<String>>()
+                    .join("\n"),
+            )
+    }
+}
+
+/// List a the version of `Toolchain::PackageManager`.
+fn list_package_manager(package_manager: &PackageManager) -> String {
+    format!("v{}{}", package_manager.version, package_manager.source)
+}
```
charlespierce, line 301: This is a personal style, but I generally try to avoid having functions whose names are only off by one character (`list_package_manager` and `list_package_managers`). Also, `list` generally implies that there are more than one, could we make this name more descriptive and unique? Maybe something like `format_single_package_manager`?

src/command/list/human.rs
```diff
@@ -58,46 +114,272 @@ fn display_all(
...
+            tools,
+            ..
+        } => {
+            let tools = match tools.len() {
+                0 => String::from(""),
+                _ => format!("{}", tools.join(", ")),
+            };
+            let width = text_width().unwrap_or(0);
+
+            // println!("{}\n{}\n{}\n{}", details.name, details.version, node, tools);
```
charlespierce, line 343: Looks like a debugging comment :smile:

src/command/list/human.rs
haochuan: Actually this one seems quite tricky to me 😄 So use `display_packages` as an example:

- `display_packages` will be called to display both `Toolchain::Packages` and `Toolchain::Active`.

- For `Toolchain::Packages`, if it's empty, then it should show:

```
⚡️ No tools or packages installed.

You can safely install packages by running `volta install <package name>`.
See `volta help install` for details and more options.
```

- For `Toolchain::Active`, if it's empty, then it should show:

```
Tool binaries available: NONE
```

That's the main reason why I have the double empty check, to handle these two different logics, but I totally agree that we should make the code more clear. I will take another look to see how I could change it.

haochuan:
TODO: 

1. better way to handle empty checking
2. rename function names to avoid names like `list_package_manager` and `list_package_managers` (multi vs single)

haochuan:
Updated the PR with the following changes:

- Make `textwrap::Wrapper` into `lazy_static`
- Clean a bunch of redundant `format` usage
- Rename a few functions, now I'm using the word `list` to indicate single vs a list of singles. e.g: `format_package_manager` vs `format_package_manager_list`. Thanks @charlespierce calling out this and I really like this way
- Remove unnecessary `is_empty`, now we only have the checks in top level functions (function names start with `display`)
- Fix some lint issues and a format issue in `all` -> `full` test.

Local test:

- `cargo test --all --features mock-network`
- `cargo clippy`
- `./dev/unix/volta-install.sh --dev` and verify results locally

src/command/list/human.rs
```diff
@@ -1,16 +1,25 @@
...
-use volta_core::style::text_width;
+use volta_core::style::{text_width, tool_version, MAX_WIDTH};
 
-use super::{Node, Package, PackageManager, Source, Toolchain};
+use super::{Node, Package, PackageManager, Toolchain};
+
+use lazy_static::lazy_static;
 
 static INDENTATION: &'static str = "    ";
 static NO_RUNTIME: &'static str = "⚡️ No Node runtimes installed!
```
charlespierce, line 12: Very minor nit, but I believe these types can be written without the `'static`, since static vars have that lifetime automatically (clippy gives a warning about this IIRC).

```suggestion
static INDENTATION: &str = "    ";
static NO_RUNTIME: &str = "⚡️ No Node runtimes installed!
```

src/command/list/human.rs
```diff
@@ -41,14 +50,55 @@ pub(super) fn format(toolchain: &Toolchain) -> Option<String> {
...
-    match (node, package_manager, packages) {
-        (None, _, _) => NO_RUNTIME.to_string(),
-        (Some(node), Some(package_manager), packages) => unimplemented!(),
-        _ => unimplemented!(),
+    match (runtime, package_manager) {
+        (None, _) => NO_RUNTIME.to_string(),
+        (Some(runtime), Some(package_manager)) => {
+            let runtime_version: String =
+                WRAPPER.fill(&format!("Node: {}", format_runtime(runtime)));
+            let package_manager_version: String = WRAPPER.fill(&format!(
```
charlespierce, line 62: Minor nit: `Wrapper::fill` returns a `String`, so the type annotations aren't necessary.

src/command/list/human.rs
```diff
@@ -41,14 +50,55 @@ pub(super) fn format(toolchain: &Toolchain) -> Option<String> {
...
+        }
+        (Some(runtime), None) => {
+            let runtime_version: String =
+                WRAPPER.fill(&format!("Node: {}", format_runtime(runtime)));
+            let package_versions = if packages.is_empty() {
+                WRAPPER.fill(&format!("Tool binaries available: NONE"))
+            } else {
+                WRAPPER.fill(&format!(
+                    "Tool binaries available:\n{}",
+                    display_packages(packages)
```
charlespierce, line 91: This block uses `display_packages` to show the package list, but the block above uses `format_tool_list` in the exact same place. Is there a need for these two to be different?

src/command/list/human.rs
```diff
@@ -58,46 +108,215 @@ fn display_all(
...
+        Package::Fetched(..) => String::new(),
+    }
+}
+
+/// format a list of `Toolchain::Node`s.
+fn format_runtime_list(runtimes: &[Node]) -> String {
+    WRAPPER.fill(
+        &runtimes
+            .iter()
+            .map(|runtime| format_runtime(&runtime))
```
charlespierce, line 232: Another _extremely_ minor nit: Can this work without the parameter? That is, can this be written as: `.map(format_runtime)` in the same way that `list_packages` does?

src/command/list/human.rs
```diff
@@ -58,46 +108,215 @@ fn display_all(
...
     }
 }
 
-/// Format the output for `Toolchain::PackageManager`.
+/// Format a set of `Toolchain::PackageManager`s.
 fn display_package_managers(package_managers: &[PackageManager]) -> String {
-    unimplemented!()
+    if package_managers.is_empty() {
+        String::from(
+            "⚡️ No <npm|Yarn> versions installed.
```
charlespierce, line 148: Since the rest of this method is purely `Yarn` focused, can we update this to only say `Yarn`? Alternatively, we could make it fully language-agnostic, but it looks like that's a little bit more work (that can be tackled after this is merged in an update or follow-on to #694)

PR #703, opened 2020-04-02
[Small] Ensure error details are written to the terminal in CI environments

### LINKED ISSUE #677: Emit error information to console on CI systems.
Currently, `volta` will print out something like the following when an error occurs:

```
error: Could not download Yarn version registry
from https://api.github.com/repos/yarnpkg/yarn/releases

Please verify your internet connection.
Error details written to /Users/runner/hostedtoolcache/volta/0.7.2/x64/log/volta-error-2020-03-11_00_29_48.091.log
```

When run on an actual users machine, outputing the more detailed error information to a file like this is **very helpful** (avoids cluttering the console output too much). Unfortunately, when ran in a more ethereal environment these log files are impossible to actually access.

I've also attempted to run with `RUST_STACKTRACE=full` and `VOLTA_LOGLEVEL=debug`, but they seem to not have an effect.

---

My suggestion is to use the [ci_info](https://crates.io/crates/ci_info) crate to decide if we should emit to the console. The crate is based on the [ci-info](https://www.npmjs.com/package/ci-info) npm package, which is very popular in the Node ecosystem. For context, ember-cli uses `ci-info` (and the general purpose `$CI` environment variable as a fallback) for **exactly** the same purpose (deciding if we should emit additional output to the console, or to a separate file).

charlespierce:
Closes #677 

Info
-----
* In CI environments, writing the error details into a log file isn't helpful, as the file is often deleted along with everything else when the run is complete.
* Instead, we should write error details directly to the console when in CI, so that they are available when reviewing the console logs later.

Changes
-----
* Updated `error::reporter.rs` to print any additional error details out to the console in CI environments.

Tested
-----
* Updated error reporting tests to mock the CI environment as needed.
* Added tests to confirm the error information is written out in CI (with mocks, so they work locally as well).
* All tests pass.

### DIFF
--- a/tests/acceptance/verbose_errors.rs
+++ b/tests/acceptance/verbose_errors.rs
@@ -1,4 +1,5 @@
 use crate::support::sandbox::sandbox;
+use ci_info::types::{CiInfo, Vendor};
 use hamcrest2::assert_that;
 use hamcrest2::prelude::*;
 use test_support::matchers::execs;
@@ -17,6 +18,9 @@ const NODE_VERSION_INFO: &str = r#"[
 fn no_cause_shown_if_no_verbose_flag() {
     let s = sandbox().node_available_versions(NODE_VERSION_INFO).build();
 
+    // Mock `is_ci` to false so that this works even when running in Volta's CI Test Suite
+    ci_info::mock_ci(&CiInfo::new());
+
     assert_that!(
         s.volta("install node@10"),
         execs()
@@ -28,6 +32,9 @@ fn no_cause_shown_if_no_verbose_flag() {
 #[test]
 fn cause_shown_if_verbose_flag() {
     let s = sandbox().node_available_versions(NODE_VERSION_INFO).build();
+
+    // Mock `is_ci` to false so that this correctly tests the verbose flag
+    ci_info::mock_ci(&CiInfo::new());
 
     assert_that!(
         s.volta("install node@10 --verbose"),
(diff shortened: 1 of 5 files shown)


tests/acceptance/verbose_errors.rs
```diff
@@ -17,6 +18,9 @@ const NODE_VERSION_INFO: &str = r#"[
 fn no_cause_shown_if_no_verbose_flag() {
     let s = sandbox().node_available_versions(NODE_VERSION_INFO).build();
 
+    // Mock `is_ci` to false so that this works even when running in Volta's CI Test Suite
+    ci_info::mock_ci(&CiInfo::new());
```
chriskrycho, line 22: 🤯

tests/acceptance/verbose_errors.rs
```diff
@@ -80,3 +90,47 @@ fn no_error_log_if_no_underlying_cause() {
...
+#[test]
+fn cause_shown_in_ci() {
+    let s = sandbox()
+        .node_available_versions(NODE_VERSION_INFO)
+        .env("VOLTA_LOGLEVEL", "error")
+        .build();
+
+    // Mock a CI environment so this works even when running locally
+    let mut ci_mock = CiInfo::new();
+    ci_mock.vendor = Some(Vendor::GitHubActions);
```
chriskrycho, line 103: 🤯 🤯 

Okay, I really have to go read up on `ci_info`; this seems *awesome*.

PR #706, opened 2020-04-08
Decouple args from ToolCommand constructors

charlespierce:
Info
-----
* As part of investigating how to implement `volta run`, it became clear that the `ToolCommand` constructors were doing too much work.
* By setting all of the args on the constructor, we needed to know all of the args at a single time, which meant that the `args` had to be passed to each Tool's `run` method, even though the only thing they were used for was to then pass along to the `ToolCommand`.
* We also already had wrapper behavior for the `current_dir` method, so it made sense to add `args` behavior in a similar way.

Changes
-----
* Updated `ToolCommand` to support `arg` and `args` methods (passing them to the eventual command)
* Updated `execute_tool` to pass the args directly to the command after it is constructed.
* Updated `binary::command` to pass its args before returning, which results in them being first in the argument list (as required).
* Updated `node`, `npm`, `npx`, and `yarn` modules to no longer need the `args` parameter.

Tested
-----
* This shouldn't result in any changes to the logic, just a change in code organization and structure.
* All existing behavior tests pass.

### DIFF
--- a/crates/volta-core/src/run/mod.rs
+++ b/crates/volta-core/src/run/mod.rs
@@ -81,7 +81,7 @@ impl ToolCommand {
     /// Build a ToolCommand that is directly calling a tool in the Volta directory
     fn direct(exe: &OsStr, path_var: &OsStr) -> Self {
         ToolCommand {
-            command: command_in(exe, path_var),
+            command: command_with_path(exe, path_var),
             on_failure: ErrorDetails::BinaryExecError,
         }
     }
@@ -89,7 +89,7 @@ impl ToolCommand {
     /// Build a ToolCommand that is calling a binary in the current project's `node_modules/bin`
     fn project_local(exe: &OsStr, path_var: &OsStr) -> Self {
         ToolCommand {
-            command: command_in(exe, path_var),
+            command: command_with_path(exe, path_var),
             on_failure: ErrorDetails::ProjectLocalBinaryExecError {
                 command: exe.to_string_lossy().to_string(),
             },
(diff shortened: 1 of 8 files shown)


crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -164,12 +163,10 @@ fn npm_pack_command_for(
...
-    let args = vec![
-        OsString::from("pack"),
-        OsString::from("--no-update-notifier"),
-        OsString::from(format!("{}@{}", name, version)),
-    ];
-    let mut command = run::npm::command(args, session)?;
+    let mut command = run::npm::command(session)?;
+    command.arg("pack");
+    command.arg("--no-update-notifier");
+    command.arg(format!("{}@{}", name, version));
```
chriskrycho, line 169: Out of curiosity, could you have left it as the previous `Vec` and just done `command.args(args)`? (I think this is equally fine, probably comparable—and equally negligible!—performance-wise, genuinely just curious.)

crates/volta-core/src/run/mod.rs
```diff
@@ -164,17 +164,27 @@ fn tool_name_from_file_name(file_name: &OsStr) -> OsString {
     }
 }
 
-fn command_for<A>(exe: &OsStr, args: A, path_var: &OsStr) -> Command
-where
-    A: IntoIterator<Item = OsString>,
-{
+fn command_in(exe: &OsStr, path_var: &OsStr) -> Command {
```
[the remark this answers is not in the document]
charlespierce: Good call on both counts, will do!

crates/volta-core/src/run/npm.rs
```diff
@@ -53,15 +50,12 @@ fn check_npm_install() -> CommandArg {
...
-        || command == Some(OsString::from("isntall"))
-        || command == Some(OsString::from("add"))
-    {
-        // `args` here picks up from where the command lookup left off, so
-        // will be the name of the package passed to the command.
-        CommandArg::GlobalAdd(args.next())
-    } else {
-        CommandArg::NotGlobalAdd
+    match command {
+        Some(cmd) if cmd == "install" || cmd == "i" || cmd == "isntall" || cmd == "add" => {
```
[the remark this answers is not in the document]
charlespierce: Yep, @dherman discovered as we were initially implementing the interception the code in `npm` that lists out the different variants that they support.

crates/volta-core/src/tool/package/fetch.rs
charlespierce: We could have left it as a `vec`, but because of how the previous `args` was typed, there were a bunch of ultimately unnecessary `OsString::from` calls. I originally wanted to switch it to a `Vec<&str>`, but the formatted argument breaks that, so I went with the individual arg approach.

chriskrycho: Makes sense, that's roughly what I guessed. 👍

charlespierce:
Updated the name and added several doc comments that were missing.

PR #713, opened 2020-04-16
[run] [Large] Implement `volta run` command

charlespierce:
Info
----
* As described in https://github.com/volta-cli/rfcs/pull/39, we want to have a `volta run` command that allows you to override the detect `node` and `yarn` versions when running a specific command.

Changes
-----
* Added `CommandLine` source for Platforms and an `InheritOption` type that represents one of 3 possibilities: A Value, No Value but allowed to inherit, and No Value but _not_ allowed to inherit.
* Added a `CliPlatform` type that represents the data we can potentially get from the CLI (including the ability to prevent inheriting from the contextual Platform).
* Created the `run` command that parses the options and passes them to the core logic to execute.

Tested
-----
* Manual testing of `volta run` in different scenarios.
* Added unit tests around new Platform behaviors (merging and `InheritOption`)
* Added acceptance tests to cover the behavior of each tool under different scenarios with `volta run`
* Added smoke tests to cover the core behaviors of `volta run`
* All tests (new and existing) pass.

### DIFF
--- a/src/command/run.rs
+++ b/src/command/run.rs
@@ -69,18 +69,18 @@ impl Command for Run {
 /// We ignore any setting that doesn't have a value associated with it
 /// We also ignore the PATH environment variable as that is set when running a command
 fn parse_envs(cli_values: Vec<String>) -> impl IntoIterator<Item = (String, String)> {
-    cli_values.into_iter().filter_map(|mut entry| {
-        entry.find('=').and_then(|index| {
-            // After `split_off`, entry will contain only the key
-            let value = entry.split_off(index);
+    cli_values.into_iter().filter_map(|entry| {
+        let mut key_value = entry.splitn(2, '=');
 
-            if entry.eq_ignore_ascii_case("PATH") {
+        match (key_value.next(), key_value.next()) {
+            (None, _) => None,
+            (Some(_), None) => None,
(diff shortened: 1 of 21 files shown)


src/command/run.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// We also ignore the PATH environment variable as that is set when running a command
+    fn parse_envs(&self) -> impl IntoIterator<Item = (&str, &str)> {
+        self.envs.iter().filter_map(|entry| {
+            let mut key_value = entry.splitn(2, '=');
+
+            match (key_value.next(), key_value.next()) {
+                (None, _) => None,
+                (Some(_), None) => None,
+                (Some(key), _) if key.eq_ignore_ascii_case("PATH") => {
+                    debug!("Skipping PATH environment variable as it will be overwritten to execute the command");
```
chriskrycho, line 118:
```suggestion
                    warn!("Skipping PATH environment variable as it will be overwritten to execute the command");
```

src/command/run.rs
```diff
@@ -0,0 +1,125 @@
+use std::ffi::OsString;
+
+use crate::command::Command;
+use crate::common::{Error, IntoResult};
+use log::debug;
```
chriskrycho, line 5:
```suggestion
use log::warn;
```

PR #714, opened 2020-04-17
Add GitHub Actions CI Workflow

charlespierce:
Info
-----
* Investigating moving CI over to GitHub actions, as opposed to Azure Pipelines
* GH Actions are more tightly integrated into the GH experience, and recent stability issues with Azure (jobs timing out) have made it worthwhile to investigate an alternative

Changes
-----
* Added GH Workflow file for CI
* Added GH Workflow file to build & publish API docs on pushes to `master`

Tested
-----
* Added a line temporarily to publish API docs on PR, to ensure that flow works as expected.
* This PR, the actions are triggered and run successfully.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+jobs:
+  tests:
+    strategy:
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+    name: Build and Test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
```
rwjblue, line 20: I find this a bit nicer in my workflows (I already know that it is `-latest`, no need to make the test title that much longer).

```suggestion
          - ubuntu
          - macos
          - windows
    name: Build and Test (${{ matrix.os }})
    runs-on: ${{ matrix.os }}-latest
```

charlespierce: Oh that is nicer, thanks!

PR #716, opened 2020-04-20
Automate creation of releases with GitHub Actions

### LINKED ISSUE #676: Publish `.zip` file for Windows on release.
While working on a GitHub Action for Volta ([rwjblue/setup-volta](https://github.com/rwjblue/setup-volta)) it would be really helpful if we published *both* the `.msi` installer (because this is really what _most_ Windows users would want anyways) **and** a `.zip` in the same format as the other platform `.tar.gz` files (which follows the file layout mentioned in the manifest that we also publish).

I'm not sure how much work it is to add this additional publishing target, and I am attempting to work around (by unpacking from the `.msi` directly with `msiexec`). Please feel free to close if it is too much effort.

charlespierce:
Closes #676 

Info
-----
* To streamline and normalize the release process, we should build all the artifacts in CI, instead of on an individual maintainer's machines.
* We can use GitHub Actions to trigger these builds whenever a tag is pushed to the repository, allowing us to still have control over when releases are created.
* GitHub Actions can also create a new release and upload the artifacts so that it is ready for publishing.

Changes
-----
* Added `release.yml` with jobs to build all the necessary binaries and create a new Release in the repository.
* Added Dockerfile and supporting scripts to `ci` directory for Linux builds.

Notes
-----
* The release process will be triggered whenever we push a tag with the name `v<version>` to the repo.
* MacOS and Windows build are straightforward, using only `cargo` commands.
* Linux is where things get complicated, as we have a couple of constraints:
    * To ensure we don't have overly-restrictive `glibc` requirements, we need to build on an older distro.
    * We also need to support 3 different OpenSSL versions: 1.0.*, 1.1.*, and the custom-bundled version used by CentOS / RHEL
* To solve all these constraints, we take the following approach:
    * All Linux builds are performed on a Docker image of CentOS 6 (Previously, builds were manually created on a Linux machine that had RHEL 7, so this is slightly older and thus less restrictive).
    * For CentOS OpenSSL, we use the version that is included in the image
    * For OpenSSL 1.0.* and 1.1.*, we download the `1_0_1-stable` and `1_1_0-stable` branches of OpenSSL and build from source, then compile against those versions.
* The release is created as a draft, with no description. This allows us to insert the release notes / changelog (without having to do complicated text parsing) as well as spot-check to make sure everything is ready to be published.
* For Windows, in addition to the MSI installer, also create a zip file of the binaries.

Tested
-----
* For testing, I had this workflow set to trigger on PRs as well, which is why I force-pushed so many times.
* I was able to create a draft release from the PR branch, using a mocked version number.
* After merging, we will likely need to do a sanity check by pushing a tag to the repo and confirming we see a release created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,9 @@ on:
   push:
     tags:
       - v*
+  pull_request:
+    branches:
+      - master
 
 name: Build and Release
 
@@ -115,6 +118,7 @@ jobs:
       - linux
       - macos
       - windows
+    if: github.event_name == 'push' && startsWith(github.ref, 'v')
     steps:
       - name: Check out code
         uses: actions/checkout@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,216 @@
+on:
+  push:
+    tags:
+      - v*
+
+name: Build and Release
+
+jobs:
+  centos:
+    name: Build - CentOS / RHEL
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Set up docker image
+        run: docker build -t volta .
+        working-directory: ./ci/docker
+      - name: Compile and package Volta
+        run: docker run --volume ${PWD}:/root/workspace --workdir /root/workspace --rm --init --tty volta /root/workspace/ci/build-and-package.sh volta-centos
+      - name: Confirm correct OpenSSL Version
+        run: |
+          objdump -p target/release/volta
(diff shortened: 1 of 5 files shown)


charlespierce:
@rwjblue That's a good point, I'll update to do the building / uploading on PRs as well. As well as getting the potential "release" artifacts, that also does a sanity check that the compile will work in the release environment, so it makes sense.

charlespierce:
@rwjblue Updated to also create a zip file of the Volta binaries on Windows, per #676

rwjblue:
Looks really good: https://github.com/volta-cli/volta/actions/runs/85196457

PR #732, opened 2020-05-06
Use `hyperx` crate for typed headers

charlespierce:
Info
-----
* We are currently relying on the `hyper-011` feature in `reqwest@0.9` to support strongly-typed headers.
    * This limit our ability to upgrade, because `reqwest@0.10` doesn't include that feature any more.
* Additionally, we are using a bespoke `headers-011` crate to make it easier to work with those typed headers when sending requests.
* Both of those features are implemented in a single crate: `hyperx`, which extracts the typed header implementation from hyper itself.

Changes
-----
* Swapped from `headers-011` to `hyperx`, using an existing crate to bring in both the definitions of the strongly-typed headers and a Trait for working with `HeaderMap` types.
* Updated the uses of `get_011` to instead use `TypedHeaders::decode`
* Also removed some unnecessary dependencies from the root project


joshuarli:
itshappening

Issue #536, opened 2019-08-22
Investigate build bloat


dherman:
A friend told me about this [recent article on Rust build sizes](https://raphlinus.github.io/rust/2019/08/21/rust-bloat.html), and specifically the article links to a useful tool: [cargo-bloat](https://github.com/RazrFalcon/cargo-bloat).

We should use that tool to see if there are places we can reduce the binary size of Volta builds (and possibly the build times as a consequence).

One particular hint I got was that [reqwest](https://github.com/seanmonstar/reqwest) is an especially large library, and there are competitors like [isahc](https://docs.rs/isahc/0.7.0/isahc/) that are considerably smaller. We'd have to investigate to see if it looks like a good tradeoff, though (in terms of feature set, quality, level of maintenance).

And generally let's keep our wits about us—smaller binary size is good, but not if it comes at too high of a cost. It's not the only goal.

joshuarli:
@dherman I briefly looked into this after building volta for myself and getting kind of mad at how long it took to compile everything. In particular, `parking_lot` has quite a large dependency tree, and it can be completely omitted from the build by:

- upgrading console and indicatif to 0.8.0 and 0.12.0 respectively, which removes their hard dependency on parking_lot
- upgrading reqwest to 0.10.0, which [upgrades hyper to 0.13](https://github.com/seanmonstar/reqwest/blob/master/CHANGELOG.md#v0100), which [upgrades tokio to 0.2.x](https://github.com/hyperium/hyper/blob/master/CHANGELOG.md#v0130-2019-12-10), which [makes parking_lot an optional internal dependency](https://github.com/tokio-rs/tokio/blob/ecc23c084a64188e272e2c1e9e370189358a088b/tokio/CHANGELOG.md#added-1)

The latter is the hard part because the hyper-011 feature in reqwest is removed in 0.10.0, so everything in https://github.com/volta-cli/volta/pull/267 needs to be ported to new reqwests and it isn't AFAICT as trivial as a revert.

I doubt doing this will help binary size but it'll at least noticeably reduce the build time. It would also be nice to finally update these libraries.

joshuarli:
There's also https://github.com/algesten/ureq if volta's willing to trade async http(s) for a significan[t boost in simplicity and correctness](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6), I'd wager it would be worth it.

charlespierce:
[...]

PR #733, opened 2020-05-06
Use attohttpc instead of reqwest for fetching

charlespierce:
Info
-----
* As originally found by @joshuarli in #536, we don't need all the async features of `reqwest` in Volta, since all of our HTTP access is synchronous.
* Unfortunately, while `reqwest` provides a synchronous API, it still uses the async engine and task executor (`tokio`) under the hood.
* Compiling that bloats our build times and executable size, by bringing in an entire async executor when we don't actually use it.
* `attohttpc` is a purely-synchronous HTTP Request client with a very similar API to `reqwest`.

Changes
-----
* Removed `reqwest` and added `attohttpc` to all of the appropriate crates.
* Upgraded `hyperx` to `1.0.0`, to match the transitive dependency on `http`
* Updated the code to fit `attohttpc`s slightly different API

Notes
-----
* This results in a ~25% reduction in the executable size for `volta`, and a ~30% reduction in the executable size for `volta-shim`.
* Anecdotally, this reduced the time for a clean `cargo build --release` by ~ 22% (from 6m12s down to 4m50s).
* For performance, again anecdotally, I see no noticeable difference for fetching a version of Node, though I do get a statistically significant speedup on the hot path (likely a result of the smaller binary).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,12 +38,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "archive"
 version = "0.1.0"
 dependencies = [
+ "attohttpc 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "cfg-if 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "flate2 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "fs-utils 0.1.0",
- "hyperx 0.15.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


joshuarli:
Oh neat, I didn't know at all about `hyperx` back when I made a shitty attempt at porting old header stuff to new header stuff to upgrade reqwest. Seems like it made things a lot easier.

charlespierce:
Yeah, definitely, the `hyperx` stuff made it a _lot_ easier, not having to re-implement the header parsing. I didn’t know about the crate either until a day or two ago :smile:

Issue #711, opened 2020-04-16
Volta list implementation is wrong for locally-available tools


chriskrycho:
In a project with a given tool installed locally (in this case, `ember-cli`)—

- **Expected:**

    ```sh
    $ volta list

    ⚡️ Currently active tools:

        Node: v12.16.2 (current @ /Users/chris/dev/test/foo/package.json)
        Yarn: v1.22.4 (current @ /Users/chris/dev/test/foo/package.json)
        Tool binaries available:
            ember (current @ /Users/chris/dev/test/foo/package.json)
    ```

- **Actual:**

    ```sh
    $ volta list

    ⚡️ Currently active tools:

        Node: v12.16.2 (current @ /Users/chris/dev/test/foo/package.json)
        Yarn: v1.22.4 (current @ /Users/chris/dev/test/foo/package.json)
        Tool binaries available:
            ember (default)
    ```

This affects both the `--format=plain` and the `--format=human` variants, which is expected: the underlying logic for both is identical.

charlespierce:
The 2nd part of this is potentially tricky for a couple of reasons: First, we don't have an obvious way to determine the version of a local binary: It would be equally wrong to show `ember-cli@3.17.0 (current @ ...)` when the local project is actually using `ember-cli@3.12.0`. The only way I can think of to determine that would be to dig into `node_modules/ember-cli/package.json` and read the version from there (we can't use the project's `package.json` because it could be a semver range).

Additionally, the logic for determining the platform that a binary will use is currently bespoke logic inside of `volta_core::run::binary`, so in order to duplicate it we would probably want to refactor it out into somewhere reusable. That would actually simplify a few other things (like the `volta run` implementation), so it's worth doing, but it's not a super simple change.

chriskrycho:
I was thinking about that first part, too, but hadn't worked out the details on the second part yet. That's tricky indeed. 🤔 

Right now, we're printing something that's actually out-and-out *wrong*, so I'm wondering if we shouldn't tackle this in two phases:

1. Don't print the platform or version info for it *at all* for local binaries; just print `ember-cli (current @ ...)` or something. That gives the user enough info to introspect the version manually, while not printing *erroneous* information.

2. Design a solution which correctly handles the bits to go ahead and supply that data *correctly* after landing `volta run`.

Thoughts? cc. @rwjblue @dherman

charlespierce:
[...]

PR #740, opened 2020-05-19
Remove erroneous information from 'volta list' output

charlespierce:
Info
-----
* As discussed in #711, the output of `volta list` (both formatters) around project-local binaries is inaccurate.
* Though we don't yet have a clear path forward for determining the version / platform for project-locals, we can remove the incorrect information so as not to confuse the user.

Changes
-----
* Updated the `list::Package` data model to not include version or platform for local packages.
* Updated the logic for detecting a local package to not try to match on version (which didn't work anyway, because package.json is almost always filled with semver requirements).
* Changed both `human` and `plain` formatters to elide the incorrect information.

Tested
-----
* Updated the tests to represent the new output.

Notes / Questions
-----
* The `expected` output from the tests shows the way it looks for both formatters.
* Does this constitute a breaking change, since we are altering the output of the `plain` formatter, which is intended to be scripted against? I lean towards Yes, but I could see an argument either way.
* If so, is fixing it again (when resolving #711) _another_ breaking change, since we'd be moving back to the old style?
* Is there a better format we can use to show the user project-local bins without giving incorrect information?

### DIFF
--- a/src/command/list/plain.rs
+++ b/src/command/list/plain.rs
@@ -134,12 +134,6 @@ fn display_package(package: &Package) -> String {
             node,
             tools,
             ..
-        }
-        | Package::Project {
-            details,
-            node,
-            tools,
-            ..
         } => {
             let tools = match tools.len() {
                 0 => String::from(" "),
@@ -157,6 +151,14 @@ fn display_package(package: &Package) -> String {
                 package_source(&package)
             )
         }
+        Package::Project { name, tools, .. } => {
+            let tools = match tools.len() {
+                0 => String::from(" "),
+                _ => format!(" {} ", tools.join(", ")),
+            };
+
+            format!("package {} /{}/{}", name, tools, package_source(&package))
+        }
         Package::Fetched(details) => format!(
             "package {} (fetched)",
             tool_version(&details.name, &details.version)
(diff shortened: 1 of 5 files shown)


src/command/list/plain.rs
```diff
@@ -346,17 +354,13 @@ mod tests {
...
-                    },
+                    name: "typescript".into(),
                     path: PROJECT_PATH.clone(),
-                    node: NODE_VERSION.clone(),
                     tools: vec!["tsc".into(), "tsserver".into()]
                 }])
                 .expect("Should always return a `String` if given a non-empty set")
                 .as_str(),
-                "package typescript@3.4.1 / tsc, tsserver / node@12.4.0 npm@built-in (current @ /a/b/c)"
+                "package typescript / tsc, tsserver / (current @ /a/b/c)"
```
chriskrycho, line 363: A couple notes here:

**First,** one of the design constraints for the `--format=plain` output was that it be *consistent* for each "type"—so all `package` listings have the same layout, and so on. This changes that subtly, in that you can now end up with `package typescript / ...` *or* `package typescript@<version> / ...`. I think this part is fine, though, because the idea was that the combination of the "type" at the beginning (`package`) and then splitting on the section separators (` / `) would be unambigous, and this hasn't changed that. 👍

**Second,** I suggest we supply `@project` as the version in this case instead of dropping it entirely. Doing so would mean that anyone parsing the string would still have the same *shape* to parse; they would just need to add a new case to handle: SemVer *or* `project`. When we fix it, they don't have to change *anything* as long as they didn't *delete* their SemVer parsing. (We can communicate that they *shouldn't* in our release notes/blog post/etc.)

The output, in this proposal, would end up looking like:

```
package typescript@project / tsc, tsserver / node@project npm@project (current @ /a/b/c)
```

This isn't the prettiest thing in the universe, but it remains unambiguous, is a minimal change for existing code and straightforward to add support for, and minimizes the change that users have to do later.

charlespierce: That makes sense, I'd also be interested to hear from @dherman and @mikrostew their thoughts on the design.

charlespierce:
Updated to use 'project' as the tag for "project-local" versions, which are tricky to determine.

PR #748, opened 2020-05-26
v0.8.1

charlespierce:
Release fix for panic #746


joshuarli:
Looking forward to the new release! I'm about to have our CI install volta soon, so those smaller binary downloads are going to be nice.

PR #756, opened 2020-06-01
Use $HOME in profile scripts if possible

### LINKED ISSUE #689: Avoid hard coding home directory path into `.zshrc` in `volta setup`
Currently, running `volta setup` will add:

```sh
export VOLTA_HOME="/User/rjackson/.volta"
```

However, this makes a given profile "not portable" (I typically share my dotfiles across both *nix machines and my local Mac).

Instead, I think we should do:

```sh
export VOLTA_HOME="$HOME/.volta"
```

This has the same overall result (`VOLTA_HOME` is setup properly), but the location of the users home directory is not hard coded.

charlespierce:
Closes #689 

Info
-----
* The current `volta setup` command always writes the full path to the Volta directory into profile scripts.
* This can cause problems for users who share profile scripts across computers, where the `$HOME` part of the path might be different between machines.

Changes
-----
* Updated `volta setup` to check for two conditions:
    * Is `HOME` set?
    * Does the `VOLTA_HOME` we are going to write begin with the value of `$HOME`?
* If both are true, then we use `$HOME` when writing profile scripts, instead of using the full path.
* If either isn't true, then we write the entire path, exactly as we did previously.

Tested
-----
* Ran `volta setup` in normal conditions and confirmed that we write `export VOLTA_HOME="$HOME/.volta"` into the profile scripts, instead of the full path.
* Ran `volta setup` with VOLTA_HOME set to a Path that _doesn't_ include the home directory, and confirmed that the full path is written and `$HOME` is _not_ used.

### DIFF
--- a/src/command/setup.rs
+++ b/src/command/setup.rs
@@ -41,6 +41,7 @@ mod os {
 
     pub fn setup_environment() -> Fallible<()> {
         let home = volta_home()?;
+        let formatted_home = format_home(home.root());
 
         debug!("Searching for profiles to update");
         let profiles = determine_profiles()?;
@@ -53,7 +54,7 @@ mod os {
                 _ => write_profile_sh,
             };
 
-            match write_profile(&profile, contents, home.root()) {
+            match write_profile(&profile, contents, &formatted_home) {
                 Ok(()) => true,
                 Err(err) => {
                     warn!(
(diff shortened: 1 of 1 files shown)


rwjblue:
Thank you!

PR #763, opened 2020-06-11
[Small] Error when installing a package manager before Node

### LINKED ISSUE #575: Installing Yarn on a fresh setup fails subtly
On a fresh volta install with nothing cached:

```bash
$ volta install yarn
success: installed and set yarn@1.19.1 as default
$ yarn
Volta error: Node is not available.

To run any Node command, first set a default version using `volta install node`
Error details written to /Users/cpierce/.volta/log/volta-error-2019-10-16_18_32_00.897.log
$ volta install node
success: installed and set node@10.16.3 (with npm@6.9.0) as default
$ yarn
Volta error: Yarn is not available.

Use `volta install yarn` to select a default version (see `volta help install for more info).
```

Even though we installed Yarn in the first step, that is getting lost or overwritten when we subsequently install Node. Since Node is the core tool that runs all of the other tools, we should require that the user install Node first. This is similar to how we require the user pin Node before pinning yarn:

```bash
$ volta pin yarn
error: Cannot pin Yarn because the Node version is not pinned in this project.

Use `volta pin node` to pin Node first, then pin a Yarn version.
```

We also, for unrelated reasons, require the users to have Node installed before running `volta install <package>`, so we should make the UX consistent between the tools.

charlespierce:
Closes #575 

Info
-----
Currently, calling `volta pin yarn` or `volta pin npm` when Node isn't yet pinned will (correctly) show an error. However, the same is not true for `volta install npm` or `volta install yarn`. Instead, `volta install npm` without a Node version set will report success, however then trying to call `npm` immediately after will fail.

Even worse, if you follow the instructions and subsequently install Node, the fact that you had installed a version of `npm` or `yarn` is lost and we tell you to install them again.

Changes
-----
Updated `Toolchain::set_active_npm` and `Toolchain::set_active_yarn` to show an error if the platform isn't set (i.e. there is no Node version selected), exactly as `Project::pin_npm` and `Project::pin_yarn` do in the same situation.

Tested
-----
Updated tests to support the new behavior and added tests to confirm that the errors are correctly thrown.

Note
-----
While making this change, I noticed a small UX issue: Currently on either `volta pin` _or_ `volta install`, we go through the entire "resolve -> fetch" process first, and only show the error once we go to persist the changes. We should be able to detect that the install / pin will fail earlier and show an error immediately, without waiting to resolve or fetch the version.

Opened #764  to address that separately.

### DIFF
--- a/crates/volta-core/src/error/kind.rs
+++ b/crates/volta-core/src/error/kind.rs
@@ -184,6 +184,11 @@ pub enum ErrorKind {
     /// Thrown when Yarn is not set at the command-line
     NoCommandLineYarn,
 
+    /// Thrown when a user tries to install a Yarn or npm version before installing a Node version.
+    NoDefaultNodeVersion {
+        tool: String,
+    },
+
     /// Thrown when there is no Node version matching a requested semver specifier.
     NodeVersionNotFound {
         matching: String,
@@ -842,6 +847,13 @@ Please ensure you have a Node version selected with `volta {} node` (see `volta 
 
 Use `volta run --yarn` to select a version (see `volta help run` for more info)."
             ),
+            ErrorKind::NoDefaultNodeVersion { tool } => write!(
+                f,
+                "Cannot install {} because the default Node version is not set.
+
+Use `volta install node` to select a default Node first, then install a {0} version.",
+                                tool
+            ),
             ErrorKind::NodeVersionNotFound { matching } => write!(
                 f,
                 r#"Could not find Node version matching "{}" in the version registry.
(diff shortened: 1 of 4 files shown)


rwjblue:
> Currently, calling volta pin yarn or volta pin npm when Node isn't yet pinned will (correctly) show an error. 

I'm curious, how does this work with `volta run`? I would expect a _similar_ issue to exist (though maybe we already error nicely) for `volta run --yarn=1.22.0 yarn install`.

charlespierce:
For `volta run` we currently do error nicely, letting the user know that there isn't a platform and that they need pick a Node version (I think there might be room to improve it, since we can detect that we're going through `volta run` and show a more targeted CTA, but that's a separate issue).

The underlying issue here is that in the current release, when it gets to the final step of `volta install`—actually writing the defaults to `~/.volta/tools/user/platform.json`—it would detect that there was no Node version and just no-op, reporting success.

rwjblue:
Awesome, thank you for explaining @charlespierce!

PR #765, opened 2020-06-12
[Small] Move node metadata into metadata submodule

charlespierce:
Info
-----
* Having the metadata about node versions defined in the `resolve` submodule but then parsed in  the `serial` submodule made it a bit confusing to find where it was defined and how it was used.

Changes
-----
* Renamed `node::serial` to `node::metadata`
* Moved the definition of the metadata types into the `metadata` module.

Tested
-----
* This is purely a reorganization of code, no functional changes should happen.
* All tests pass.


charlespierce:
On second thought, I think I need to work through the full refactor before making smaller changes, as these things might change during that process.

PR #771, opened 2020-06-18
[Medium] Fetch Yarn info from registry, instead of GitHub API

### LINKED ISSUE #672: Cannot install older versions of Yarn
On a clean installation, running:

```sh
volta install yarn@1.7
```

Fails with the following output:

```sh
❯ volta install yarn@1.7
error: Could not find Yarn version matching ">= 1.7.0 < 1.8.0" in the version registry.

Please verify that the version is correct.
```

---

Initial triaging by @charlespierce and @chriskrycho in the discord chat suggests that the issue is that volta is using the GitHub API (this endpoint specifically https://api.github.com/repos/yarnpkg/yarn/releases), and that endpoint only returns the 30 most recent releases (see [docs here](https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository)). 

I think we'll need to follow the pagination info to get a full list of versions in order to fix the underlying issue. However, at the moment, perfect "fine" / "safe" projects are broken due to new (normal) yarn releases.

### LINKED ISSUE #678: Allow specifying GitHub authentication token for `yarn` version resolution
Currently, when running `volta install yarn@1` or `volta install yarn@1.22` volta will attempt to download the list of known releases from `https://api.github.com/repos/yarnpkg/yarn/releases` so that it can resolve the SemVer range accordingly.

https://github.com/volta-cli/volta/blob/83578540a04d6db1a503aea9936c56cec5c407df/crates/volta-core/src/tool/yarn/resolve.rs#L73-L82

Unfortunately, this results in `volta install yarn@1` erroring when the GitHub REST API limit is reached. This can be very problematic when, for example, running `volta install yarn@1` from a network that is using NAT to masquerade many machines behind a single IP address (which is what is happening on most corporate networks, and in many CI systems).

Solving the general case (e.g. corporate networks, with volta running on developer machines) seems quite difficult, but I think when running volta from CI systems we could easily provide a GitHub token to properly authenticate. For example, GitHub Actions provide a [built in read only token](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token). 

---

I'd like to be able to specify a command line flag or environment variable (e.g. `VOLTA_GITHUB_TOKEN`) that would be used to make these requests authenticated to avoid this "random" rate limiting issue.

charlespierce:
Closes #672 
Closes #678 

Info
-----
The current behavior for fetching Yarn is to use the GitHub API for both the index and the downloads. However, this has a couple of issues:
* The GitHub API is rate limited, so can run into problems when running in shared CI environments or behind corporate networks.
* The GitHub API by default only shows the most recent 30 releases, which as of right now means we only get information about Yarn 1.10.0 and later.
To resolve these, as well as start the groundwork necessary to support Yarn 2.0, we can use the npm registry to resolve and fetch Yarn versions (in the same way that we use it for `npm` itself).

Changes
-----
* Updated `yarn::resolve` to use the npm registry to resolve SemVer versions and tags, with caveats for hooks:
    * To preserve the hooks API contract, Volta will continue to expect the GitHub API response format for any data fetched using hooks.
    * This includes errors for using tags that aren't `latest` while hooks are defined, since tags were not supported previously.
    * If there are no hooks, then Volta will use the new behavior and fetch from the npm registry.
* Updated `yarn::fetch` to use the npm registry as the default URL for downloading Yarn tarballs.

Tested
-----
* Updated tests to support the new URLs in mocks.
* Added tests to confirm the behavior when hooks are used
* Added a smoke test to confirm that we can download versions of Yarn older than 1.10.0

Notes
-----
* For compatibility, the filenames in the Volta `inventory` directory are remaining the same, so we download the file from npm but then rename it to match the existing expected cache filename.


charlespierce:
Rebased now that #769 is merged.

chriskrycho:
For the caveats around hooks—

> - To preserve the hooks API contract, Volta will continue to expect the GitHub API response format for any data fetched using hooks.
> - This includes errors for using tags that aren't latest while hooks are defined, since tags were not supported previously.
> - If there are no hooks, then Volta will use the new behavior and fetch from the npm registry.

—should we consider doing a phased rollout of a change here, i.e. a deprecation in the release where we add this, with some well-specified time window (1 month or 3 months or 6 months or something) before changing it in a 0.9 release?

crates/volta-core/src/tool/package/fetch.rs
```diff
@@ -2,10 +2,11 @@
 
 use std::fs::{rename, write, File};
 use std::io::{Read, Seek, SeekFrom};
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
+use super::super::registry::find_unpack_dir;
```
chriskrycho, line 7: Nit: once I hit `super::super::` (really, anything beyond one `super`) I tend to reach for `crate::` instead because I find that easier to follow. Not a thing I think you *need* to change, by any means, just my own feeling.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -8,56 +12,90 @@ use crate::session::Session;
...
+    }
+}
+
+fn resolve_tag(tag: VersionTag, hooks: Option<&ToolHooks<Yarn>>) -> Fallible<Version> {
+    // This triage is complicated because we need to maintain the legacy behavior of hooks
+    // First, if the tag is 'latest' and we have a 'latest' hook, we use the old behavior
+    // Next, if the tag is 'latest' and we _do not_ have a 'latest' hook, we use the new behavior
+    // Next, if the tag is _not_ 'latest' and we have an 'index' hook, we show an error since
+    //     the previous behavior did not support generic tags
+    // Finally, we don't have any relevant hooks, so we can use the new behavior
```
chriskrycho, line 36: I totally get wanting to maintain legacy behavior and avoid introducing surprises. However, it seems inconsistent to add support for the new behavior of `'latest'` but *not* other tags—what's the failure mode there? (I assume there is one that just isn't obvious to me! 😅)

charlespierce: It may be that the comment isn't clear, the goal is that if you have hooks set, we use the old behavior (which _only_ supported the `latest` tag and no others), while if you don't have hooks, then we use the new logic. The cases are:

1. We have a `latest` hook and are looking up the `latest` tag: This was supported under the old system, so we can use the legacy logic for resolving latest.
2. We are looking up the `latest` tag but _don't_ have the `latest` hook: Previously, that would use the default, public registry, so we can port that over to use the new latest logic.
3. We are looking up a custom tag and we have an `index` hook: This was _not_ supported previously. Since we'd be using the `index` hook to look up data in the GitHub API format, there would be no way to resolve a tag anyway, so we can fail fast.
4. Last if we don't have an `index` hook either, then we are using the new default behavior from `npm` (the most common case).

We could potentially change it to have a slightly clearer break:

If you have any hooks defined (i.e. if the `hooks` parameter is `Some(_)`), then we always use the old logic, regardless of whether we actually have a relevant hook or not. If you have no hooks then we use the new logic.

That would simplify the split here, but would also mean that adding an `index` hook would change the behavior of `volta install yarn@latest`, which seems like it would introduce more confusion.

crates/volta-core/src/tool/package/fetch.rs
charlespierce: Yeah, it's pretty close between `super::super::` and `crate::tool::`. Not a big deal either way, I think my personal cutoff is at 3 😀

charlespierce:
> should we consider doing a phased rollout of a change here, i.e. a deprecation in the release where we add this, with some well-specified time window (1 month or 3 months or 6 months or something) before changing it in a 0.9 release?

For this, do you mean phasing out / deprecating the old hooks behavior, and eventually replacing it with hooks that match the new expectation (i.e. the npm Registry format)? That's a reasonable possibility, though my concern would be that we would want to set up some transitional hook logic that supports _both_ formats (with some sort of detection logic). Otherwise there's always going to be some hard cutoff where the format makes a sharp change, which is difficult to deal with in a large organization. If you've got hooks specified in multiple different projects, you need to coordinate upgrading everyone's Volta version _and_ everyone's hooks simultaneously, which is very hard.

To that point, I've actually been somewhat unhappy with the hooks format, now that it's got some use we've had a chance to see how it works out in practice, so I've been trying to come up with a different way of customizing the download locations. I will likely create an RFC for that at some point in the next couple weeks, as there are a number of interrelated constraints.

crates/volta-core/src/tool/yarn/resolve.rs
chriskrycho: This makes sense enough to me to proceed. My *only* remaining question is: why not just use the *new* behavior for everything (which I think would simplify things) and mark it as a breaking change—perhaps with a feature flag approach, doing a release which only supports the old way but deprecates it, then a bit later enabling the feature flag and removing the old way? (Not a blocker by any stretch.)

PR #774, opened 2020-06-22
[Small] Add extension and filename replacements for template hooks

### LINKED ISSUE #773: Provide additional replacements for template hooks
Noted on Discord: https://discordapp.com/channels/413753499835301908/413753499835301917/723696616317386783

Our current hooks aren't great for working with existing Node mirror sites. The `prefix` hook doesn't work correctly because Node mirrors have an extra `/v{{version}}/` directory in them, while the `template` hook doesn't work well because there's no replacement for the extension (`.tar.gz` vs `.zip`) or the filename as a whole. This makes it impossible to have hooks that work cross-platform to point to a Node mirror site.

Unfortunately, we can't change the `prefix` hook without it being a breaking change. However, we can _add_ `{{ext}}` and `{{filename}}` to the potential replacements in the prefix hook without breaking existing hooks. That will allow the prefix hook, at least, to work for Node mirrors.

Longer-term, we've run into a few issues with the scalability of hooks and ideally would like to re-think them in some way.

charlespierce:
Closes #773 

Info
-----
* As noted in the issue, our hooks are currently inadequate to handle the (common) use-case of a Node mirror.
* Adding new replacements to the `template` hooks is a backwards-compatible change we can make to improve this behavior.
* In order to support cross-platform Node mirrors, we need to provide the file extension at a minimum, however we can also use the expected/public file name as a replacement for simplicity.

Changes
-----
* Added `{{ext}}` to the distro hook replacement, pulling the extension from the expected filename.
    * Note: This logic needed to be customized, since the "extension" could be multi-part, in the case of tarballs (`tar.gz`)
* Added `{{filename}}` to the distro hook replacement, using the expected filename.

Tested
-----
* Added tests around the hook resolution and the calculation of the extension.

### DIFF
--- a/crates/volta-core/src/hook/tool.rs
+++ b/crates/volta-core/src/hook/tool.rs
@@ -16,6 +16,8 @@ use semver::Version;
 const ARCH_TEMPLATE: &str = "{{arch}}";
 const OS_TEMPLATE: &str = "{{os}}";
 const VERSION_TEMPLATE: &str = "{{version}}";
+const EXTENSION_TEMPLATE: &str = "{{ext}}";
+const FILENAME_TEMPLATE: &str = "{{filename}}";
 
 lazy_static! {
     static ref REL_PATH: String = format!(".{}", std::path::MAIN_SEPARATOR);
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/hook/tool.rs
```diff
@@ -47,6 +52,31 @@ impl DistroHook {
...
+/// the standard extension.
+fn calculate_extension(filename: &str) -> Option<&str> {
+    let mut parts = filename.rsplit('.');
+    match (parts.next(), parts.next(), parts.next()) {
+        (Some(ext), Some("tar"), Some(_)) => {
+            // .tar.gz style extension, return both parts
+            //                          tar  .   gz
+            let index = filename.len() - 3 - 1 - ext.len();
+            filename.get(index..)
+        }
```
chriskrycho, line 67: Could we not just return `"." + tar + "." + ext` (modulo the change to make `+` work here)? It seems like it would be more robust against conceivable variations in extensions, vs. the index-based approach. (Presumably similar would go for the later matches as well.)

charlespierce: In this case, it’s the literal value `tar` that we’re looking for, it’s not a generic part, so we can be confident that the index is valid.

We could use something like `”.tar.” + ext`, however that would also require a bit of gymnastics around the return type and creating the new `String`, and I’m not sure if it’s ultimately more readable.

Another option would be to use a regex, though that’s (imo anyway) often even more opaque than any of the existing options. Also a bit of a pain to maintain long-term, because reading regex is always an adventure.

chriskrycho: I just realized (thanks to @charlespierce telling me so in Discord 😂) that I had misread this; this seems fine, since it's using `ext.len()`—for some reason I totally misread it and thought it was hard coded.

Issue #761, opened 2020-06-09
`volta list` says no tools installed when package is not configured for volta


chriskrycho:
In a folder which has a `package.json`, but no `volta` key set in it, running `volta list` outputs:

```
⚡️ No Node runtimes installed!

    You can install a runtime by running `volta install node`. See `volta help install` for
    details and more options.
```

I would expect this to (and I'm pretty sure it *used* to) instead print the user's defaults. I'm guessing we accidentally short-circuit when there's no `volta` key locally.

charlespierce:
From what I can tell looking at the history of `toolchain.rs`, I don't believe this ever worked in this scenario. It looks like the code in [`Lookup::active_tool`](https://github.com/volta-cli/volta/blob/42918c1c083a55164079ddc9c2888a829dbde6fd/src/command/list/toolchain.rs#L83) is attempting to duplicate the logic that `volta_core` uses to determine the project platform.

A short-term fix for this would be to make this logic aware of the fact that the project platform may not contain all of the tools, and fall-back to the default if it can't find it for any reason in the project.

Long-term, since we now store the `Source` per-tool in `Platform`, it probably makes sense to pull the platform information from `Platform::current` in the same way that the shims do. That way we can be confident that we are always getting the same information and not duplicating the behavior.

chriskrycho:
Curious! I must have missed this scenario in my tests originally—makes some sense; I was worried about making sure *partial* platforms resolved correctly, but never tried the *no* platform scenario it seems.

I think we also talked about merging those bits of source information originally, but the `Platform`-side wasn't ready (and wouldn't be for a while), so didn't pursue it. If I get some time, I'll push this forward!

detj:
Not sure if I'm facing this same issue. I uninstalled all prior node, npm, npm global packages before installing volta via:

```sh
$ curl https://get.volta.sh | bash
```

And then installed latest stable LTS node by running:

```sh
$ volta install node@12
```

But when I run `volta list`, it says `⚡️ No Node runtimes installed!`.

```sh
$ volta list
⚡️ No Node runtimes installed!

    You can install a runtime by running `volta install node`. See `volta help install` for
    details and more options.
```

But when I run `volta list all`, it outputs this

```sh
$ volta list all
⚡️ User toolchain:

    Node runtimes:
        v12.18.3

    Package managers:


    Packages:
        @cloudflare/wrangler@1.9.2 (default)
            binary tools: wrangler
            platform:
                runtime: node@12.18.3
[...]

PR #778, opened 2020-06-29
Fix #712

### LINKED ISSUE #712: `volta list` does not display defaults when project version is set
The design for `volta list` specified that we would output both the default version from the user's toolchain and the version set in a local package:

```sh
$ volta list all --format=plain
runtime node@v12.2.0
runtime node@v11.9.0
runtime node@v10.15.3 (default)
runtime node@v8.16.0 (current @ ~/node-only/package.json)
```

We currently only print the `(current @ ...)` when in a package with one set, however:

```sh
$ volta list all --format=plain
runtime node@v12.2.0
runtime node@v11.9.0
runtime node@v10.15.3
runtime node@v8.16.0 (current @ ~/node-only/package.json)
```

As expected, this affects both "plain" and "human" formats.

2hu12:
Fix #712: `volta list` does not display defaults when project version is set

Example output of release version
image

Example output of local dev version
image

### DIFF
--- a/src/command/list/toolchain.rs
+++ b/src/command/list/toolchain.rs
@@ -41,7 +41,7 @@ enum Lookup {
 }
 
 impl Lookup {
-    fn version_from_spec(self) -> impl Fn(&PlatformSpec) -> Option<Version> {
+    fn version_from_spec(&self) -> impl Fn(&PlatformSpec) -> Option<Version> + '_ {
         move |spec| match self {
             Lookup::Runtime => Some(spec.node.clone()),
             Lookup::Npm => spec.npm.clone(),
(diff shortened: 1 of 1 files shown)


src/command/list/toolchain.rs
```diff
@@ -55,6 +55,17 @@ impl Lookup {
         default_platform: Option<&PlatformSpec>,
         version: &Version,
     ) -> Source {
+        let default_or_none = default_platform
+            .clone()
```
charlespierce, line 59: suggestion: It looks like this clone is redundant. It was there before, so if we're refactoring it makes sense to also clean that up.

```suggestion
```

2hu12:
> Thanks for this @2hu12! Out of curiosity, do you have an example of the updated output when there is a project version?
> 
> I don't think this actually fixes #712, as that issue is around Packages, which are loaded differently than the runtime or package mangers. The code for loading the list of packages (and their sources) is actually in `command/list/mod.rs`. That said, this looks like a nice fix that improves the output for runtimes / package managers in the case that the version is inherited.

src/command/list/toolchain.rs
2hu12: For sure, thanks!

charlespierce (approved):
Hi @2hu12 I completely dropped the ball on this, my apologies! You were right to be confused by my comment, I think I was looking at #711 when I made that reference to packages.

As a bonus, I believe this also fixes #761 and #922 as they all have the same root cause.

PR #784, opened 2020-07-21
[Small] Retry 'rename' operations to avoid race condition with virus scanners

### LINKED ISSUE #781: Installation of Node Version failing on Windows 10 due to McAfee Virus Scanner collision
Using Volta v0.8.3 on Windows 10,  I've experienced intermittent but frequent issues with node installation failures.  In my instance this appears due to a clash with the McAfee Virus Scanner.   However I'd suspect it also apply to any real-time virus scanner being used on a Windows 10 installation as well.

### **Issue Manifestation**
Run a Command to install a version of node.  This example is for node 12 but it applies to other versions as well
The screenshot below shows that this fails
install-node12

Inspecting the error log shows the follow error
```
"C:\\Program Files\\Volta\\volta.exe" "install" "node@12"
Volta v0.8.3

Could not create environment for Node v12.18.2
at C:\Users\davecare\AppData\Local\Volta\tools\image\node\12.18.2

Please ensure you have correct permissions to the Volta directory.

Error cause: Access is denied. (os error 5)
```
### **Issue Analysis**
I ran a trace with procmon and the cause can be seen in the **access_denied**  message in the screen-show.  
McAfee-Access-Denied-On-Rename

The fact that the trace also shows the McAfee Scanner scanning the directory being renamed suggest that it has locked the directory as this is the root cause of the issue.

### **Resolution Proposal**
This is a common annoyance with real-time virus scanners and obviously a work-around is to stop the virus scanner.
However on many corporate networks this isn't a viable option so my ideal would be a programmatic resolution, which is why I've raised this issue request

A good example of a similar scenario and a subsequent fix is document by the rust installer project in this issue https://github.com/rust-lang/rustup/issues/1870.     It would be ideal if this pattern could be applied to volta to deal with this issue in the same way

charlespierce:
Closes #781 

Info
-----
* When installing a new tool version, we first unpack the tool into a temporary directory and then `rename` it to the proper location within the Volta directory.
* Unfortunately, on systems with Virus scanners, that `rename` can sometimes fail because the directory is in-use as the scanner reads it.
* A similar issue was fixed in Rustup here: https://github.com/rust-lang/rustup/pull/1873

Changes
-----
* Added the `retry` crate and helper method to retry `rename` if the problem is permissions (indicating one of the files is in use).
* Used the same Fibonacci backoff and ~28 second max duration as Rustup (we may want to monitor that and determine how often it's needed and how much time is actually needed).
* Updated all uses of `std::fs::rename` to use the helper method instead, ensuring they all use the retry behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,7 +88,7 @@ name = "atty"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.58 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.73 (registry+https://github.com/rust-lang/crates.io-index)",
  "termion 1.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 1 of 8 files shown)


charlespierce:
@dacarey If you have time, would you mind testing this change to see if it resolves the issue for you? You can download the installer created by the PR here: https://github.com/volta-cli/volta/suites/946881950/artifacts/11765267

dogmum99:
which file should be added for retry option , where could we find it in directory structure

PR #792, opened 2020-07-28
[Tiny] Update version requirement parsing to support hyphen ranges

charlespierce:
Info
-----
Our current `parse_requirements` function is appending an `=` to any requirement that starts with a digit. This causes hyphen ranges (e.g. `8.11 - 8.17`) to be turned into invalid requirements (`=8.11 - 8.17`), which in turn causes the parse to fail. It appears that `VersionReq::parse_compat` already correctly parses `1` as `=1`, and so on, so adding the additional `=` character is redundant for those cases.

@mikrostew Do you have more context on why we were adding the `=` character? Is there a common case that is missed by excluding that character?

Changes
-----
* Updated `parse_requirements` to avoid prepending the `=` character, but still strip off a leading `v`

Tested
-----
* Added a unit test for hyphen ranges, which now passes.
* All the existing tests pass without changes.

### DIFF
--- a/crates/volta-core/src/version/serial.rs
+++ b/crates/volta-core/src/version/serial.rs
@@ -12,16 +12,9 @@ use semver::{Compat, ReqParseError, VersionReq};
 // then serializing them to pass to `npm view`, they need to be handled in
 // a Node-compatible way (or we get the wrong version info returned).
 pub fn parse_requirements(src: &str) -> Result<VersionReq, ReqParseError> {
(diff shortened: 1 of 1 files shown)


charlespierce:
Update: Looking at the history of that file, it looks like the prepending of `=` pre-dates the new `semver-parser` dependency. So that may have been an attempt to patch over the differences between Cargo-style and Node-style parsing, which was fixed more robustly with `parse_compat`

PR #796, opened 2020-08-12
[Medium] Extend Locks through entire install process

charlespierce:
Info
-----
* With the implementation of #684, we are locking the Volta directory while fetching new versions of tools.
* However, we are _not_ maintaining the lock while performing additional actions (such as writing the default platform file or installing package dependencies).
* We want to make sure that the lock is maintained throughout the `volta install` process, while also ensuring that we don't get deadlocks from the `ensure_fetched` lock (which is needed because it can be called separately by the shims).
* Ultimately, the goal of `VoltaLock` is to be a _per-process_ lock, where we prevent other Volta processes from modifying the Volta directory, but allow the current one to make changes.

Changes
-----
* Refactored the `VoltaLock` type to keep a count of active locks and only release the file lock when all of them have been released—essentially a refcount style of locking.
* Ensured that each `install` function acquires and keeps a lock through the entire `volta install` process.
* Also added a lock to the `regenerate_shims_for_dir` function, to prevent overlapping shim modifications.
* Added a debug statement to the `Drop` impl to show when the lock is dropped.

Tested
-----
* Confirmed through debug statements that the lock is maintained throughout the entire install process for each tool.
* Confirmed that package installs that require a Node download don't deadlock.

### DIFF
--- a/crates/volta-core/src/error/kind.rs
+++ b/crates/volta-core/src/error/kind.rs
@@ -171,6 +171,9 @@ pub enum ErrorKind {
         errors: Vec<String>,
     },
 
+    /// Thrown when unable to acquire a lock on the Volta directory
+    LockAcquireError,
+
     /// Thrown when BinConfig (read from file) does not contain Platform info.
     NoBinPlatform {
         binary: String,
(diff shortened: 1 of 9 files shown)


charlespierce:
Will do, thanks!

chriskrycho:
Those explanatory notes in the code are fabulous. Thank you!

PR #809, opened 2020-09-04
Switch to reqwest for proxy support

### LINKED ISSUE #798: Can volta using proxy?
I come from china, due to the firewall, when using 'volta install ' command, the net speed is so slow.
I configured socks proxy to speedup the http/https/git, the installer script works, but 'volta install ' command faild.
Can you add this feature to volta?

gregjopa:
This PR changes the http library from `attohttpc` to `reqwest` for proxy support to resolve #798. 

The [reqwest library supports proxies](https://docs.rs/reqwest/0.10.8/reqwest/#proxies) by default and can be set using environment variables. Ex:

```bash
export https_proxy=http://proxy.giantcorp.com:8080/
# uses the proxy
volta install npm@latest
```

### DIFF
--- a/crates/volta-core/src/tool/node/resolve.rs
+++ b/crates/volta-core/src/tool/node/resolve.rs
@@ -265,6 +265,13 @@ fn resolve_node_versions(url: &str) -> Fallible<RawNodeIndex> {
                 .text()
                 .with_context(registry_fetch_error("Node", url))?;
 
+            let index: RawNodeIndex =
+                serde_json::de::from_str(&response_text).with_context(|| {
+                    ErrorKind::ParseNodeIndexError {
+                        from_url: url.to_string(),
+                    }
+                })?;
+
             let cached = create_staging_file()?;
 
             let mut cached_file: &File = cached.as_file();
@@ -288,13 +295,6 @@ fn resolve_node_versions(url: &str) -> Fallible<RawNodeIndex> {
 
             let expiry = create_staging_file()?;
             let mut expiry_file: &File = expiry.as_file();
-
-            let index: RawNodeIndex =
-                serde_json::de::from_str(&response_text).with_context(|| {
-                    ErrorKind::ParseNodeIndexError {
-                        from_url: url.to_string(),
-                    }
-                })?;
 
             write!(expiry_file, "{}", expires).with_context(|| {
                 ErrorKind::WriteNodeIndexExpiryError {
--- a/crates/volta-core/src/tool/node/resolve.rs
+++ b/crates/volta-core/src/tool/node/resolve.rs
@@ -1,6 +1,7 @@
 //! Provides resolution of Node requirements into specific versions, using the NodeJS index
 
 use std::fs::File;
+use std::io::Read;
 use std::io::Write;
 use std::str::FromStr;
 use std::time::{Duration, SystemTime};
@@ -15,7 +16,6 @@ use crate::session::Session;
 use crate::style::progress_spinner;
 use crate::tool::Node;
 use crate::version::{VersionSpec, VersionTag};
-use attohttpc::Response;
 use cfg_if::cfg_if;
 use fs_utils::ensure_containing_dir_exists;
 use hyperx::header::{CacheControl, CacheDirective, Expires, HttpDate, TypedHeaders};
@@ -226,8 +226,8 @@ fn read_cached_opt(url: &str) -> Fallible<Option<RawNodeIndex>> {
 }
 
 /// Get the cache max-age of an HTTP reponse.
-fn max_age(headers: &attohttpc::header::HeaderMap) -> u32 {
-    if let Ok(cache_control_header) = headers.decode::<CacheControl>() {
+fn max_age(response: &reqwest::blocking::Response) -> u32 {
+    if let Ok(cache_control_header) = response.headers().decode::<CacheControl>() {
         for cache_directive in cache_control_header.iter() {
             if let CacheDirective::MaxAge(max_age) = cache_directive {
                 return *max_age;
@@ -249,15 +249,11 @@ fn resolve_node_versions(url: &str) -> Fallible<RawNodeIndex> {
             debug!("Node index cache was not found or was invalid");
             let spinner = progress_spinner(&format!("Fetching public registry: {}", url));
 
-            let (_, headers, response) = attohttpc::get(url)
-                .send()
-                .and_then(Response::error_for_status)
-                .with_context(registry_fetch_error("Node", url))?
-                .split();
+            let mut response: reqwest::blocking::Response =
+                reqwest::blocking::get(url).with_context(registry_fetch_error("Node", url))?;
 
-            let response_text = response
-                .text()
-                .with_context(registry_fetch_error("Node", url))?;
+            let mut response_text = String::new();
+            response.read_to_string(&mut response_text).unwrap();
 
             let index: RawNodeIndex =
                 serde_json::de::from_str(&response_text).with_context(|| {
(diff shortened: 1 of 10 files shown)


crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -249,15 +249,11 @@ fn resolve_node_versions(url: &str) -> Fallible<RawNodeIndex> {
...
-                .and_then(Response::error_for_status)
-                .with_context(registry_fetch_error("Node", url))?
-                .split();
+            let mut response: reqwest::blocking::Response =
+                reqwest::blocking::get(url).with_context(registry_fetch_error("Node", url))?;
 
-            let response_text = response
-                .text()
-                .with_context(registry_fetch_error("Node", url))?;
+            let mut response_text = String::new();
```
gregjopa, line 255: I struggled with this code here. The goal is to make a blocking request and get both the response body and also the response headers which is needed lower in the file. Let me know if there's a better approach.

charlespierce:
Thanks for this @gregjopa! I'll take a look later today!

crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -249,15 +249,11 @@ fn resolve_node_versions(url: &str) -> Fallible<RawNodeIndex> {
             debug!("Node index cache was not found or was invalid");
             let spinner = progress_spinner(&format!("Fetching public registry: {}", url));
 
-            let (_, headers, response) = attohttpc::get(url)
-                .send()
-                .and_then(Response::error_for_status)
-                .with_context(registry_fetch_error("Node", url))?
-                .split();
+            let mut response: reqwest::blocking::Response =
+                reqwest::blocking::get(url).with_context(registry_fetch_error("Node", url))?;
```
charlespierce, line 253: I think we want to make sure to preserve the `.and_then(Response::error_for_status)` call here, which converts an error status code (4XX or 5XX) into an actual error.

crates/volta-core/src/tool/npm/resolve.rs
```diff
@@ -42,11 +42,12 @@ fn fetch_npm_index(hooks: Option<&ToolHooks<Npm>>) -> Fallible<(String, PackageI
...
     let spinner = progress_spinner(&format!("Fetching public registry: {}", url));
-    let metadata: RawPackageMetadata = attohttpc::get(&url)
+    let http_client = Client::new();
+    let metadata: RawPackageMetadata = http_client
+        .get(&url)
         .header(ACCEPT, NPM_ABBREVIATED_ACCEPT_HEADER)
         .send()
-        .and_then(Response::error_for_status)
-        .and_then(Response::json)
+        .and_then(|resp| resp.json())
```
charlespierce, line 50: Similarly here, we should preserve the `.and_then(Response::error_for_status)` call. I believe we can also keep the existing point-free style of `.and_then(Response::json)` as well, which avoids creating a new closure.

crates/volta-core/src/tool/package/resolve.rs
```diff
@@ -183,17 +182,14 @@ fn npm_view_command_for(name: &str, version: &str, session: &mut Session) -> Fal
 // fetch metadata for the input url
 fn resolve_package_metadata(package_name: &str, package_info_url: &str) -> Fallible<PackageIndex> {
     let spinner = progress_spinner(&format!("Fetching package metadata: {}", package_info_url));
-    let response_text = attohttpc::get(package_info_url)
-        .send()
-        .and_then(Response::error_for_status)
-        .and_then(Response::text)
+    let response_text = reqwest::blocking::get(package_info_url)
+        .and_then(|resp| resp.error_for_status())
+        .and_then(|resp| resp.text())
```
charlespierce, line 187: Similarly here, I believe we should be able to keep the same point-free style with `.and_then(Response::error_for_status).and_then(Response::text)` to avoid creating unnecessary closures.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -72,11 +72,12 @@ fn resolve_semver(matching: VersionReq, hooks: Option<&ToolHooks<Yarn>>) -> Fall
...
     let spinner = progress_spinner(&format!("Fetching public registry: {}", url));
-    let metadata: RawPackageMetadata = attohttpc::get(&url)
+    let http_client = Client::new();
+    let metadata: RawPackageMetadata = http_client
+        .get(&url)
         .header(ACCEPT, NPM_ABBREVIATED_ACCEPT_HEADER)
         .send()
-        .and_then(Response::error_for_status)
-        .and_then(Response::json)
+        .and_then(|resp| resp.json())
```
charlespierce, line 80: Same here, we should maintain the `Response::error_for_status` and the calling style where possible.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -96,10 +97,8 @@ fn resolve_custom_tag(tag: String) -> Fallible<Version> {
 }
 
 fn resolve_latest_legacy(url: String) -> Fallible<Version> {
-    let response_text = attohttpc::get(&url)
-        .send()
-        .and_then(Response::error_for_status)
-        .and_then(Response::text)
+    let response_text = reqwest::blocking::get(&url)
+        .and_then(|resp| resp.text())
```
charlespierce, line 101: Same comment here.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -133,10 +132,8 @@ fn resolve_semver_from_registry(matching: VersionReq) -> Fallible<Version> {
 
 fn resolve_semver_legacy(matching: VersionReq, url: String) -> Fallible<Version> {
     let spinner = progress_spinner(&format!("Fetching public registry: {}", url));
-    let releases: RawYarnIndex = attohttpc::get(&url)
-        .send()
-        .and_then(Response::error_for_status)
-        .and_then(Response::json)
+    let releases: RawYarnIndex = reqwest::blocking::get(&url)
+        .and_then(|resp| resp.json())
```
charlespierce, line 136: Same comment here.

crates/volta-core/src/tool/node/resolve.rs
charlespierce: Yeah, the lack of a way to destructure the `Response` makes this tricky. I think the easiest approach might be to determine the `expires` string _before_ reading the data with something like:
```rust
let expires = if let Ok(expires_header) = response.headers().decode::<Expires>() {
    expires_header.to_string()
} else {
    let expiry_date = SystemTime::now() + Duration::from_secs(max_age(&response).into());
    HttpDate::from(expiry_date).to_string()
};
```
Then we can use:
```
let response_text = response.text().with_context(registry_fetch_error("Node", url))?;
```
Here, and finally do `write!(expiry_file, "{}", expires).with_context(...)` at the end of the file. So we calculate everything we need from the response _first_, then consume it to get the raw text, and then use the pre-calculated bit later on when we need it.

charlespierce (changes requested):
This is great @gregjopa, thanks! A couple of comments on making sure to preserve the existing error behavior and a slightly modified approach to the destructuring problem you noted in `node/resolve.rs`

gregjopa:
Thanks for the feedback @charlespierce! Your detailed comments are helping me learn Rust 😄  I think I was able to address everything. Please have another look.

charlespierce (approved):
This looks great, thanks @gregjopa!

While I'm sad to lose the binary-size improvements from using `attohttpc`, I don't find any significant difference in the startup time for the hot-path shims, so it doesn't appear to impact the performance at all.

PR #811, opened 2020-09-05
[Medium] Generate config files and shims during package install

charlespierce:
Info
-----
* Part 2 of package install rework: Complete the `volta install <package>` flow.
* Parse `package.json` to determine the exact version and any bins that are installed by the package
* Write the package config, bin configs, and shims

Changes
-----
* Added `metadata` module with the various metadata files that are used. Each is similar to an existing `metadata` with the old flow, but implemented differently to fit the new workflow.
* Added `configure` module with methods for parsing the manifest and writing the configs and shims
* Restructured the data flow in the `Package::install` method to ensure all of the appropriate data is available and can be used. This was also done somewhat looking forward to enabling `npm i -g` to reuse much of the code.

Notes
-----
* This only represents the end of the `volta install` flow. The changes needed to the `run` flow so that bins can actually be executed will be added in the next PR
* The requirement that a package contain at least one bin was dropped. While not technically supported yet, the goal of these changes is to eventually support that (along with many other things), so it seemed reasonable to drop that requirement now instead of adding it and then removing it.

### DIFF
--- a/crates/volta-core/src/tool/package_global/mod.rs
+++ b/crates/volta-core/src/tool/package_global/mod.rs
@@ -40,13 +40,13 @@ impl Package {
     pub fn complete_install(self, image: &Image) -> Fallible<PackageManifest> {
         let manifest = self.parse_manifest()?;
 
-        self.write_config_and_shims(&manifest, &image)?;
         self.persist_install()?;
+        self.write_config_and_shims(&manifest, &image)?;
 
         Ok(manifest)
     }
 
-    fn persist_install(self) -> Fallible<()> {
+    fn persist_install(&self) -> Fallible<()> {
         let home = volta_home()?;
         let package_dir = new_package_image_dir(home, &self.name);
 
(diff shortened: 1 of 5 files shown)


crates/volta-core/src/tool/package_global/mod.rs
```diff
@@ -31,7 +37,16 @@ impl Package {
         })
     }
 
-    pub fn persist_install(self) -> Fallible<()> {
+    pub fn complete_install(self, image: &Image) -> Fallible<PackageManifest> {
+        let manifest = self.parse_manifest()?;
+
+        self.write_config_and_shims(&manifest, &image)?;
+        self.persist_install()?;
```
mikrostew, line 44: Seems like `persist_install()` should be called before `write_config_and_shims()`? So that for the case where there is an error with `persist_install()`, there won't be any shims or config written for a package that doesn't exist in the expected location. For this current order, if there is an error with `persist_install()`, you can have shims and config that exist, but the package may not be there.

mikrostew (changes requested):
Just one thing I noticed, but otherwise looks great!

crates/volta-core/src/tool/package_global/mod.rs
charlespierce: I went back and forth on this, since the reverse could happen as well (the directory gets overwritten and then writing the configs fails). But thinking about it more, for a new install, in that case at least we don't leave things with broken shims. Updated.

mikrostew (approved):

PR #814, opened 2020-09-08
[Small] Rework Shim code path to work with global packages

charlespierce:
Info
-----
* Part 3 of package install rework.
* With the changes to the directory structure and config format for global packages, the shim code path for actually running a global executable needs to be updated.

Changes
-----
* Updated `run::binary` to load binary information from the new configuration format when the feature flag is enabled.
* Changed calculation of the path to the binary to use the new directory structure for packages.

Tested
-----
* Manually tested on both Windows and Mac that executing a global package works correctly after installing it with `volta install`.


mikrostew (approved):
LGTM

PR #816, opened 2020-09-09
[Tiny] Complete removal of 'package' module under feature flag

charlespierce:
Info
-----
* Part 5 of package rework: Completely remove `tool::package` module when the feature flag is active
* Previously, to simplify implementation, both the `package` and `package_global` modules were available when the feature flag was enabled.
* Now that work has progressed far enough, we can completely replace the `package` module with the feature flag.

Changes
-----
* Updated the module definition to load `package_global/mod.rs` _as_ the module `package` when the feature flag is enabled.
* Added some more feature flags to prevent warnings with now-unused methods on `ToolCommand`
* Cleaned up some imports that were referencing `package_global` as a separate module.

Tested
-----
* Confirmed that Volta builds without errors / warnings both with and without the feature flag.

### DIFF
--- a/crates/volta-core/src/tool/mod.rs
+++ b/crates/volta-core/src/tool/mod.rs
@@ -9,9 +9,11 @@ use log::{debug, info};
 
 pub mod node;
 pub mod npm;
+#[cfg(not(feature = "package-global"))]
 pub mod package;
 #[cfg(feature = "package-global")]
-pub mod package_global;
+#[path = "package_global/mod.rs"]
+pub mod package;
 mod registry;
 mod serial;
 pub mod yarn;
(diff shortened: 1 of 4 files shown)


mikrostew (approved):
LGTM!

crates/volta-core/src/tool/mod.rs
```diff
@@ -9,9 +9,11 @@ use log::{debug, info};
 
 pub mod node;
 pub mod npm;
+#[cfg(not(feature = "package-global"))]
 pub mod package;
 #[cfg(feature = "package-global")]
-pub mod package_global;
+#[path = "package_global/mod.rs"]
+pub mod package;
```
chriskrycho, line 16: TIL you can do this!

PR #817, opened 2020-09-09
[Tiny] Remove Package Hooks under new install workflow

charlespierce:
Info
-----
* Part 6 of package rework: Remove the unused `Package` hooks under the new workflow
* Since we are now delegating to `npm install --global` to handle package installs, we no longer want to have a hybrid Hooks & `.npmrc` method for redirecting the download.
    * Even before the change, Package hooks were soft-deprecated and didn't fully work as expected
    * With the new workflow, we can cleanly remove them and rely only on the `.npmrc` files used by the package manager.

Changes
-----
* Marked the `Package` hook kinds as only available without the feature flag enabled.

Tested
-----
* Confirmed that Volta compiles without warnings or errors both with and without the feature flag.

### DIFF
--- a/crates/volta-core/src/hook/mod.rs
+++ b/crates/volta-core/src/hook/mod.rs
@@ -9,7 +9,9 @@ use std::path::Path;
 use crate::error::{Context, ErrorKind, Fallible};
 use crate::layout::volta_home;
 use crate::project::Project;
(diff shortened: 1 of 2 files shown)


mikrostew (approved):
So excited for this work to land

PR #819, opened 2020-09-09
Remove use of #[feature] and build docs on stable

charlespierce:
Info
-----
* Building the docs currently fails on nightly due to some incomprehensible `serde` errors.
* The only reason we need to use nightly is that we have a feature declared in `volta_core`: `#[feature(doc_cfg)]`, which claims to help support cross-platform docs.
* However, we don't seem to be taking advantage of that feature within the API docs anywhere.

Changes
-----
* Removed the `#[feature]` attribute that we weren't using anyway.
* Updated the CI configuration to build the API docs on stable Rust.

Tested
-----
* Ran `cargo doc` on stable to ensure that it completes correctly.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -18,7 +18,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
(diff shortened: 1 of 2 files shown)


chriskrycho:
🎉 stable everywhere!

PR #820, opened 2020-09-09
[Medium] Add new layout version for new package installation workflow

charlespierce:
Info
-----
* Part 8 of package rework: Create a new layout version to represent the new package installation process.
* This new layout is a small change from the previous (`package_image_dir` no longer uses `version`), however it also will serve as a marker for migrating existing package installations.

Changes
-----
* Added `v3` layout in `volta_layout`
* Updated `layout` module in `volta_core` to use `v3` layout when the feature flag is active.
* Removed `new_package_image_dir` function and switched to calling `package_image_dir` on the `VoltaHome` instance.
* Removed `inventory::package_available` method when the feature flag is on, since it isn't used any more.

Notes
-----
* This change only includes the updated `layout`, but does _not_ include a migration for that layout yet. The actual migration will be added in the next PR.

### DIFF
--- a/crates/volta-core/src/layout/mod.rs
+++ b/crates/volta-core/src/layout/mod.rs
@@ -6,7 +6,10 @@ use cfg_if::cfg_if;
 use double_checked_cell::DoubleCheckedCell;
 use dunce::canonicalize;
 use lazy_static::lazy_static;
+#[cfg(not(feature = "package-global"))]
 use volta_layout::v2::{VoltaHome, VoltaInstall};
+#[cfg(feature = "package-global")]
+use volta_layout::v3::{VoltaHome, VoltaInstall};
 
 cfg_if! {
     if #[cfg(unix)] {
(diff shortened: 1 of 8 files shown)


crates/volta-core/src/layout/mod.rs
```diff
@@ -6,7 +6,10 @@ use cfg_if::cfg_if;
 use double_checked_cell::DoubleCheckedCell;
 use dunce::canonicalize;
 use lazy_static::lazy_static;
+#[cfg(not(feature = "package-global"))]
 use volta_layout::v2::{VoltaHome, VoltaInstall};
+#[cfg(feature = "package-global")]
+use volta_layout::v3::{VoltaHome, VoltaInstall};
```
chriskrycho, line 12: Layout functionality continues to pay dividends! 🎉

chriskrycho (approved):
Looks good!

PR #825, opened 2020-09-11
[Medium] Refactor install logic to use bare functions

charlespierce:
Info
-----
* The next phase of the package rework involves supporting direct global installs (e.g. `npm install --global typescript`) to work within Volta.
* To do this, we will need to support a similar, but not identical interface for the install steps.
* To allow for reuse across different interfaces, refactor the various supporting methods into bare functions, which avoids them being tied to one specific struct (`Package`).

Changes
-----
* Update the `configure` and `install` modules in `package_global` to use bare functions that accept what they need as parameters.
* Similarly, updated the `persist_install` method to be a bare function.
* Added `run_install` to the `impl Package` block directly, so the entire public interface is in one place.

Tested
-----
* There should be no functional change due to this refactor, all tests still pass.

Notes
-----
* The changes look somewhat substantial, but the majority of them are reformatting (one fewer level of indentation with bare functions as opposed to associated functions) or changes related to using parameters directly, instead of properties.

### DIFF
--- a/crates/volta-core/src/tool/package_global/configure.rs
+++ b/crates/volta-core/src/tool/package_global/configure.rs
@@ -1,88 +1,90 @@
+use std::path::PathBuf;
+
 use super::metadata::{BinConfig, PackageConfig, PackageManifest};
-use super::Package;
 use crate::error::{ErrorKind, Fallible};
 use crate::layout::volta_home;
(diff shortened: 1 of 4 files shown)


chriskrycho (approved):
⚡ ship it!

PR #826, opened 2020-09-12
[Small] Add PackageManger type to install logic

charlespierce:
Info
-----
* In order to support direct global installs, we need to track the idiosyncrasies of the different package manager installs.
* This requires us to keep track of which package manager actually performed the install of a package.

Changes
-----
* Added a `PackageManager` enum to the `tool::package` module. This type has associated functions to determine the bin and source locations for a package and to set up a command to install into the appropriate directory.
* Updated all of the existing install logic to use the `PackageManager::Npm` type, since it uses `npm install --global` internally.
* Included the package manager in the package and bin config files, so they can be used on lookup if needed.
* Updated the binary execution to take advantage of the stored package manager and correctly locate the binary file.
* Added the package manager to the fixture data on the `volta_uninstall` tests.

Tested
-----
* Confirmed that all existing tests pass.
* Confirmed that the package manager is written into the package and binary configs when installing a package.

### DIFF
--- a/crates/volta-core/src/run/binary.rs
+++ b/crates/volta-core/src/run/binary.rs
@@ -110,14 +110,8 @@ pub struct DefaultBinary {
 impl DefaultBinary {
     #[cfg(feature = "package-global")]
     pub fn from_config(bin_config: BinConfig, session: &mut Session) -> Fallible<Self> {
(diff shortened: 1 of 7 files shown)


chriskrycho (approved):
Thank you again for breaking these up—makes it suuuuper easy to review. ⚡ Ship it!

PR #827, opened 2020-09-17
[Large] Refactor 'run' module to allow for better extensibility

charlespierce:
Info
-----
* Looking forward to allowing `npm i -g <package>` to work seamlessly within Volta, we need to ensure that the `run` module can handle more extensibility.
* Since there will be more possible branches (`npm i -g` will need to include extra behavior both before and after running a command), we will want to allow a command to be built up more piecemeal than is currently supported.
* We also want to leave things in a state that they can be easily extended to new use-cases (i.e. `npm i -g yarn` should delegate to the `volta install yarn` behavior, instead of treating it like any other package).

Changes
-----
* Added new `run_package_global` directory, used as the `run` module when the feature flag is enabled.
* Removed all the feature flag attributes from `run`, as it is now only enabled when the feature flag is off.
* Split the customization logic (error messages, specific Platform modifications, project-local delegations, etc.) into 2 phases:
    * First, on creation: Each submodule of `run_package_global` provides a function to create a new `ToolCommand`, annotated with the appropriate `ToolKind` so that it knows which command is being run.
    * Second, on execution: Each submodule _also_ provides a function to get the execution context (`PATH` value and happy error message if launching the tool fails), which is called by `ToolCommand` immediately before executing the command.
* This separation allows us to handle common modifications (like adding the CLI-provided Platform from `volta run`) in between the two customizations without having to duplicate / thread those modifications through the entire creation process.
* The implementation of the two points of customization in each submodule is based entirely on the existing implementation in the `run` directory, modified to handle the two points in the process as appropriate.
* Updated the duplicated debug output of `Using <tool>@<version> from <configuration>` to instead print out the entire `Image`, including Node, npm, and Yarn versions where appropriate.

Tested
-----
* Locally confirmed that running commands works as expected.
* Updated the acceptance tests to handle the new output format for the platform.

Notes
-----
* While there is a lot of new code, the bulk of the change is in the `run_package_global/mod.rs` and `run_package_global/executor.rs`. Everything else is the re-implementation of the existing tool-specific logic.
* The plan for customizing the install command is to add an `enum` wrapper around `ToolCommand` that allows different command variants (including ones that call the `volta install node`, `volta install yarn`, etc. logic, as well as the package logic).
* Since the next step is to implement the `npm i -g` logic, the global install interception was intentionally omitted.

### DIFF
--- a/crates/volta-core/src/run_package_global/binary.rs
+++ b/crates/volta-core/src/run_package_global/binary.rs
@@ -1,0 +1,165 @@
+use std::ffi::{OsStr, OsString};
+use std::path::PathBuf;
+
+use super::executor::{ToolCommand, ToolKind};
+use super::{debug_active_image, debug_no_platform};
+use crate::error::{ErrorKind, Fallible};
+use crate::layout::volta_home;
+use crate::platform::{Platform, Sourced, System};
+use crate::session::Session;
+use crate::tool::package::BinConfig;
+use log::debug;
+
+/// Determine the correct command to run for a 3rd-party binary
+///
+/// Will detect if we should delegate to the project-local version or use the default version
+pub(super) fn command(
+    exe: &OsStr,
+    args: &[OsString],
+    session: &mut Session,
+) -> Fallible<ToolCommand> {
+    let bin = exe.to_string_lossy().to_string();
+    // First try to use the project toolchain
+    if let Some(project) = session.project()? {
+        // Check if the executable is a direct dependency
+        if project.has_direct_bin(exe)? {
+            let path_to_bin =
+                project
+                    .find_bin(exe)
+                    .ok_or_else(|| ErrorKind::ProjectLocalBinaryNotFound {
+                        command: exe.to_string_lossy().to_string(),
+                    })?;
+
+            debug!("Found {} in project at '{}'", bin, path_to_bin.display());
+
+            let platform = Platform::current(session)?;
+            return Ok(ToolCommand::new(
+                path_to_bin,
+                args,
+                platform,
+                ToolKind::ProjectLocalBinary(bin),
+            ));
+        }
+    }
+
+    // Try to use the default toolchain
+    if let Some(default_tool) = DefaultBinary::from_name(exe, session)? {
+        debug!(
+            "Found default {} in '{}'",
+            bin,
+            default_tool.bin_path.display()
+        );
+
+        return Ok(ToolCommand::new(
+            default_tool.bin_path,
+            args,
+            Some(default_tool.platform),
+            ToolKind::DefaultBinary(bin),
+        ));
+    }
+
+    // At this point, the binary is not known to Volta, so we have no platform to use to execute it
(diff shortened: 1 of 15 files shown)


crates/volta-core/src/run_package_global/binary.rs
```diff
@@ -0,0 +1,165 @@
...
+        Ok(DefaultBinary { bin_path, platform })
+    }
+
+    pub fn from_name(tool_name: &OsStr, session: &mut Session) -> Fallible<Option<Self>> {
+        let bin_config_file = match tool_name.to_str() {
+            Some(name) => volta_home()?.default_tool_bin_config(name),
+            None => return Ok(None),
+        };
+
+        if bin_config_file.exists() {
```
charlespierce, line 158: Clean this up to use `BinConfig::from_file_if_exists` from previous PR.

crates/volta-core/src/run_package_global/binary.rs
```diff
@@ -0,0 +1,165 @@
...
+        let platform = Platform {
+            node: Sourced::with_binary(bin_config.platform.node),
+            npm: bin_config.platform.npm.map(Sourced::with_binary),
+            yarn: yarn.map(Sourced::with_binary),
+        };
+
+        Ok(DefaultBinary { bin_path, platform })
+    }
+
+    pub fn from_name(tool_name: &OsStr, session: &mut Session) -> Fallible<Option<Self>> {
```
chriskrycho, line 152: I think we should have *either* a doc-string explaining what `Some` and `None` mean, or make it a custom `enum` indicating `Found { bin_path, platform }` – probably the former! Just that way it's clear to any future readers (including us in a year 😂) what the intent is.

crates/volta-core/src/run_package_global/npx.rs
```diff
@@ -0,0 +1,48 @@
...
+    platform: Option<Platform>,
+    session: &mut Session,
+) -> Fallible<(OsString, ErrorKind)> {
+    match platform {
+        Some(plat) => {
+            let image = plat.checkout(session)?;
+
+            // npx was only included with npm 5.2.0 and higher. If the npm version is lower
+            // that that, we should include a helpful error message
+            let required_npm = parse_version("5.2.0")?;
```
chriskrycho, line 28: Follow-up per our discussion: maybe make this a `lazy_static!` that panics (presumably via `expect`) if it's done incorrectly, since users will never be able to do *anything* if it is wrong.

chriskrycho (approved):
:shipit: ⚡

crates/volta-core/src/run_package_global/binary.rs
charlespierce: Done!

crates/volta-core/src/run_package_global/npx.rs
charlespierce: Moved it to a `lazy_static` and actually used the infallible `Version::new` constructor, since we know the values for the version parts.

PR #833, opened 2020-09-21
[Small] Implement direct global uninstalls

charlespierce:
Info
-----
* Since we now support installing globals directly with `npm i -g <package>`, we should also support removing them with `npm uninstall -g <package>`
* This should always delegate to the internal Volta logic, so we can be sure that the shims and config files are correctly cleaned up.

Changes
-----
* Added a new `Executor` for uninstall commands, which calls out to the `tool::Spec::uninstall` method.
* Updated the parsing of npm and Yarn commands to check for global uninstalls and use the new Executor.

Tested
-----
* Added new tests to the parsing logic to confirm `uninstall` / `remove` are handled correctly.
* Added new acceptance tests to confirm that running `npm uninstall -g` or `yarn global remove` correctly uninstall all of the Volta extras.

Notes
-----
* As with the global installs, this currently doesn't support uninstalling multiple packages at once, only the first is read. That functionality will be added in a future PR.

### DIFF
--- a/crates/volta-core/src/run_package_global/executor.rs
+++ b/crates/volta-core/src/run_package_global/executor.rs
@@ -246,7 +246,7 @@ impl InternalInstallCommand {
         info!(
             "{} using Volta to install {}",
             note_prefix(),
-            tool_name(&self.tool)
+            self.tool.name()
         );
 
         self.tool.resolve(session)?.install(session)?;
@@ -279,7 +279,7 @@ impl UninstallCommand {
         info!(
             "{} using Volta to uninstall {}",
             note_prefix(),
-            tool_name(&self.tool)
+            self.tool.name()
         );
 
         self.tool.uninstall()?;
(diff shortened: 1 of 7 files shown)


crates/volta-core/src/run_package_global/executor.rs
```diff
@@ -259,3 +260,45 @@ impl From<InternalInstallCommand> for Executor {
...
+}
+
+fn tool_name(tool: &Spec) -> &str {
+    match tool {
+        Spec::Node(_) => "Node",
+        Spec::Npm(_) => "npm",
+        Spec::Yarn(_) => "Yarn",
+        Spec::Package(name, _) => &name,
+    }
+}
```
chriskrycho, line 304: **Suggestion:** make this a `pub fn` on `Spec` so that we can just write `&self.tool.name()` above; it seems likely it's the kind of thing that may be useful in the future but in any case clarifies the relationship now.

chriskrycho (approved):

crates/volta-core/src/run_package_global/executor.rs
charlespierce: Done, good idea!

Issue #823, opened 2020-09-11
Don't rely on exists() for filesystem operations


charlespierce:
As noted in this discussion: https://github.com/volta-cli/volta/pull/815#discussion_r487098657 For filesystem operations, we shouldn't do an `exists()` check first, followed by a File open. Instead we should directly open the file and then handle the "File not found" error case separately (as needed).

This prevents race conditions where the file is deleted _after_ the `exists()` check but before it is opened. It also reduces the complexity of the happy path slightly, not requiring an extra filesystem check before opening the target file.

BProg:
Hi, I can try to work on this

charlespierce:
@BProg Great, we'd be happy to have a contribution! Let me know if I can help.

BProg:
@charlespierce I think this can be closed

PR #834, opened 2020-09-22
Handle ErrorKind NotFound instead of file.exists()

BProg:
@charlespierce I am ok with any kind of comments. I would like to know if this is the correct direction you want to take in regards to [issue 823](https://github.com/volta-cli/volta/issues/823).

I have changed only the places that do an `open` operation on FS, and left alone other operations on FS like `rename` or `delete`
There may be some missing places that I could not find

### DIFF
--- a/crates/volta-core/src/error/mod.rs
+++ b/crates/volta-core/src/error/mod.rs
@@ -1,6 +1,6 @@
+use std::error::Error;
 use std::fmt;
 use std::process::exit;
-use std::{error::Error, io};
 
 mod kind;
 mod reporter;
@@ -45,20 +45,6 @@ impl VoltaError {
     pub fn kind(&self) -> &ErrorKind {
         &self.inner.kind
     }
-
-    /// Verifies if the inner source is `io::ErrorKind::NotFound`
-    pub fn is_not_found_error_kind(&self) -> bool {
-        self.source()
-            .and_then(|source| source.downcast_ref::<io::Error>())
-            .and_then(|io_err| {
-                if io_err.kind() == io::ErrorKind::NotFound {
-                    Some(true)
-                } else {
-                    None
-                }
-            })
-            .unwrap_or(false)
-    }
 }
 
 impl fmt::Display for VoltaError {
@@ -80,83 +66,6 @@ impl From<ErrorKind> for VoltaError {
         }
     }
 }
-
-pub trait AcceptableErrorToValue<T> {
-    fn accept_error_as_value_if<F1, F2>(self, accept_if: F1, as_value: F2) -> Fallible<T>
-    where
-        F1: FnOnce(&VoltaError) -> bool,
-        F2: FnOnce() -> T;
-}
-
-pub trait AcceptableErrorToDefault<T>: AcceptableErrorToValue<T>
-where
-    T: Default,
-    Self: Sized,
-{
-    /// Transforms an `Fallible<T>::Err(x)` into `Fallible<T>::Ok(T::default())`
-    ///
-    ///# Example
-    ///```
-    ///use std::io;
-    ///use io::Error;
-    ///use volta_core::error::{Fallible, VoltaError, ErrorKind};
-    ///use crate::volta_core::error::AcceptableErrorToDefault;
-    ///
-    ///let not_found = Error::new(io::ErrorKind::NotFound, "file not found");
-    ///let volta_error = VoltaError::from_source(not_found, ErrorKind::BinaryExecError);
-    ///let fallible: Fallible<i32> = Err(volta_error);
-    ///if let Ok(zero) = fallible.accept_error_as_default_if(|e| e.is_not_found_error_kind()) {
-    ///    assert_eq!(zero, 0);
-    ///} else {
-    ///    assert!(false, "accept_error_as_value_if is not equal to 10");
-    ///}
-    ///```
-    fn accept_error_as_default_if<F>(self, accept: F) -> Fallible<T>
-    where
-        F: FnOnce(&VoltaError) -> bool,
-    {
-        self.accept_error_as_value_if(accept, || T::default())
-    }
-}
-
-/// Extension to `Fallible<T>` to transform a `VoltaError` into a value
-impl<T> AcceptableErrorToValue<T> for Fallible<T> {
-    ///
-    ///# Example
-    ///```
-    ///use std::io;
-    ///use io::Error;
-    ///use volta_core::error::{Fallible, VoltaError, ErrorKind};
-    ///use crate::volta_core::error::AcceptableErrorToValue;
-    ///
-    ///let not_found = Error::new(io::ErrorKind::NotFound, "file not found");
-    ///let volta_error = VoltaError::from_source(not_found, ErrorKind::BinaryExecError);
-    ///let fallible: Fallible<i32> = Err(volta_error);
-    ///if let Ok(ten) = fallible.accept_error_as_value_if(|e| e.is_not_found_error_kind(), || 10) {
-    ///    assert_eq!(ten, 10);
-    ///} else {
-    ///    assert!(false, "accept_error_as_value_if is not equal to 10");
-    ///}
-    ///```
-    fn accept_error_as_value_if<F1, F2>(self, accept_if: F1, as_value: F2) -> Fallible<T>
-    where
-        F1: FnOnce(&VoltaError) -> bool,
-        F2: FnOnce() -> T,
-    {
-        match self {
-            Err(error) => {
-                if accept_if(&error) {
-                    Ok(as_value())
-                } else {
-                    Err(error)
-                }
-            }
-            Ok(v) => Ok(v),
-        }
-    }
-}
-
-impl<T> AcceptableErrorToDefault<T> for Fallible<T> where T: Default {}
 
 /// Trait providing the with_context method to easily convert any Result error into a VoltaError
 pub trait Context<T> {
--- a/crates/volta-core/src/error/mod.rs
+++ b/crates/volta-core/src/error/mod.rs
@@ -8,8 +8,6 @@ mod reporter;
 pub use kind::ErrorKind;
 pub use reporter::report_error;
 
-use crate::fs::is_not_found_error_kind;
-
 pub type Fallible<T> = Result<T, VoltaError>;
 
 /// Error type for Volta
@@ -48,25 +46,18 @@ impl VoltaError {
         &self.inner.kind
     }
 
-    /// TODO:
-    /// 1. find a good name
-    /// 2. determine if it can be removed
-    pub fn not_found_to_ok<T>(self, to: T) -> Fallible<T> {
+    /// Verifies if the inner source is `io::ErrorKind::NotFound`
+    pub fn is_not_found_error_kind(&self) -> bool {
         self.source()
             .and_then(|source| source.downcast_ref::<io::Error>())
             .and_then(|io_err| {
                 if io_err.kind() == io::ErrorKind::NotFound {
-                    Some(to)
+                    Some(true)
                 } else {
                     None
                 }
             })
-            .ok_or(self)
-    }
-
-    /// TODO: find a good name
-    pub fn is_io_not_found(&self) -> bool {
-        is_not_found_error_kind(self)
+            .unwrap_or(false)
     }
 }
 
@@ -102,7 +93,25 @@ where
     T: Default,
     Self: Sized,
 {
-    fn error_to_default_if<F>(self, accept: F) -> Fallible<T>
+    /// Transforms an `Fallible<T>::Err(x)` into `Fallible<T>::Ok(T::default())`
+    ///
+    ///# Example
+    ///```
+    ///use std::io;
+    ///use io::Error;
+    ///use volta_core::error::{Fallible, VoltaError, ErrorKind};
+    ///use crate::volta_core::error::AcceptableErrorToDefault;
+    ///
+    ///let not_found = Error::new(io::ErrorKind::NotFound, "file not found");
+    ///let volta_error = VoltaError::from_source(not_found, ErrorKind::BinaryExecError);
+    ///let fallible: Fallible<i32> = Err(volta_error);
+    ///if let Ok(zero) = fallible.accept_error_as_default_if(|e| e.is_not_found_error_kind()) {
+    ///    assert_eq!(zero, 0);
+    ///} else {
+    ///    assert!(false, "accept_error_as_value_if is not equal to 10");
+    ///}
+    ///```
+    fn accept_error_as_default_if<F>(self, accept: F) -> Fallible<T>
     where
         F: FnOnce(&VoltaError) -> bool,
     {
@@ -110,7 +119,25 @@ where
     }
 }
 
+/// Extension to `Fallible<T>` to transform a `VoltaError` into a value
 impl<T> AcceptableErrorToValue<T> for Fallible<T> {
+    ///
+    ///# Example
+    ///```
+    ///use std::io;
+    ///use io::Error;
+    ///use volta_core::error::{Fallible, VoltaError, ErrorKind};
+    ///use crate::volta_core::error::AcceptableErrorToValue;
+    ///
+    ///let not_found = Error::new(io::ErrorKind::NotFound, "file not found");
+    ///let volta_error = VoltaError::from_source(not_found, ErrorKind::BinaryExecError);
+    ///let fallible: Fallible<i32> = Err(volta_error);
+    ///if let Ok(ten) = fallible.accept_error_as_value_if(|e| e.is_not_found_error_kind(), || 10) {
+    ///    assert_eq!(ten, 10);
+    ///} else {
+    ///    assert!(false, "accept_error_as_value_if is not equal to 10");
+    ///}
+    ///```
     fn accept_error_as_value_if<F1, F2>(self, accept_if: F1, as_value: F2) -> Fallible<T>
     where
         F1: FnOnce(&VoltaError) -> bool,
--- a/crates/volta-core/src/error/mod.rs
+++ b/crates/volta-core/src/error/mod.rs
@@ -1,12 +1,14 @@
-use std::error::Error;
 use std::fmt;
 use std::process::exit;
+use std::{error::Error, io};
 
 mod kind;
 mod reporter;
 
 pub use kind::ErrorKind;
 pub use reporter::report_error;
+
+use crate::fs::is_not_found_error_kind;
 
 pub type Fallible<T> = Result<T, VoltaError>;
 
@@ -45,6 +47,24 @@ impl VoltaError {
     pub fn kind(&self) -> &ErrorKind {
         &self.inner.kind
     }
+
+    /// TODO: doc
+    pub fn not_found_to_ok<T>(self, to: T) -> Fallible<T> {
+        self.source()
+            .and_then(|source| source.downcast_ref::<io::Error>())
+            .and_then(|io_err| {
+                if io_err.kind() == io::ErrorKind::NotFound {
+                    Some(to)
+                } else {
+                    None
+                }
+            })
+            .ok_or(self)
+    }
+
+    pub fn is_io_not_found(&self) -> bool {
+        is_not_found_error_kind(self)
+    }
 }
 
 impl fmt::Display for VoltaError {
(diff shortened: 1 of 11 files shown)


crates/volta-core/src/error/mod.rs
```diff
@@ -45,6 +45,20 @@ impl VoltaError {
     pub fn kind(&self) -> &ErrorKind {
         &self.inner.kind
     }
+
+    /// Verifies if the inner source is `io::ErrorKind::NotFound`
+    pub fn is_not_found_error_kind(&self) -> bool {
```
charlespierce, line 50: I'm a little wary of having a method on `VoltaError` that is only relevant for one specific kind of one specific source error. I like the idea in general of being able to possibly downcast the source, but I wonder if we can make it more generic and then provide a helper function (possibly in `volta_core::fs` for the specific "not found" case?

Something like:

```rust
pub fn caused_by<E, F>(&self, predicate: F) -> bool
where
    E: Error + 'static,
    F: FnOnce(&E) -> bool,
{
    match self.source().and_then(|e| e.downcast_ref::<E>()) {
        Some(err) => predicate(err),
        None => false,
    }
}
```
and
```rust
pub fn not_found(err: &io::Error) -> bool {
    err.kind() == io::ErrorKind::NotFound
}
```

At which point this calling this function could be turned from: `err.is_not_found_error_kind()` to `err.caused_by(not_found)`, which maintains the expressiveness while allowing us to re-use it in other spots where we might want to look up the sub-error.

Edit: After looking at the scope of other changes, I'm not sure if this will still be needed (though it may be inside the `from_file_if_exists` methods, I'm not sure on that). But if it is, keeping things more generic would be useful I think.

crates/volta-core/src/error/mod.rs
```diff
@@ -67,6 +81,83 @@ impl From<ErrorKind> for VoltaError {
...
 }
 
+pub trait AcceptableErrorToValue<T> {
+    fn accept_error_as_value_if<F1, F2>(self, accept_if: F1, as_value: F2) -> Fallible<T>
+    where
+        F1: FnOnce(&VoltaError) -> bool,
+        F2: FnOnce() -> T;
+}
+
+pub trait AcceptableErrorToDefault<T>: AcceptableErrorToValue<T>
```
charlespierce, line 91: I haven't worked all the way through it, but I think a lot of this machinery could be handled with a helper function and the built-in `Result::or_else` method, which I think would simplify it a fair amount.

crates/volta-core/src/project/mod.rs
```diff
@@ -147,10 +147,17 @@ impl Project {
...
-            }
+
+            return match BinConfig::from_file(config_path) {
+                Err(error) => {
+                    if error.is_not_found_error_kind() {
+                        Ok(false)
+                    } else {
+                        return Err(error);
+                    }
+                }
```
charlespierce, line 158: This is a fair amount of extra boilerplate that is duplicated in a lot of spots. Perhaps we could create an extra method on `BinConfig` and `PackageConfig` that handles the not found error and returns a `Fallible<Option<ConfigType>>`? Then this section could be condensed to:

```rust
match BinConfig::from_file_if_exists(config_path)? {
    Some(config) => Ok(self.has_direct_dependency(&config.package)),
    None => Ok(false),
}
```

crates/volta-core/src/run/binary.rs
```diff
@@ -167,11 +167,15 @@ impl DefaultBinary {
...
-            Ok(None) // no config means the tool is not installed
+        match BinConfig::from_file(bin_config_file) {
+            Err(error) => {
+                if error.is_not_found_error_kind() {
+                    Ok(None)
+                } else {
+                    Err(error)
+                }
+            }
+            Ok(bin_config) => DefaultBinary::from_config(bin_config, session).map(Some),
```
charlespierce, line 178: Same here, this could be:

```rust
match BinConfig::from_file_if_exists(bin_config_file)? {
    Some(bin_config) => DefaultBinary::from_config(bin_config, session).map(Some),
    None => Ok(None),
}
```

crates/volta-core/src/tool/package/install.rs
```diff
@@ -183,19 +183,27 @@ fn read_bins(name: &str, version: &Version) -> Fallible<HashMap<String, String>>
...
-        if bin_config_file.exists() {
-            let bin_config = BinConfig::from_file(bin_config_file)?;
-            // if the bin was installed by the package that is currently being installed,
-            // that's ok - otherwise it's an error
-            if name != bin_config.package {
-                return Err(ErrorKind::BinaryAlreadyInstalled {
-                    bin_name: bin_name.clone(),
-                    existing_package: bin_config.package,
-                    new_package: name.to_string(),
+        match BinConfig::from_file(bin_config_file) {
```
charlespierce, line 186: Similarly here

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -132,25 +132,33 @@ pub fn uninstall(name: &str) -> Fallible<()> {
...
             }
-            true
+            Ok(true)
         } else {
-            false
+            Ok(false)
+        }
+    };
+
+    let package_found = match PackageConfig::from_file(&package_config_file) {
```
charlespierce, line 148: And here if we do the same to `PackageConfig`

crates/volta-core/src/tool/package_global/uninstall.rs
```diff
@@ -21,25 +22,30 @@ pub fn uninstall(name: &str) -> Fallible<()> {
 
     // If the package config file exists, use that to remove any installed bins and shims
     let package_config_file = home.default_package_config_file(name);
-    let package_found = if package_config_file.exists() {
-        let package_config = PackageConfig::from_file(&package_config_file)?;
 
-        for bin_name in package_config.bins {
-            remove_config_and_shim(&bin_name, name)?;
+    let package_found = match PackageConfig::from_file(&package_config_file) {
```
charlespierce, line 26: This I think would be simplified by a `PackageConfig::from_file_if_exists` as well.

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -182,20 +190,17 @@ fn remove_config_and_shim(bin_name: &str, pkg_name: &str) -> Fallible<()> {
...
+            if config.package == package {
+                return Some(config.name);
+            }
+        };
+        None
+    })
+    .with_context(|| ErrorKind::ReadBinConfigDirError {
+        dir: bin_config_dir.to_owned(),
+    })
+    .accept_error_as_default_if(|err| err.is_not_found_error_kind())
```
charlespierce, line 205: Since `dir_entry_match` actually returns an `io::Error`, I think this could be simplified to an `or_else` call on the original Result (_before_ calling `with_context`). We actually already have a (currently private) function in `volta_core::fs` named `ok_if_not_found`, which is set up to work for unit types only, but could be made generic over any type that implements `Default`.

crates/volta-core/src/tool/package_global/uninstall.rs
```diff
@@ -72,20 +78,18 @@ fn remove_config_and_shim(bin_name: &str, pkg_name: &str) -> Fallible<()> {
...
-    } else {
-        Ok(Vec::new())
-    }
+        }
+        None
+    })
+    .with_context(|| ErrorKind::ReadBinConfigDirError {
+        dir: bin_config_dir.to_owned(),
+    })
+    .accept_error_as_default_if(|e| e.is_not_found_error_kind())
```
charlespierce, line 94: Same comment here about `dir_entry_match` and `volta_core::fs::ok_if_not_found`

crates/volta-migrate/src/v1.rs
```diff
@@ -87,17 +87,18 @@ impl TryFrom<V0> for V1 {
...
 
             let old_shim_bin = new_home.root().join("shim");
-            if old_shim_bin.exists() {
-                remove_file(&old_shim_bin)
-                    .with_context(|| ErrorKind::DeleteFileError { file: old_shim_bin })?;
-            }
+
+            remove_file(&old_shim_bin)
+                .with_context(|| ErrorKind::DeleteFileError { file: old_shim_bin })
+                .accept_error_as_default_if(|err| err.is_not_found_error_kind())?;
```
charlespierce, line 101: We actually have a helper in `volta_core::fs` named `remove_file_if_exists`, which explicitly doesn't error if the file doesn't exist, so I think that's what we want here.

crates/volta-migrate/src/v2.rs
```diff
@@ -94,24 +94,31 @@ impl TryFrom<V1> for V2 {
...
+        }) {
+            Err(error) => {
+                if error.is_not_found_error_kind() {
+                    return Ok(());
+                } else {
+                    return Err(error);
+                }
+            }
+            Ok(json) => json,
+        };
```
charlespierce, line 109: I think this can be simplified a little by removing the call to `with_context` and instead making it part of the error branch:

```rust
let platform_json = match read_to_string(platform_file) {
    Err(error) => {
        if error.kind() == io::ErrorKind::NotFound {
            return Ok(());
        } else {
            return Err(VoltaError::from_source(error, ErrorKind::ReadPlatformError { file: platform_file.to_owned() }));
        }
    }
    Ok(json) => json,
}
```

crates/volta-migrate/src/v1.rs
```diff
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 use super::empty::Empty;
 use super::v0::V0;
 use log::debug;
-use volta_core::error::{Context, ErrorKind, Fallible, VoltaError};
+use volta_core::error::{AcceptableErrorToDefault, Context, ErrorKind, Fallible, VoltaError};
```
charlespierce, line 10: The `remove_file` calls are only in the `unix` code, so any imports should be marked as `#[cfg(unix)]` to avoid unused import Warnings on Windows.

charlespierce (changes requested):
Thanks for this @BProg! This looks like it cleans up all of the spots where we were doing the `exists()` + `open()` anti-pattern. My comments are mainly around reducing boilerplate and making things more generic.

In general, I think we should try to have the checks for `NotFound` as close as possible to the IO operation that actually caused them. That will help reduce the amount of times we need to check in separate locations for the same condition, and make it easier to maintain going forward.

crates/volta-migrate/src/v2.rs
```diff
@@ -79,11 +79,11 @@ impl TryFrom<V1> for V2 {
...
         let old_layout_file = old.home.layout_file();
-        if old_layout_file.exists() {
-            remove_file(old_layout_file).with_context(|| ErrorKind::DeleteFileError {
+        remove_file(old_layout_file)
+            .with_context(|| ErrorKind::DeleteFileError {
                 file: old_layout_file.to_owned(),
-            })?;
-        }
+            })
+            .accept_error_as_default_if(|err| err.is_not_found_error_kind())?;
```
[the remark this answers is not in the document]
BProg: Done

crates/volta-core/src/error/mod.rs
BProg: Done

crates/volta-core/src/project/mod.rs
BProg: Done

crates/volta-core/src/run/binary.rs
BProg: Done

crates/volta-core/src/tool/package/install.rs
BProg: Done

crates/volta-core/src/tool/package/mod.rs
BProg: Done

crates/volta-core/src/tool/package_global/uninstall.rs
BProg: Done

crates/volta-core/src/tool/package/mod.rs
BProg: Done

crates/volta-core/src/tool/package_global/uninstall.rs
BProg: Done

crates/volta-migrate/src/v1.rs
BProg: Done

crates/volta-migrate/src/v2.rs
BProg: Simplified by using `.or_else(ok_if_not_found)`

crates/volta-migrate/src/v1.rs
BProg: Done

crates/volta-core/src/tool/package/metadata.rs
```diff
@@ -260,6 +281,24 @@ impl BinConfig {
...
+                } else {
+                    Err(VoltaError::from_source(
+                        error,
+                        ErrorKind::ReadBinConfigError { file },
+                    ))
+                }
+            }
+            Ok(config_src) => RawBinConfig::from_json(config_src)?
+                .try_into()
+                .map(|config| Some(config)),
```
charlespierce, line 299: Clippy raises a warning here: You actually don't need the closure `|config| Some(config)`, you can instead do this with:

```suggestion
                .map(Some),
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -3,14 +3,17 @@ use std::path::{Path, PathBuf};
...
 use crate::inventory::package_available;
 use crate::layout::volta_home;
 use crate::session::Session;
 use crate::shim;
 use crate::style::{success_prefix, tool_version};
 use crate::sync::VoltaLock;
+use crate::{
+    error::{Context, ErrorKind, Fallible},
+    fs::ok_if_not_found,
+};
```
charlespierce, line 16: To keep the diffs simpler, we've generally tried to keep our use statements to single modules at a time. So instead of grouping the submodules together like this, we would have 2 separate lines:

```rust
use crate::error::{Context, ErrorKind, Fallible};
use crate::fs::ok_if_not_found;
```
(With the appropriate ordering provided by `cargo fmt`.

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -132,25 +135,27 @@ pub fn uninstall(name: &str) -> Fallible<()> {
...
 
-        for bin_name in package_config.bins {
-            remove_config_and_shim(&bin_name, name)?;
-        }
-
-        remove_file_if_exists(package_config_file)?;
-        true
-    } else {
-        // there is no package config - check for orphaned binaries
+    let remove_orphan_binaries_if_any = || -> Fallible<bool> {
```
charlespierce, line 139: Since we aren't re-using this function anywhere else, I think it would be slightly clearer if this block was included in the `match` arm below, instead of creating a closure and then calling it.

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -182,20 +187,17 @@ fn remove_config_and_shim(bin_name: &str, pkg_name: &str) -> Fallible<()> {
...
-        })
-        .with_context(|| ErrorKind::ReadBinConfigDirError {
-            dir: bin_config_dir.to_owned(),
-        })
-    } else {
-        Ok(vec![])
-    }
+    dir_entry_match(&bin_config_dir, |entry| {
+        let path = entry.path();
+        if let Ok(Some(config)) = BinConfig::from_file_if_exists(path) {
```
charlespierce, line 192: Very minor, but since we're only matching on `Ok(Some(_))` here, I think we can leave this line as it was (since it amounts to the same thing):

```suggestion
        if let Ok(config) = BinConfig::from_file(path) {
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -182,20 +187,17 @@ fn remove_config_and_shim(bin_name: &str, pkg_name: &str) -> Fallible<()> {
...
+            if config.package == package {
+                return Some(config.name);
+            }
+        };
+        None
+    })
+    .or_else(ok_if_not_found)
```
charlespierce, line 199: This is _so_ much cleaner than the `if exists() { } else { Ok(vec![]) }` blocks! 🎉

crates/volta-core/src/tool/package_global/metadata.rs
```diff
@@ -91,6 +115,29 @@ impl BinConfig {
...
+                        error,
+                        ErrorKind::ReadBinConfigError {
+                            file: file.as_ref().to_owned(),
+                        },
+                    ))
+                }
+            }
+            Ok(config) => serde_json::from_reader(config)
+                .with_context(|| ErrorKind::ParseBinConfigError)
+                .map(|config| Some(config)),
```
charlespierce, line 137: Same suggestion here about the unnecessary closure:

```suggestion
                .map(Some),
```

crates/volta-migrate/src/v2.rs
```diff
@@ -7,11 +7,14 @@ use super::v1::V1;
...
 use tempfile::tempdir_in;
-use volta_core::error::{Context, ErrorKind, Fallible, VoltaError};
 use volta_core::fs::{read_dir_eager, remove_dir_if_exists, rename};
 use volta_core::tool::load_default_npm_version;
 use volta_core::toolchain::serial::Platform;
 use volta_core::version::parse_version;
+use volta_core::{
+    error::{Context, ErrorKind, Fallible, VoltaError},
+    fs::ok_if_not_found,
+};
```
charlespierce, line 17: Same comment here about keeping the imports separated. It also looks like we actually have a line with `use volta_core::fs::{...}`, so the additional import can be added to that line instead of a separate one.

crates/volta-migrate/src/v1.rs
```diff
@@ -87,17 +87,18 @@ impl TryFrom<V0> for V1 {
 
             debug!("Removing old Volta binaries");
             let old_volta_bin = new_home.root().join("volta");
-            if old_volta_bin.exists() {
-                remove_file(&old_volta_bin).with_context(|| ErrorKind::DeleteFileError {
+
+            remove_file(&old_volta_bin)
+                .or_else(ok_if_not_found)
```
charlespierce, line 92: Instead of directly doing `remove_file().or_else()`, there actually is a helper in `volta_core::fs` named `remove_file_if_exists`, so we should re-use that (it internally does the same thing since it's a common pattern).

crates/volta-migrate/src/v1.rs
```diff
@@ -87,17 +87,18 @@ impl TryFrom<V0> for V1 {
...
-            }
 
             let old_shim_bin = new_home.root().join("shim");
-            if old_shim_bin.exists() {
-                remove_file(&old_shim_bin)
-                    .with_context(|| ErrorKind::DeleteFileError { file: old_shim_bin })?;
-            }
+
+            remove_file(&old_shim_bin)
+                .or_else(ok_if_not_found)
```
charlespierce, line 100: Same here about using `remove_file_if_exists` instead of `remove_file().or_else()`

crates/volta-migrate/src/v2.rs
```diff
@@ -79,12 +82,11 @@ impl TryFrom<V1> for V2 {
 
         // Remove the V1 layout file, since we're now on V2 (do this after writing the V2 so that we know the migration succeeded)
         let old_layout_file = old.home.layout_file();
-        if old_layout_file.exists() {
-            remove_file(old_layout_file).with_context(|| ErrorKind::DeleteFileError {
+        remove_file(old_layout_file)
+            .or_else(ok_if_not_found)
```
charlespierce, line 86: Here as well with `remove_file_if_exists`

crates/volta-migrate/src/v2.rs
```diff
@@ -94,24 +96,23 @@ impl TryFrom<V1> for V2 {
...
-                            || ErrorKind::WritePlatformError {
-                                file: platform_file.to_owned(),
-                            },
-                        )?;
-                    }
+    let platform_json = read_to_string(platform_file)
+        .or_else(ok_if_not_found)
+        .with_context(|| ErrorKind::ReadPlatformError {
+            file: platform_file.to_owned(),
+        })?;
```
charlespierce, line 103: This happens to work because the following line `Platform::from_json` handles an empty string, but I'm not sure if we should be relying on that behavior. Instead, I think this can be a little more explicit about handling the `NotFound` error (similar to the implementation of `from_file_if_exists` above):

```rust
let platform_json = match read_to_string(platform_file) {
    Ok(json) => json,
    Err(error) => if error.kind() == io::ErrorKind::NotFound {
        return Ok(());
    } else {
        return Err(VoltaError::from_source(error, ErrorKind::ReadPlatformError { ... }));
    }
};
```

charlespierce (changes requested):
This looks _so_ much simpler than the `path.exists()` checks, thanks @BProg! The remaining comments are primarily around style and cleaning up some clippy warnings, nothing major.

charlespierce (approved):
This is awesome, thanks for the help @BProg! :zap: :tada:

PR #835, opened 2020-09-23
[Large] Add support for installing multiple packages in one command

charlespierce:
Info
-----
* The implementation of direct package installs in #829 (and the follow-ups) only supported the _first_ package supplied to the install command.
* If the user ran `npm install --global typescript ember-cli`, which is a valid `npm` command, only `typescript` would get installed.
* This PR updates the install logic to support multiple installs with a single command.

Changes
-----
* Added a `Multiple` variant to `Executor` that handles running a collection of `Executors`
* Moved the argument parsing logic out of `npm.rs` and `yarn.rs` and into `parser.rs`
* Updated the parser logic to support multiple arguments:
    * When multiple packages are installed at the same time, each install is run separately, so that we can keep each package sandboxed.
    * Each install uses the flags that were passed to the command.
    * If there is no default platform (i.e. the user has never run `volta install node`), then no interception is done and we rely on the pass-through logic.
    * Uninstalls are similarly run separately (but all uninstalls go through the internal logic, so there is less coordination necessary there).

Tested
-----
* Moved parsing tests into the `parser` module and added tests around the multiple install behavior.
* Updated acceptance tests to test the multiple logic.
* Modified smoke tests for direct installs / uninstalls to cover the multiple package case.

Notes
-----
* The _big_ caveat to this behavior is that command-line flags with values _must_ be passed as `--prefix=/usr/local/bin`, and not as `--prefix /usr/local/bin`. Unfortunately, without duplicating the understanding of which flags for each package manager are expected to have values, it's not possible to handle the general case, since `--prefix /usr/local/bin` would represent a single flag, while `--global typescript` would represent a flag and a tool. I created a follow-up issue to document this limitation and hopefully fix it in the future (#836).


crates/volta-core/src/run_package_global/parser.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// If there are multiple packages specified to install, then they will be broken out into
+    /// individual commands and run separately. That allows us to keep Volta's sandboxing for each
+    /// package while still supporting the ability to install multiple packages at once.
+    pub fn executor(self, default_platform: &PlatformSpec) -> Fallible<Executor> {
+        let mut executors = Vec::with_capacity(self.tools.len());
+
+        for tool in self.tools {
+            match Spec::try_from_str(&tool.to_string_lossy())? {
+                Spec::Package(name, _) => {
+                    let platform = default_platform.as_default();
```
chriskrycho, line 133: This is a little thing that we don't need to tackle now, but needing to do this seems like a signal that our handling of platforms can still keep improving such that we don't end up writing something like this. I wonder if maybe just naming it `platform` would help, so we end up calling `platform.as_default()` instead?

chriskrycho (approved):
⚡

PR #838, opened 2020-09-23
[Small] Add VOLTA_UNSAFE_GLOBAL env variable to new package install

charlespierce:
Info
-----
* Previously, we used the (undocumented) environment variable `VOLTA_UNSAFE_GLOBAL` to disable the global install interception.
* While our new logic allows global installs, it still makes modifications to the process, so we should keep the escape valve to run the bare command.

Changes
-----
* Updated the `CommandArg::for_npm` and `CommandArg::for_yarn` to short-circuit to `NotGlobal` (No interception) if the `VOLTA_UNSAFE_GLOBAL` environment variable is set.

Tested
-----
* Added acceptance tests to confirm that global installs are not modified when the environment variable is set.

### DIFF
--- a/crates/volta-core/src/run_package_global/mod.rs
+++ b/crates/volta-core/src/run_package_global/mod.rs
@@ -1,6 +1,5 @@
 use std::collections::HashMap;
-use std::env;
-use std::env::ArgsOs;
+use std::env::{self, ArgsOs};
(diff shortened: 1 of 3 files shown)


chriskrycho (approved):
⚡

PR #839, opened 2020-09-23
[Small] Remove errors that aren't being used

charlespierce:
Info
-----
* With the `package-global` changes making significant changes to the package install flow, several errors are no longer being used.
* Additionally, there were some `ErrorKind` variants left around that weren't being used at all (even before the changes).

Changes
-----
* Marked all errors that are not used with the feature flag as `#[cfg(not(feature = "package-global"))]` so they won't be included.
* Removed several errors that weren't in use anywhere in the app.

Tested
-----
* Confirmed that Volta compiles and the tests pass both with and without the feature flag.


chriskrycho (approved):
⚡

PR #840, opened 2020-09-23
[Tiny] Ensure a VoltaLock is acquired when installing packages directly

charlespierce:
Info
-----
* When installing a package with `npm i -g` or `yarn global add`, we are still modifying the Volta directory, so we still should have a lock on the process.

Changes
-----
* Added a call to `VoltaLock::acquire()` in the `execute` method of the `PackageInstallCommand` executor.
* Removed an erroneous `?` on the call to `VoltaLock::acquire()` in the `Package` implementation of `Tool`


chriskrycho (approved):
⚡

PR #842, opened 2020-09-24
[Medium] Handle global installs from alternative sources

charlespierce:
Info
-----
* Both npm and Yarn support installing packages from alternative sources, such as local tarballs and git coordinates.
* Since we now support installing tools directly with `npm i -g` and `yarn global add`, we should also support installing globals from tarballs and git coordinates.

Changes
-----
* Updated the `DirectInstall` helper to not need a package name, instead reading the package name from the installed directory.
    * For `npm`, the top-level `node_modules` directory is not hoisted, so it will only contain the newly-installed package.
    * For `yarn`, all dependencies are hoisted, but they provide a "global" `package.json` file that we can read the `dependencies` of to get the name of the newly-installed package.
* Updated the `InstallArgs` implementation to handle the case where parsing an argument into a `tool::Spec` fails. In that case, we assume it is an alternate package source and treat it like a package install.

Tested
-----
* Local testing to confirm that installing packages from tarballs works as expected.
* Added a fixture package and updated the direct install smoke tests to also test installing from a tarball.


chriskrycho (approved):
MOAR ⚡

PR #844, opened 2020-09-25
[Medium] Allow global binaries to access global libraries

charlespierce:
Info
-----
* As described in [the RFC](https://github.com/volta-cli/rfcs/pull/45), we want to provide a way for global packages to "see" other global packages when running a `require` statement, to match up with the standard global install behavior.
* To do this, we create a shared directory with symlinks to each of the global packages and add it to the `NODE_PATH` environment variable when executing a global binary.

Changes
-----
* Added the `shared_lib_root` and `shared_lib_dir` to the V3 layout (no change to the migration is necessary as the directory is already created by the `create()` method).
* Updated the global install process (for both `volta install` and `npm i -g`) to create a symlink from the package directory into the `shared_lib_root`
* Updated the global uninstall process to clean up the symlink.
* Modified the execution of a global package to append the `shared_lib_root` (which will have sub-directory symlinks for each globally installed package) to the `NODE_PATH` environment variable, making sure that the global packages can be found by the Node module resolution algorithm.

Tested
-----
* Tested locally that `ts-node` works and can find a globally-installed peer dependency on `typescript`.
* Tested locally and confirmed that `yo` can find Yeoman generators installed globally as well.

### DIFF
--- a/crates/volta-core/src/tool/package_global/mod.rs
+++ b/crates/volta-core/src/tool/package_global/mod.rs
@@ -4,7 +4,7 @@ use std::process::Command;
 
 use super::Tool;
 use crate::error::{Context, ErrorKind, Fallible};
-use crate::fs::{create_staging_dir, remove_dir_if_exists, rename};
+use crate::fs::{create_staging_dir, remove_dir_if_exists, rename, symlink_dir};
 use crate::layout::volta_home;
 use crate::platform::{Image, PlatformSpec};
 use crate::session::Session;
@@ -56,6 +56,7 @@ impl Package {
             configure::parse_manifest(&self.name, self.staging.path().to_owned(), manager)?;
 
         persist_install(&self.name, &self.version, self.staging.path())?;
+        link_package_to_shared_dir(&self.name, manager)?;
         configure::write_config_and_shims(&self.name, &manifest, image, manager)?;
 
         Ok(manifest)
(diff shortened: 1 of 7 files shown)


crates/volta-core/src/tool/package_global/mod.rs
```diff
@@ -155,8 +165,27 @@ where
     rename(staging_dir, &package_dir).with_context(|| ErrorKind::SetupToolImageError {
         tool: package_name.into(),
         version: package_version.to_string(),
-        dir: package_dir,
+        dir: package_dir.clone(),
```
chriskrycho, line 168: Looks extraneous!

chriskrycho (approved):
Seems good, one tiny tweak for an unnecessary clone!

crates/volta-core/src/tool/package_global/mod.rs
charlespierce: Fixed!

insidewhy:
For me `typescript-language-server` can't find `typescript` with volta `1.1.0`, I've read through several issues to try to get this to work and still can't find it out.

insidewhy:
For now I just hacked it by doing this:

```bash
cd $(dirname $(volta which typescript-language-server))/../lib/node_modules/typescript-language-server/node_modules
ln -s ../../../../../typescript/lib/node_modules/typescript
```

PR #845, opened 2020-09-26
[Small] Leave shim directory on PATH to allow bins to call other bins

charlespierce:
Info
-----
* In a non-Volta environment, global bins are put on the PATH and can freely call one another.
* Volta should allow that behavior as well, since there are cases where it is needed (multiple related bins delivered as a single package, for example).
* To do that, we need to leave the shim directory on the PATH when executing a command.
* However, since the Volta shims for global packages are located in the same directory as the shims for node, npm, yarn, etc., we need to make sure that a bad state doesn't cause an infinite recursion.
    * The built-in shims (Node, npm, npx, and Yarn) each call their command by putting the appropriate directory on the PATH and then calling the bare command.
    * This is in contrast to the binary shims, which call their command directly with the path to the executable.
    * For the former case, if something is misconfigured, it could happen that the shim directory is the earliest valid location on the PATH with that command, which would cause the shim to call itself recursively over and over.
* To avoid infinite recursion, we set an environment variable `_VOLTA_TOOL_RECURSION` when executing a command. Then, when one of the built-in shims tries to execute, iff that environment variable is already set, it will instead use the pass-through logic (which _does_ remove the shim directory from the PATH), breaking the recursive chain.
* One caveat is that when we execute `volta run <command>`, we explicitly remove the environment variable, because it is perfectly valid for e.g. a Node script to call `volta run npm` to force re-evaluation of the context (maybe in a different directory). Since `volta run` can never call itself directly, there's no risk of an infinite recursion.

Changes
-----
* Updated `Image::path` to no longer remove the shim directory from the PATH.
* Set the recursion environment variable when executing a shim command (either directly or a global install).
* Explicitly un-set the environment variable when executing a `volta run` command.
* Updated the logic on the built-in shims to detect the environment variable and skip platform detection if it is set.

Tested
-----
* Updated the `Image::path` tests to account for the new behavior.
* Locally tested that two separate globally installed binaries can see one another.
* Also tested that directly breaking the environment doesn't result in an infinite recursion.

Notes
-----
* The Binaries ignore the recursion environment variable for two reasons:
    1. Binaries are called directly, so it's not possible for executing the shim to immediately re-invoke itself.
    2. Different binaries may have different platforms, so we need to make sure to re-evaluate when they run.

### DIFF
--- a/crates/volta-core/src/platform/image.rs
+++ b/crates/volta-core/src/platform/image.rs
@@ -3,7 +3,9 @@ use std::path::PathBuf;
 
 use super::{build_path_error, Sourced};
 use crate::error::{Context, Fallible};
-use crate::layout::{env_paths, volta_home};
+#[cfg(not(feature = "package-global"))]
+use crate::layout::env_paths;
+use crate::layout::volta_home;
 use crate::tool::load_default_npm_version;
 use semver::Version;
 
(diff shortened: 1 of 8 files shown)


chriskrycho (approved):
⚡

PR #848, opened 2020-10-02
Fully enable feature flag and remove gated code

charlespierce:
Info
-----
* With the feature work complete on the revamped package installs, we can remove the feature flag gate and make the new code the default.

Note
-----
* To keep the diff readable, the `tool::package` and `run` modules are left at their renamed locations (`package_global` and `run_package_global`, respectively). In a follow-on PR, those directories will be renamed and the `#[path]` attributes will be removed.


chriskrycho (approved):
⚡ So much red! It's beautiful!

PR #849, opened 2020-10-02
Move modules to their conventional locations

charlespierce:
Follow-up to #848 - Move the new package global modules (`run` and `tool::package`) to their conventional locations and remove the need for `#[path]` attributes.

This is separated to make the diff easier to parse, since the previous PR removes the directories that were at those locations.

To see only the changes from this PR (much nicer than the combined), use [this link](https://github.com/volta-cli/volta/pull/849/commits/6507cbc42276a4567bbc2a8d16b04aa7227d419b)


chriskrycho (approved):
:shipit: ⚡

PR #851, opened 2020-10-12
Move back to attohttpc, now that it supports proxies

charlespierce:
Info
-----
* In #809, we switched to `reqwest` for handling HTTP requests to support proxies via `HTTP_PROXY` and `HTTPS_PROXY` environment variables.
* However, that change came with an increase in build time and binary size, since `reqwest` is a much larger library than `attohttpc`.
* Additionally, `reqwest` currently doesn't support the `NO_PROXY` environment variable for carving out exceptions to the proxy environment.
* `attohttpc` recently (with version `0.16.0`) added support for proxies, including the `NO_PROXY` environment variable.

Changes
-----
* Switched back to using `attohttpc`, but requiring version 0.16.0.
* This is mostly a revert of #809, accounting for merge conflicts.

Tested
-----
* All existing automated tests work.
* Local testing to confirm that proxy environment variables are supported.
* Local testing to confirm that `NO_PROXY` is respected as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,12 +38,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "archive"
 version = "0.1.0"
 dependencies = [
+ "attohttpc 0.16.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


chriskrycho (approved):
Nice reversion, quick turnaround I like it!

PR #854, opened 2020-10-20
v0.9.0

### LINKED ISSUE #427: Support `volta install`ing a tarball
There are times (like when you want to test a prebuilt `npm pack` result) that you need to install a global package but not from the registry.

Without `notion` you would do `npm install --global ./path/to/tarball.tgz`, but `notion` intercepts the `--global` invocation errors with:

```
Notion error: Global package installs are not recommended
```

Since we are _ultimately_ using `npm install` to do the heavy lifting here anyways, and `npm` supports:

```
npm install --global package-name@./some/path/to/tarball.tgz
```

We should be able to make something like this work (based on the work in https://github.com/notion-cli/notion/pull/383):

```
notion install ember-cli@./ember-cli-3.10.0.tgz
```

However, at the moment ^ causes the following error:

```
error: Could not parse version "ember-cli-3.10.0.tgz"
```

---

Reported by @turbo87 in discord

### LINKED ISSUE #472: Support `volta install <github-name>/<github-repo>`
When I am trying to test different versions of a tool from github I am unable to install a global package from there with either npm or volta.

### LINKED ISSUE #492: Leave Shims on the PATH when executing a tool
Currently, when we build the PATH to execute a tool, we remove the `~/.volta/bin` shim directory from the path, then prepend the relevant `node` and `yarn` directories. This allows a command like `node` to run and pick up the correct version of the `node` executable.

However, removing the shim directory means that any tool that is executed is not able to access other tools (except for `node`, `npm`, and `yarn` that are added to the path). This causes issues when two tools are designed to interoperate, such as `bower` and `bower-away`. The tools are both installed "globally", but when running one, you can't access the other.

Instead, we should leave the shim directory on the path, but make sure that the `node` and `yarn` directories are added _first_. This will allow a command like `node index.js` to still resolve to the correct version of Node (and not the shim), but will also let any tool see any other global tool that is installed (through the shim).

This has some risks: If we call `node` and for some reason the `node` executable isn't where we expect it, the call could come back to the shim, which will then try again, and again, until the OS kills the recursion. This shouldn't happen unless someone manually deletes the file, and if it does, it's a bug, but we might want to investigate if it's possible to detect a recursive call like that and break out early.

Another issue with tools calling each other is they'll each be adding their platform to the path, so it's possible that we wind up creating a merged platform that we didn't expect, or running into size limits on the length of the PATH, but I think those problems are fairly low-impact.

@dherman @chriskrycho @rwjblue Thoughts?

### LINKED ISSUE #540: Investigate using local `.npmrc` settings for global installs

### LINKED ISSUE #555: Support for non-binary global installs
Volta throws an error `Volta error: Global package installs are not supported.` on adding a global yarn/npm package. I understand the reasoning for this for *projects* but a valid use case of global installs are scripts to be run locally outside a project directory, and volta fundamentally breaks this.

I've seen `VOLTA_UNSAFE_GLOBAL` env posted, however this still doesn't allow requiring the package globally, so it doesn't really serve as a workaround. Is there any other recourse?

### LINKED ISSUE #605: Yarn / NPM installing incorrect binary for gRPC
I'm running into an issue where `npm` seems to be building and installing incorrect versions of binary modules.  Specifically it's the gRPC binary.  I've tried `npm rebuild` but it results in the same error (see below)

OS: macOS Catalina 10.15.1
Volta: 0.6.6
Node: v12.13.0
NPM: 6.12.0
Yarn: 1.19.1

I've ensured that all `yarn`, `node` and `npm` are executing from the proper `.vola` bin link.  I've also attempted to execute them directly with no effect on the outcome.  

Anyone have any ideas? 

```sh
0|dda:crypt:grpc  | Error: Failed to load gRPC binary module because it was not installed for the current system
0|dda:crypt:grpc  | Expected directory: node-v79-darwin-x64-unknown
0|dda:crypt:grpc  | Found: [node-v72-darwin-x64-unknown]
0|dda:crypt:grpc  | This problem can often be fixed by running "npm rebuild" on the current system
0|dda:crypt:grpc  | Original error: Cannot find module '/Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/src/node/extension_binary/node-v79-darwin-x64-unknown/grpc_node.node'
0|dda:crypt:grpc  | Require stack:
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/src/grpc_extension.js
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/src/client_interceptors.js
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/src/client.js
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/index.js
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/mali/lib/app.js
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/src/main.js
0|dda:crypt:grpc  | - /Users/nathansculli/src/cardcorp/dda/containers/crypt/bin/grpc-server.js
0|dda:crypt:grpc  |     at Object.<anonymous> (/Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/src/grpc_extension.js:53:17)
0|dda:crypt:grpc  |     at Module._compile (internal/modules/cjs/loader.js:1063:30)
0|dda:crypt:grpc  |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1103:10)
0|dda:crypt:grpc  |     at Module.load (internal/modules/cjs/loader.js:914:32)
0|dda:crypt:grpc  |     at Function.Module._load (internal/modules/cjs/loader.js:822:14)
0|dda:crypt:grpc  |     at Module.require (internal/modules/cjs/loader.js:956:19)
0|dda:crypt:grpc  |     at Module.Hook._require.Module.require (/Users/nathansculli/.volta/tools/image/packages/pm2/4.1.2/node_modules/require-in-the-middle/index.js:70:39)
0|dda:crypt:grpc  |     at require (internal/modules/cjs/helpers.js:74:18)
0|dda:crypt:grpc  |     at Object.<anonymous> (/Users/nathansculli/src/cardcorp/dda/containers/crypt/node_modules/grpc/src/client_interceptors.js:144:12)
0|dda:crypt:grpc  |     at Module._compile (internal/modules/cjs/loader.js:1063:30) {
0|dda:crypt:grpc  |   code: 'MODULE_NOT_FOUND'
0|dda:crypt:grpc  | }
```

### LINKED ISSUE #615: Volta error: Could not determine path to executable
Using Volta to add the NPM package `empty-trash-cli`, this is my console output

```bash
alexdiliberto@AlexBookPro: ~/.volta/bin
$ volta install empty-trash-cli
Package empty-trash-cli@1.0.1 is already installed

alexdiliberto@AlexBookPro: ~/.volta/bin
$ which -a empty-trash 
/Users/alexdiliberto/.volta/bin/empty-trash

alexdiliberto@AlexBookPro: ~/.volta/bin
$ VOLTA_LOGLEVEL=debug RUST_BACKTRACE=full empty-trash 
Volta error: Could not determine path to executable 'empty-trash'

Please rerun the command that triggered this error with the environment
variables `VOLTA_LOGLEVEL` set to `debug` and `RUST_BACKTRACE` set to `full`, and open
an issue at https://github.com/volta-cli/volta/issues with the details!
[verbose] Error cause: No such file or directory (os error 2)

stack backtrace:
   0: backtrace::backtrace::trace::haa64515333367434 (0x1022ac55c)
   1: backtrace::capture::Backtrace::new_unresolved::hed822f3deb5bab22 (0x1022ab848)
   2: failure::backtrace::internal::InternalBacktrace::new::he4e485359fae4acb (0x1022aaf85)
   3: <failure::backtrace::Backtrace as core::default::Default>::default::h047799f9c4ee0416 (0x1022ab185)
   4: <E as volta_fail::FailExt>::with_context::h5ddf5cb489fd6bf0 (0x101f6ea2d)
   5: volta_core::tool::package::bin_full_path::hc23047158f24eb23 (0x101f4be4b)
   6: volta_core::run::binary::DefaultBinary::from_config::h2f62779c68c56def (0x101f51b68)
   7: volta_core::run::binary::DefaultBinary::from_name::h5f4673d41c3767e8 (0x101f522bf)
   8: volta_core::run::execute_tool::h0d31734251e3d5e5 (0x101f52f5b)
   9: shim::execute::run_shim::h8d9fb7c8ec58508f (0x101f03ef3)
  10: std::rt::lang_start::{{closure}}::h1b9027d55f6f07ad (0x101f02e86)
  11: std::panicking::try::do_call::ha1a4587e2a4eb439 (0x1022c8428)
  12: ___rust_maybe_catch_panic (0x1022d0e0f)
  13: std::rt::lang_start_internal::hcf96e32a124891dc (0x1022c8ece)
  14: _main (0x101f02b39)
Error details written to /Users/alexdiliberto/.volta/log/volta-error-2019-11-29_14_38_46.478.log

```

### LINKED ISSUE #622: Ignore devDependencies in `volta install`
[As of now](https://github.com/prisma/prisma2/issues/1054), Volta looks into the `devDependencies` of a package, that is being installed through `volta install`.

However, by definition, these `devDependencies` don't need to be installed for a library, so it would be great, if volta would ignore these.

This would fix the situation, that a version scheme is being used, which Volta doesn't know, in the linked issue, it's the `workspace:` version protocol, supported by yarn and pnpm.

It's a separate issue, that this `workspace:` protocol shouldn't even end up in the `devDependencies`, which is being tracked here: https://github.com/pnpm/pnpm/pull/2203

### LINKED ISSUE #631: Cutomize the pinned version node used by a package binary
When you use `volta install` to install a package binary from NPM, volta chooses a node version to pin to that package using the following rules (from the docs):

> If the package has engines specified in package.json, use the latest LTS version of Node that meets the requirements
> If no LTS versions meet the requirements then use the latest overall version that satisfies engines
> If engines isn’t available, use the most recent LTS version of Node

This often results in a good choice. However, sometimes you need to pick a different version than volta would have picked to workaround bugs. For example, if a dependency of the binary is broken under latest node, there is currently no supported way to downgrade.

A current workaround is to manually edit json files under `~/.volta/tools/user/bins` and `~/.volta/tools/user/packages`.

A related issue is that the only supported way to upgrade the pinned node version is to uninstall and reinstall the package binary. A solution to the downgrade problem could probably be designed to also offer upgrades, without needing to reinstall the whole package binary.

### LINKED ISSUE #652: ts-node peerDependancy not found
Hi, 

I was in the process of transferring from nvm to volta (Windows & Linux) when I hit a missing feature (?). Apparently even though I installed `typescript` via volta `ts-node` was not able to resolve it.

I also tried to work around by installing a global npm package, which volta rightfully prevented 😅  

I think this is the case because typescript is not a dependency but a peerDependancy of `ts-node`
 
WindowsTerminal_2020-01-25_02-29-21

### LINKED ISSUE #681: Can't install windows-build-tools
Volta has generally been working well for me, but I recently tried using it to build on the VS Code repo. 

The Visual Studio Code repo apparently has some fancy stuff going on that expects Windows users to either have VS2017 installed, or [windows-build-tools](https://www.npmjs.com/package/windows-build-tools); however I think that because this package just exists to install a bunch of Windows executables globally, Volta isn't happy with it.

```
> volta install --verbose windows-build-tools
[verbose] Found project manifest in 'C:\Users\Daniel\vscode'
[verbose] Found default configuration at 'C:\Users\Daniel\AppData\Local\Volta\tools\user\platform.json'
[verbose] Using npm@6.13.4 from default configuration
[verbose] Running command: `"cmd.exe" "/C" "npm" "view" "--json" "windows-build-tools@latest"`
[verbose] [parsed package metadata (single)]
NpmViewData { name: "windows-build-tools", version: Version { major: 5, minor: 2, patch: 2, pre: [], build: [] }, dist: RawDistInfo { shasum: "1aabd53e40fb0a34cbd4b1023220573fab4dc772", tarball: "https://registry.npmjs.org/windows-build-tools/-/windows-build-tools-5.2.2.tgz" }, dist_tags: {"alpha": Version { major: 4, minor: 0, patch: 0, pre: [AlphaNumeric("alpha"), Numeric(13)], build: [] }, 
"latest": Version { major: 5, minor: 2, patch: 2, pre: [], build: [] }, "beta": Version { major: 3, minor: 0, patch: 0, pre: [AlphaNumeric("beta"), Numeric(3)], build: [] }} }
[verbose] Found windows-build-tools latest version (5.2.2) from https://registry.npmjs.org/windows-build-tools/-/windows-build-tools-5.2.2.tgz
[verbose] windows-build-tools@5.2.2 has already been fetched, skipping download
error: Package has no executables to install.
```

I'm not familiar enough with all of this, but I guess my questions are:

* Is WBT fundamentally incompatible with Volta?
* If it's compatible, does WBT just have to specify global executables that it will place in the system?
* Or if it's none of these, is there something Volta can do?

### LINKED ISSUE #762: Cannot find module when sub-package depends on the root package
Due to our file structure not exactly matching what Node expects (i.e. we don't have `node_modules` and we have an extra `version` directory), sub packages within an installed binary can't `require` anything from the main package. This causes problems in `@vue/cli`, as running the command `vue ui` results in `@vue/cli-ui` (a dependency) trying to `require('@vue/cli/lib/util/rcPath')` (a module in the root package), which fails because Node can't find the root package.

We should come up with a directory structure that adheres to Node's resolution algorithm or otherwise come up with a way of installing tools so that the root package can be accessed from sub-packages.

### LINKED ISSUE #806: https://nodejs.org down, volta cannot install global npm packages
```
volta install helman
```

```
⠠ Fetching public registry: https://nodejs.org/dist/index.json
error: Could not download Node version registry
from https://nodejs.org/dist/index.json

Please verify your internet connection.
Error details written to /home/izelnakri/.volta/log/volta-error-2020-08-30_13_21_16.019.log
```

This is happening because https://nodejs.org/ is currently down, rare situation but it happened today :) Global install of npm packages shouldnt depend on a nodejs.org network call.

charlespierce:
Release Volta 0.9.0

Cleans up a bunch of issues related to how global packages work:

Closes #806 
Closes #762 
Closes #681 
Closes #652 
Closes #631 
Closes #622 
Closes #615 
Closes #605 
Closes #555 
Closes #540 
Closes #492 
Closes #472 
Closes #427

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -3,6 +3,7 @@
 - Support Proxies through environment variables (#809, #851)
 - Avoid unnecessary `exists` calls for files (#834)
 - Rework package installs to allow for directly calling package manager (#848, #849)
+- **Breaking Change**: Remove support for `packages` hooks (#817)
 
 # Version 0.8.7
 
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,9 @@
+# Version 0.9.0
+
+- Support Proxies through environment variables (#809, #851)
+- Avoid unnecessary `exists` calls for files (#834)
+- Rework package installs to allow for directly calling package manager (#848, #849)
+
 # Version 0.8.7
 
 - Support fetching older versions of Yarn (#771)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,408 +4,457 @@
 name = "adler32"
 version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7e522997b529f05601e05166c07ed17789691f562762c7f3b987263d2dedee5c"
 
 [[package]]
 name = "aho-corasick"
 version = "0.6.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "81ce3d38065e618af2d7b77e10c5ad9a069859b4be3c2250f674af3840d9c8a5"
 dependencies = [
- "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
 name = "aho-corasick"
 version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8716408b8bc624ed7f65d223ddb9ac2d044c0547b6fa4b0d554f3a9540496ada"
 dependencies = [
- "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
 name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
 dependencies = [
- "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.9",
 ]
 
 [[package]]
 name = "approx"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08abcc3b4e9339e33a3d0a5ed15d84a687350c05689d825e0f6655eef9e76a94"
 
 [[package]]
 name = "archive"
 version = "0.1.0"
 dependencies = [
- "attohttpc 0.16.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "flate2 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "fs-utils 0.1.0",
- "hyperx 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "progress-read 0.1.0",
- "tar 0.4.26 (registry+https://github.com/rust-lang/crates.io-index)",
- "tee 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "thiserror 1.0.16 (registry+https://github.com/rust-lang/crates.io-index)",
- "verbatim 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "zip 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "attohttpc",
+ "cfg-if",
+ "flate2",
+ "fs-utils",
+ "hyperx",
+ "progress-read",
+ "tar",
+ "tee",
+ "thiserror",
+ "verbatim",
+ "zip",
 ]
 
 [[package]]
 name = "argon2rs"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f67b0b6a86dae6e67ff4ca2b6201396074996379fba2b92ff649126f37cb392"
 dependencies = [
- "blake2-rfc 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)",
- "scoped_threadpool 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "blake2-rfc",
+ "scoped_threadpool",
 ]
 
 [[package]]
 name = "arrayvec"
 version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "92c7fb76bc8826a8b33b4ee5bb07a247a81e76764ab4d55e8f73e3a4d8808c71"
 dependencies = [
- "nodrop 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "nodrop",
 ]
 
 [[package]]
 name = "attohttpc"
 version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1082810677916862c7704351dfe4696a837aaf34da0dd6431abc60783e71ee8f"
 dependencies = [
- "flate2 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "http 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "native-tls 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "openssl 0.10.30 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.92 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
- "url 2.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "wildmatch 1.0.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "flate2",
+ "http",
+ "log",
+ "native-tls",
+ "openssl",
+ "serde",
+ "serde_json",
+ "url",
+ "wildmatch",
 ]
 
 [[package]]
 name = "atty"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9a7d5b8723950951411ee34d271d99dddcc2035a16ab25310ea2c8cfd4369652"
 dependencies = [
- "libc 0.2.73 (registry+https://github.com/rust-lang/crates.io-index)",
- "termion 1.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc",
+ "termion",
+ "winapi 0.3.9",
 ]
 
 [[package]]
 name = "autocfg"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


RELEASES.md
```diff
@@ -1,3 +1,9 @@
+# Version 0.9.0
+
+- Support Proxies through environment variables (#809, #851)
+- Avoid unnecessary `exists` calls for files (#834)
+- Rework package installs to allow for directly calling package manager (#848, #849)
+
```
chriskrycho, line 6: If I understood the changes and read [the updated docs](https://github.com/volta-cli/website/pull/81/files#diff-c1a34c7015d9c7909962653b885f64b8458a3c9dcae05df63d7aa4ef3e2403f3R43) correctly, the changes to hooks *are* breaking changes and we should make a point to call them out here!

chriskrycho (changes requested):
This is great! I caught one thing we should call out, but otherwise let's do it!

RELEASES.md
charlespierce: Good call! I included it in the blog post, but not here. I'll add that!

chriskrycho (approved):

Issue #803, opened 2020-08-28
release tag question


chenrui333:
👋  looks like right now the pre-release tag usage for the project is blocking homebrew to ship the latest version of the formula.

With that being said, I want to clarify two things:
- Can we use the `latest release` instead of `pre-release` tag? In this way, it sounds more semantically correct?
- Besides the tag usage, do you folks consider the most recent pre-release is a stable release? 

Thanks!!

cc @SMillerDev @dtrodrigues

relates to https://github.com/Homebrew/homebrew-core/pull/60168

charlespierce:
Hi @chenrui333, thanks for the question! Our general philosophy has been to mark it as a `pre-release` until we reach 1.0, for reasons of both stability and feature readiness. That said, at this point we definitely consider Volta stable, so the outstanding concern is having all of the features we consider necessary for 1.0.

However, given that version 1.0 itself serves as a marker for that feature-completeness and that we are stable and committed to a smooth upgrade process for new versions, I think it's reasonable to drop the `pre-release` label. Our next release will most likely be 0.9.0, which will include the last of the significant changes needed before 1.0, so at that point I think dropping the `pre-release` tag makes sense.

chenrui333:
@charlespierce wonderful, I will wait for the soon-release 0.9.0 then. Thanks!!

chenrui333:
Yeah, I will seek keeping the formula up to date later after v0.9.0

chenrui333:
@charlespierce 👋 just checking, looks like the latest release 0.9.0 is still pre-release? Thanks!

charlespierce:
Hi @chenrui333, thanks for checking! I knew there was something with the homebrew formula I was missing. Let me remove the tag (and update the CI script to not add it any more).

charlespierce:
Updated the release to not be pre-release any more :+1:

chenrui333:
> Hi @chenrui333, thanks for checking! I knew there was something with the homebrew formula I was missing. Let me remove the tag (and update the CI script to not add it any more).

😄  thanks!! Appreciate it.

PR #856, opened 2020-10-22
Remove pre-release flag from auto-generated releases

charlespierce:
As discussed in #803, though we aren't yet at 1.0, Volta is production-ready (and we're using it at scale within LinkedIn). We have been committed to stability and backwards-compatibility for some time, so it is reasonable to no longer use the "Pre-release" flag on GitHub releases.

This also unlocks updates to the Volta formula in Homebrew, which requires that packages not be pre-release to be included in the public repo.

Updating the CI script that creates the draft release to _not_ set the pre-release flag.


chriskrycho (approved):
👍

PR #868, opened 2020-11-13
Use the matches! macro instead of manual match blocks

charlespierce:
Info
-----
Clippy now gives warnings about match expressions that could be better written using the [`matches!`](https://doc.rust-lang.org/std/macro.matches.html) macro.

Changes
-----
Updated each expression to use `matches!` instead of a hand-written match block.

Note
-----
The `matches!` macro was added in Rust 1.42, so we need to bump our MSRV. It's currently at 1.41, so it's not a huge bump. Created https://github.com/volta-cli/website/pull/86 to update the documentation to better reflect our workflow.


chriskrycho (approved):
A lovely little ergonomic improvement! 💙 ⚙️

PR #872, opened 2020-11-24
Remove 'tar' output from installer

charlespierce:
As pointed out on Discord, the installer currently includes some output from `tar` that looks out of place:

```
x volta
x volta-shim
x volta-migrate
```

This is because `tar` is invoked with `-v` for verbose output. We actually don't need that output to be part of the install process, so removing that flag from the installer.

Note: Also removed `-z` which has no use when extracting a tarball.


chriskrycho (approved):
🤩 ⚡

PR #873, opened 2020-11-24
Add feature flag for pnpm support

charlespierce:
Info
-----
* Starting work towards supporting `pnpm` in Volta
* Adding a feature flag so that the work can be broken up into multiple PRs without interrupting the trunk.

Changes
-----
* Added `pnpm` feature to the root project, as well as `volta-migrate`, `volta-layout`, and `volta-core`


dherman (approved):
Great call rolling this out with a feature flag. :shipit: !

PR #874, opened 2020-11-24
[Small] Add new layout for pnpm dirs

charlespierce:
Info
-----
* To add support for pnpm, we need to have a `pnpm` image & inventory directories to store the installed files.
* We also want to automatically generate that directory when necessary.

Changes
-----
* Added a `v4` layout and migration, which is primarily a copy of the `v3` layout except that it adds the `pnpm` inventory & image directories (and functions to access the subdirectories).
* Updated the `layout` module in `volta_core` to use `v4` when the feature flag is set.

Tested
-----
* Confirmed that everything compiles correctly both with and without the feature flag enabled.

### DIFF
--- a/crates/volta-core/src/layout/mod.rs
+++ b/crates/volta-core/src/layout/mod.rs
@@ -6,7 +6,10 @@ use cfg_if::cfg_if;
 use double_checked_cell::DoubleCheckedCell;
 use dunce::canonicalize;
 use lazy_static::lazy_static;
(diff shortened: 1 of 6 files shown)


charlespierce:
Closing as `pnpm` support needs some more design / technical work before it can be implemented. See the [Road to 1.0 RFC](https://github.com/volta-cli/rfcs/pull/38) for more info.

PR #875, opened 2020-11-24
[Medium] Add pnpm resolve function and placeholder Tool implementation

charlespierce:
Info
-----
* First step of installing a version of `pnpm`: We need to resolve the version to an absolute version.
* There is a fair amount of interconnectivity, so need to create some placeholder implementations to get everything compiling correctly.

Changes
-----
* Add `resolve` function for `pnpm`, fetching the information from the npm registry (Same as we do for npm and Yarn)
* Add hooks support to change the index URL and support installing pnpm from a private repository as well.
* Create placeholder implementation of `Tool` to wire everything up correctly.

Tested
-----
* Confirmed that everything builds correctly both with and without the feature flag enabled.
* Verified locally that running `volta install pnpm` with the feature flag enabled correctly resolves the version before panicking.
* Also fixed a small typo ("Node" vs "npm") in an unrelated error message.

Note
-----
* This PR builds on #874, so will remain a draft until that is merged. To see the changes from only this PR, use [this link](https://github.com/volta-cli/volta/pull/875/commits/547d7a296b2b2017c5fb36f752dc6d8e2c88ad60)

PR #876, opened 2020-11-25
[Medium] Implement Fetching for Pnpm

charlespierce:
Info
-----
* Next step in making `pnpm` work: Implement the ability to fetch a specific version to the local cache.
* This mirrors the approach that we use for `npm` nearly exactly, except that instead of overwriting the launchers, we need to create them (as the `pnpm` package doesn't include them by default).

Changes
-----
* Updated the `inventory` functions to include pnpm.
* Added the `fetch` function to the `pnpm` module, mostly copied from `npm::fetch`, except using "pnpm" everywhere.
* Updated the `overwrite_launcher` and `overwrite_cmd_launcher` methods to be `create_launcher` and `create_cmd_launcher` and make them point at the correct `pnpm` files.
* Also removed an unnecessary `pub(super)` from the `Node`, `Npm`, and `Yarn` structs.

Tested
-----
* Confirmed that the build completes successfully both with and without the `pnpm` feature flag.
* Verified locally that `volta fetch pnpm` works as expected, and that the resulting launcher scripts work.

Notes
-----
* This doesn't implement the `install` or `pin` behaviors, as those will rely on changes to the configuration formats.
* This PR builds on #875 and so will remain a draft until that is merged. To see only the changes from this PR, use [this link](https://github.com/volta-cli/volta/pull/876/commits/f000c1500fd9813d55f41dab799f82148685542d)

### DIFF
--- a/crates/volta-core/src/error/kind.rs
+++ b/crates/volta-core/src/error/kind.rs
@@ -307,6 +307,12 @@ pub enum ErrorKind {
         tool: String,
     },
 
+    /// Thrown when there is no pnpm version matching the requested Semver/Tag
+    #[cfg(feature = "pnpm")]
+    PnpmVersionNotFound {
+        matching: String,
+    },
+
     /// Thrown when executing a project-local binary fails
     ProjectLocalBinaryExecError {
         command: String,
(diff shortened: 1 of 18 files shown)


charlespierce:
Closing as `pnpm` support needs some more design / technical work before it can be implemented. See the [Road to 1.0 RFC](https://github.com/volta-cli/rfcs/pull/38) for more info.

PR #885, opened 2020-12-03
Clean up warnings from clippy

charlespierce:
Info
-----
* The latest clippy is showing a few new warnings to clean up in the code

Changes
-----
* Updated code to conform to clippy's suggested style

Tested
-----
* All tests continue to pass
* Manually verified that the Node index cache still respects the `url` and that the `version_serde` deserializer still strips a single preceding `v` character.


chriskrycho (approved):
These are all really nice. 👍 I especially like the switch to `ok_or` where there’s no allocation involved since the enum variant is just the value.

PR #888, opened 2020-12-08
[Large] Add support for 'npm link'

charlespierce:
Info
-----
* One of the main outstanding features prior to 1.0 release: Support for `npm link`, both to link the current project as a global _and_ to pull a global in as a link into the current project.
* The bare command (`npm link`) works similarly to a global install, except that internally npm create a symlink instead of copying the files into the global directory.
    * Notably, it _does_ hook up any bins of the current project as global binaries.
* The command with package arguments (`npm link my_package`) will install the global `my_package` into the current project as a symlink.

Changes
-----
* Updated the command interception parser to handle `npm link` both with and without additional arguments.
    * In the case that there are _no_ additional arguments, we can re-use the `PackageInstallCommand` executor, since an `npm link` is functionally the same as an `npm install --global`, except for the symlink behavior.
    * In the case that there _are_ arguments, we split the calls out and run them sequentially, similar to how we handle `npm i -g` with multiple arguments. This allows us to set the appropriate environment variables so that `npm link my_package` works as expected and can find the appropriate global package.
* Added a `PackageLinkCommand` executor which handles the second case above.
    * Using the name of an individual tool, we can set the `prefix` option so that `npm` will correctly locate the global link.
    * Also added some checks to confirm that the command will work and provide useful error messages / warnings for when things may go wrong.

Tested
-----
* Added unit tests to cover the parsing logic
* Acceptance tests will be added in a follow-up PR
* Local testing confirms that a local package can be linked to global and then linked into a separate package.

Notes
-----
* There is no support here for `yarn link`: `yarn link` has different semantics than `npm link`, as it uses a different store location that doesn't overlap with the global installs. Yarn also already uses a directory _outside_ of where it's installed to hold the links, so `yarn link` already works exactly as it was designed and doesn't need additional support from Volta.
* Support for `npm unlink` will be added in a follow-up PR

### DIFF
--- a/crates/volta-core/src/tool/package/manager.rs
+++ b/crates/volta-core/src/tool/package/manager.rs
@@ -124,15 +124,16 @@ fn get_npm_package_name(mut source_dir: PathBuf) -> Option<String> {
 fn get_single_directory_name(parent_dir: &Path) -> Option<String> {
     let mut entries = read_dir_eager(parent_dir)
         .ok()?
-        .filter_map(
-            |(entry, metadata)| {
-                if metadata.is_dir() {
-                    Some(entry)
-                } else {
-                    None
-                }
-            },
-        );
+        .filter_map(|(entry, metadata)| {
+            // If the entry is a symlink, _both_ is_dir() _and_ is_file() will be false. We want to
+            // include symlinks as well as directories in our search, since `npm link` uses
+            // symlinks internally, so we only exclude files from this search
+            if !metadata.is_file() {
+                Some(entry)
+            } else {
+                None
+            }
+        });
 
     match (entries.next(), entries.next()) {
         (Some(entry), None) => entry.file_name().into_string().ok(),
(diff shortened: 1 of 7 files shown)


crates/volta-core/src/tool/package/manager.rs
```diff
@@ -124,15 +124,16 @@ fn get_npm_package_name(mut source_dir: PathBuf) -> Option<String> {
...
-                } else {
-                    None
-                }
-            },
-        );
+        .filter_map(|(entry, metadata)| {
+            // If the entry is a symlink, _both_ is_dir() _and_ is_file() will be false. We want to
+            // include symlinks as well as directories in our search, since `npm link` uses
+            // symlinks internally, so we only exclude files from this search
+            if !metadata.is_file() {
```
chriskrycho, line 131:
```suggestion
            // If the entry is a symlink, _both_ is_dir() _and_ is_file() will be false. We want to
            // include symlinks as well as directories in our search, since `npm link` uses
            // symlinks internally, so we only exclude files from this search
            if metadata.is_dir() || metadata.file_type().is_symlink() {
```

chriskrycho (approved):

PR #890, opened 2020-12-08
Use bordered logo to support viewing in Dark Mode

charlespierce:
Info
-----
* GitHub recently released Dark mode, and our Logo doesn't display well when that is turned on, because the text is black.
* In https://github.com/volta-cli/assets/pull/2, we added logos that include a white border to ensure they remain readable even in dark mode.

Changes
-----
* Update the logo in the README to use the bordered logo asset.
* Renamed the asset to use the correct file extension (it always was a `.png` file, even though it was named `.jpg`)
* Before the change, the README in dark mode:
Screen Shot 2020-12-08 at 11 39 33 AM
* After, it now looks like this:
Screen Shot 2020-12-08 at 11 39 02 AM


dherman (approved):
Big improvement :) 🚢

PR #892, opened 2020-12-09
Only show stdout and stderr of spawned process in dev mode

haochuan:
- Use `cfg!(debug_assertions)` to check if it's a dev build or release build.
- For dev build, pipe `stdout` and `stderr` of the `command` in `spawn_process`, so it's easier to debug.
- For release build, pipe `stdout` and `stderr` to "nowhere", so the spawn_process won't mess up with `stdout` of `volta` if there is any thing from `stdout` or `stderr`

### DIFF
--- a/crates/volta-core/src/monitor.rs
+++ b/crates/volta-core/src/monitor.rs
@@ -46,8 +46,8 @@ fn spawn_process(command: &str) -> Option<Child> {
         child.args(command.split(' ').skip(1));
         child.stdin(Stdio::piped());
 
-        #[cfg(debug_assertions)]
-        // Only show stdout and stderr of spawned process in dev mode
+        #[cfg(not(debug_assertions))]
+        // Hide stdout and stderr of spawned process in release mode
         child.stdout(Stdio::null()).stderr(Stdio::null());
 
         match child.spawn() {
--- a/crates/volta-core/src/monitor.rs
+++ b/crates/volta-core/src/monitor.rs
@@ -42,22 +42,15 @@ impl Monitor {
 
 fn spawn_process(command: &str) -> Option<Child> {
     command.split(' ').take(1).next().and_then(|executable| {
-        // Only show stdout and stderr of spawned process in dev mode
-        let child = if cfg!(debug_assertions) {
-            create_command(executable)
-                .args(command.split(' ').skip(1))
-                .stdin(Stdio::piped()) // JSON data is sent over stdin
-                .spawn()
-        } else {
-            create_command(executable)
-                .args(command.split(' ').skip(1))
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/monitor.rs
```diff
@@ -42,11 +42,21 @@ impl Monitor {
...
+                .stdin(Stdio::piped()) // JSON data is sent over stdin
+                .spawn()
+        } else {
+            create_command(executable)
+                .args(command.split(' ').skip(1))
+                .stdin(Stdio::piped()) // JSON data is sent over stdin
+                .stdout(Stdio::null()) // Pipe stdout to nowhere
+                .stderr(Stdio::null()) // Pipe stdout to nowhere
+                .spawn()
+        };
```
charlespierce, line 58: Since the only difference between the two cases is the `stdout` and `stderr` calls, I think we can DRY this slightly to make it more clear by creating the command and applying the changes individually instead of chaining them. Something like:

```rust
let mut command = create_command(executable);
command.args(...);
command.stdin(...);

#[cfg(debug_assertions)]
{
    command.stdout(...);
    command.stderr(...);
}

match command.spawn() {
```

I haven't fully worked out if something like that works with the return values, but the key thing is that you don't _have_ to chain all of the `Command` methods, you can call them individually.

charlespierce (changes requested):
Thanks for this, this is a great change to prevent leaking confusing data to the end users! One small suggestion for avoiding duplication, but other than that LGTM!

crates/volta-core/src/monitor.rs
haochuan: Cool got it thanks!

crates/volta-core/src/monitor.rs
```diff
@@ -42,12 +42,15 @@ impl Monitor {
...
-            // .stdout(Stdio::piped()) // let the plugin write to stdout for now
-            .spawn();
-        match child {
+        let mut child = create_command(executable);
+        child.args(command.split(' ').skip(1));
+        child.stdin(Stdio::piped());
+
+        #[cfg(debug_assertions)]
+        // Only show stdout and stderr of spawned process in dev mode
+        child.stdout(Stdio::null()).stderr(Stdio::null());
```
charlespierce, line 51: I think this is backwards. Setting the output to `Stdio::null()` will hide it, so this is only hiding the output in debug mode while leaving it in release mode.
```suggestion
        #[cfg(not(debug_assertions))]
        // Hide stdout and stderr of spawned process in release mode
        child.stdout(Stdio::null()).stderr(Stdio::null());
```

charlespierce (approved):
LGTM, thanks!

PR #893, opened 2020-12-09
pass detailed activity kind to event for node/yarn/npm/npx

haochuan:
I'm not 100% confident that this is the right way, so feel free to let me know your suggestions if there is any.

My current change uses a helper function `get_tool_name` in `shim.rs` to help getting the actual command name (`node/yarn/npm/npx`), then pass the corresponding `ActivityKind` to `add_event_start` and `add_event_end`. In this way I think the code looks a little cleaner, though in `.../run/mod.rs` there will be a duplicated `get_tool_name` call.

We can also avoid the duplicated call by:

- move `get_tool_name` out of `.../run/mod.rs` and only use it in `shim.rs`
- pass additional args to `execute_shim`

But I feel this will let the code look a little complicated, since the cost of running `get_tool_name` twice should be very minimum.


charlespierce (changes requested):
My initial thought is that in this case we should leave the `Tool` events there as the high-level wrapper (in the same way that we have `Volta` events for Volta commands.

Instead, we should add an _additional_ event for the start of the command processing for each tool kind. We can do this in the `command` method for each tool in `volta_core::run` (e.g. https://github.com/volta-cli/volta/blob/main/crates/volta-core/src/run/node.rs#L11). That way we don't need to rely on additional string matching to pull out the tool kind, as we've already done that.

I think I mentioned this, but we used to have it set up to do exactly that (see https://github.com/volta-cli/volta/blob/v0.8.7/crates/volta-core/src/run/node.rs#L10), but it looks like I may have accidentally deleted it during a refactor to 0.9.0.

haochuan:
Sorry seems like my `upstream fetch` messed up the commits

charlespierce:
Yeah, any chance you could do a rebase? Or, if that gets wildly out of hand (as I suspect it might since a bunch of the commits are duplicates), start the branch over and force push it? The change itself is :+1:, just want to avoid cluttering up the history if possible.

PR #896, opened 2020-12-10
[DO NOT MERGE] Build with rustls instead of OpenSSL

charlespierce:
Experimenting how using Rustls affects final binary size

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,49 +9,17 @@ on:
 name: Production
 
 jobs:
-  centos:
-    name: Build - CentOS / RHEL
-    runs-on: ubuntu-16.04
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v2
-      - name: Set up docker image
-        run: docker build -t volta .
-        working-directory: ./ci/docker
-      - name: Compile and package Volta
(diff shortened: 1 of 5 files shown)


charlespierce:
Looks like `ring` doesn't like to compile on CentOS 6 since it's so old, but that's the level we've been supporting, so using RusTLS would mean changing our supported environments.

chriskrycho:
@charlespierce @dherman @thoov @rwjblue thoughts on leaving behind CentOS 6 as one of the things we do with 1.0? Even if that means we cut a 0.10.0 with support for it first, and then 1.0.0 which drops it (and does e.g. this if it's useful).

charlespierce:
@chriskrycho It's not specifically CentOS 6, we were using that to get the CentOS flavor of OpenSSL, however using an older docker image allows us to create a build linked against an older `libc` version than the oldest GH Actions runner supports (Ubuntu 16.04). That makes Volta work on a wider range of Linux distros out of the box.

I'm hesitant to drop that support, since there's such a variety of distros and support timeframes. For reference, I believe `rustup` is actually built against CentOS _5_, to support an even wider range of possible distros. If, in the future, we decide we want to use RusTLS, we can probably do it in a backwards-compatible way by creating a feature flag and still building the old OpenSSL versions on an older image. It will require making the installer script a little smarter, but that shouldn't be too hard.

chriskrycho:
Oh, I see. Thank you for the clarification!

PR #897, opened 2020-12-11
use debug! instead of error! in monitor.rs

haochuan:
There are two places in `monitor.rs`, we should use `debug!` instead of `error!`.


charlespierce (approved):
:shipit:

Issue #489, opened 2019-06-27
Convert custom fallible `into_...` implementations to `TryFrom` implementations


chriskrycho:
Rust 1.34.0 [stabilized] the [`TryFrom`] and [`TryInto`] traits. We have a number of conversion methods in our code which have signatures like `fn into_foo(bar: Bar) -> Fallible<Foo>`; all of these could be implemented using `TryFrom` instead, which would simplify call sites, allowing them to use `from` or `into` directly.

[stabilized]: https://blog.rust-lang.org/2019/04/11/Rust-1.34.0.html#tryfrom-and-tryinto
[`TryFrom`]: https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html
[`TryInto`]: https://doc.rust-lang.org/stable/std/convert/trait.TryInto.html

(I'm happy to mentor someone who wants a small first contribution through this one – it should be relatively mechanical, but it'd be a good way to get your feet wet in the codebase and it's valuable to us!)

dagadbm:
hey!
so I am a currently a front end developer (worked with Java/C#/Python in the past).
I have no idea of how rust works but I know I like volta and I just used it on WSL on windows and on my mac to replace the god awful nvm .
I also wanted to get into open source but i have no idea where to start and this seems an "easy" find and replace. what can I do to help on this?

chriskrycho:
Hey @dagadbm – welcome, we're glad to have you and happy to help you spin up a bit!

The gist is that a bunch of implementations that currently look like this:

https://github.com/volta-cli/volta/blob/1c8c9b2257bca69e883a89229715da10fef9469f/crates/volta-core/src/inventory/serial.rs#L128-L146

Will become this:

```rust
use std::convert:TryFrom;

impl TryFrom<NodeIndex> for <super::NodeIndex> {
    pub fn into_index(self) -> Fallible<super::NodeIndex> {
        let mut entries = Vec::new();
        for entry in self.0 {
            if let Some(npm) = entry.npm {
                let data = super::NodeDistroFiles {
                    files: HashSet::from_iter(entry.files.into_iter()),
                };
                entries.push(super::NodeEntry {
                    version: entry.version,
                    npm,
                    files: data,
                    lts: entry.lts,
                });
            }
        }
        Ok(super::NodeIndex { entries })
    }
}
```

I recommend digging in just a *little bit* into Rust docs first is because some of the functions which look like they *might* be amenable to this *aren't*, because they require some additional arguments.

[...]

PR #907, opened 2021-01-05
Replace Platform::from_json with TryFrom trait implementation

sirreal:
Replace `Platform::from_json` with `Platform::try_from`, implement `TryFrom<String>` trait.

This is an attempt at part of https://github.com/volta-cli/volta/issues/489.

I'm aware that this may not be what was intended in #489. I'm prepared to accept that feedback and close if that's the case 🙂 

One drawback is that `try_from` doesn't mention JSON, whereas `from_json` clearly expects a JSON string.


charlespierce (approved):
This is exactly what was meant in #489, thanks @sirreal!

PR #908, opened 2021-01-05
Replace Platform::into_platform with Into implementation

sirreal:
Part of https://github.com/volta-cli/volta/issues/489

Replace into_platform with TryInto implementation. 

The first commit, 39e6d3a, includes a very simple change that implements `Into<Option<PlatformSpec>>`.

The second commit goes further and implements `TryInto<PlatformSpec>` that may result in a `VoltaError` of kind `ErrorKind::NoPlatform`. Implementing `TryInto` seems more idiomatic than implementing `into` to return an `Option`. I am very new to this codebase and a Rust novice, this may be going too far and the first commit may be sufficient.

### DIFF
--- a/crates/volta-core/src/toolchain/mod.rs
+++ b/crates/volta-core/src/toolchain/mod.rs
@@ -1,4 +1,3 @@
-use std::convert::Into;
 use std::convert::TryFrom;
 use std::fs::write;
 
--- a/crates/volta-core/src/toolchain/mod.rs
+++ b/crates/volta-core/src/toolchain/mod.rs
@@ -1,3 +1,4 @@
+use std::convert::Into;
 use std::convert::TryFrom;
 use std::fs::write;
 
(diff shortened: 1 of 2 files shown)


charlespierce:
Hi @sirreal, thanks for taking this on! For this case, I think that the first commit is closer to what we want: The platform not existing isn't an error, rather it's just an indication that there isn't anything set (which is perfectly valid).

Within that commit, the main change I would make is that it's more idiomatic Rust to implement `From` instead of `Into`. The reason for that is there is a blanket implementation of `Into` for all types that implement `From` (see [the docs](https://doc.rust-lang.org/std/convert/trait.Into.html#impl-Into%3CU%3E)), which means by implementing `From`, you get `Into` for free.

Also, both `From` and `Into` are part of the Rust prelude, so you shouldn't need to manually import them with `use std::convert::From`

charlespierce (approved):
Looks perfect, thanks! I merged #907, so it looks like that created some merge conflicts.

sirreal:
Thanks for the feedback! I've rebased on main after the merge. It looks like there are some CI failures, but the issues don't appear to be related.

charlespierce:
Yep, looks like an intermittent failure with the network connectivity (we've seen this before). Seems to be alright now :+1: Thanks again!

PR #910, opened 2021-01-08
Remove now-unused 'manifest' module

### LINKED ISSUE #749: Clean up `manifest` module
Follow-on to implementing Workspaces Support (https://github.com/volta-cli/rfcs/pull/43): The bulk of the `manifest` crate is the `BinManifest` type, which is only used in one place: `tool::package::install`. We should move it closer to where it is actually used, and not have it sitting alone in a different module.

We should also look to avoid using [`Deref`](https://doc.rust-lang.org/std/ops/trait.Deref.html) and [`DerefMut`](https://doc.rust-lang.org/std/ops/trait.DerefMut.html), since the docs for those indicate they should only be used for smart pointers.

This can likely be combined with #526 as a general refactor of the custom types in the `tool` module, perhaps adding a `metadata` sub-module that holds the various structs used to interpret metadata about packages (index formats, `npm view`, etc.)

charlespierce:
Closes #749 

Info
-----
* As part of the package installation refactor in 0.9.0, the `BinManifest` type was rebuilt as `PackageManifest` in `tool::package::metadata`.
* A result of this is that `BinManifest`, which was the only type remaining in the `manifest` module, is no longer used.

Changes
-----
* Removed the entire `manifest` module, as none of that code is actually used any more.


mikrostew (approved):
Love these kind of PRs 💯

PR #915, opened 2021-01-13
[Small] Check Node OS & Arch availability and support M1 native builds

### LINKED ISSUE #34: resolving public node should check OS availability
`NodeCatalog::resolve_public` scans for available version of node from the public list, but it doesn't make sure each version it checks is actually available for the current OS. Add this logic to the `skip_while` callback.

charlespierce:
Closes #34 

Info
-----
* Now that Rust 1.49 is out and has stable [Tier 2 support for Apple M1 chips](https://blog.rust-lang.org/2020/12/31/Rust-1.49.0.html#64-bit-arm-macos-and-windows-reach-tier-2), we can create an M1 native build of Volta.
* Currently, Node doesn't provide pre-built binaries for M1 chips, so we still need to fetch the x64 version of Node and run it through Rosetta 2.
* At the same time, we should also confirm when resolving a Node version that it is available for the current OS & Architecture.

Changes
-----
* Updated the `NODE_DISTRO_*` constants to also include `NODE_DISTRO_IDENTIFIER`, representing the identifier used in the Node index `files` property.
    * Currently this uses the `osx-x64-tar` identifier on M1 MacOS, since Node doesn't provide pre-built M1 binaries and hasn't yet indicated what the identifier will be).
* Updated the parsing of the Node index to exclude any versions that aren't available for the current OS / Architecture.
* Removed some unused properties and traits.

### DIFF
--- a/crates/volta-core/src/tool/node/metadata.rs
+++ b/crates/volta-core/src/tool/node/metadata.rs
@@ -1,37 +1,19 @@
 use std::collections::HashSet;
-use std::str::FromStr;
 
+use super::NODE_DISTRO_IDENTIFIER;
 use crate::version::{option_version_serde, version_serde};
 use semver::Version;
 use serde::{Deserialize, Deserializer};
-use serde_json::Error;
 
 /// The index of the public Node server.
 pub struct NodeIndex {
     pub(super) entries: Vec<NodeEntry>,
 }
 
-impl FromStr for NodeIndex {
-    type Err = Error;
-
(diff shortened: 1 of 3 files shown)


chriskrycho (approved):
Looks good! ⚡

chriskrycho:
Tested on my M1 MacBook Air—

- launched native (non-Rosetta) Terminal
- created a new project
- `volta pin node@15 npm@7`

—it works perfectly!

PR #916, opened 2021-01-13
[Small] Remove now-unused 'LtsRequirement' tag

charlespierce:
Info
-----
* With the previous package installation logic, we relied on finding a Node version that matched `engines` to determine which Node version to use when installing a global package.
* To better match user expectations, we added an internal-only `LtsRequirement` tag, which represented finding the latest LTS version that matched a requirement (if possible), falling back to the latest version overall.
* Now that we match `npm i -g` more closely and use the current default Node version for global installs, that internal tag is no longer needed.

Changes
-----
* Removed the `LtsRequirement` variant from the `VersionTag` enum.
* Removed the `resolve_lts_semver` function which was used only with that variant.


chriskrycho (approved):
:shipit:

PR #917, opened 2021-01-14
[Small] Build Apple Silicon Native build in CI

charlespierce:
Info
-----
* With #915, we can now create native M1 MacOS binaries.
* We should create them in CI so they are included in future releases.

Changes
-----
* Updated the CI scripts to build both Intel and ARM Mac builds (the latter being named `volta-{version}-macos-aarch64.tar.gz`
* Updated the release job to also include the native build in the release artifacts.


chriskrycho (approved):
😍 ⚡

PR #926, opened 2021-01-20
[Small] Clean up 'volta list' source lookup behavior in non-Volta projects

### LINKED ISSUE #761: `volta list` says no tools installed when package is not configured for volta

charlespierce:
Closes #761 

Info
-----
* Our logic for determining the source of a tool for `volta list` currently has an incorrect short-circuit when we are in a project that doesn't have a Volta setting at all.

Changes
-----
* Updated the logic to correctly fall through to check the default if we have a project but no platform specified.

Tested
-----
* Locally confirmed that running `volta list` and `volta list node` from a project with no `"volta"` key works as it should.

### DIFF
--- a/src/command/list/toolchain.rs
+++ b/src/command/list/toolchain.rs
@@ -55,30 +55,30 @@ impl Lookup {
         default_platform: Option<&PlatformSpec>,
         version: &Version,
     ) -> Source {
-        let default_or_none = default_platform
-            .and_then(self.version_from_spec())
-            .and_then(|ref default_version| {
-                if default_version == version {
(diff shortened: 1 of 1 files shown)


mikrostew (approved):
LGTM!

PR #941, opened 2021-02-14
[Tiny] Make tool name case-insensitive on Windows

### LINKED ISSUE #939: volta-shim(s) not callable by capitalized extension (EXE) on Windows 10
```
C:\Users\user>volta-shim.EXE
'volta-shim.EXE' is not recognized as an internal or external command
```

```
C:\Users\user>volta-shim.exe
Volta error: 'volta-shim' should not be called directly.
```

I came accross this issue from https://github.com/starship/starship/issues/2313 which recently changed the way the node-binary is called

charlespierce:
Closes #939 

Info
-----
* Our logic for detecting the tool name on Windows was assuming that the command would always be called as e.g. `node` or `node.exe`, fully lowercase.
* On Windows, the file system is case-insensitive, so it's possible to call the tool as `NODE.EXE` or even `nOdE.eXe` or any capitalization.
* We should detect those situations and treat them appropriately.

Changes
-----
* Updated the Windows implementation of getting the tool name from the file name to convert the tool to lowercase before removing the `.exe` and trying to match the command name.

Tested
-----
* Confirmed that running `node.exe --version`,  `NODE.EXE --version`, and `nOdE.eXe --version` all work correctly.

Notes
-----
* The detection of binaries is based on locating the configuration on disk, so lowercasing the tool name should still work correctly, as the configuration file will be case-insensitive as well, meaning it will be correctly found.

### DIFF
--- a/crates/volta-core/src/run/mod.rs
+++ b/crates/volta-core/src/run/mod.rs
@@ -107,7 +107,7 @@ fn tool_name_from_file_name(file_name: &OsStr) -> OsString {
     // On Windows PowerShell, the file name includes the .exe suffix
     // We need to remove that to get the raw tool name
     match file_name.to_str() {
-        Some(file) => OsString::from(file.trim_end_matches(".exe")),
+        Some(file) => OsString::from(file.to_ascii_lowercase().trim_end_matches(".exe")),
         None => OsString::from(file_name),
     }
 }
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/run/mod.rs
```diff
@@ -107,7 +107,7 @@ fn tool_name_from_file_name(file_name: &OsStr) -> OsString {
     // On Windows PowerShell, the file name includes the .exe suffix
```
mikrostew, line 107:
```suggestion
    // On Windows PowerShell, the file name includes the .exe suffix,
    // and the Windows file system is case-insensitive
```

mikrostew (approved):
Added a comment to the code, but LGTM

PR #945, opened 2021-02-20
[Medium] Add extra directory when linking scoped package

### LINKED ISSUE #934: "npm link" fails to link correctly (may be an npm v7 issue)
When I use `npm link` in a package (with scope), e.g. `@foo/bar` it generates a shortcut link points to a non-existing place.

So if I look in 

```shell
ls -l /Users/XXXXX/.volta/tools/image/packages/@foo/bar/lib/node_modules/@bar
```

I get `amazon-product-catalog-client -> ../../../../../../../../code/mono/packages/XXX/bar` which is _almost_ fine, except that it has too many `../`-s.

Given that there's no open issue on this, and given that I'm using npm v7 (with Node v15.7), then I'm guessing the issue is compatibility with npm v7.

charlespierce:
Closes #934 

Info
-----
* With npm 7, when we run `npm link` it uses a relative path to link the local directory to the global directory.
* Since we link into a temp directory and then `rename` to the final `tools/images` directory, we need to make sure that the temp directory has the same level of nesting as the final directory, so that relative symlinks work correctly.
* When linking a package that has a scope, the final directory includes an extra level of nesting, from `@scope/package`, so we need to account for that when creating the temp directory.

Changes
-----
* Updated the install executor and helper struct (`DirectInstall`) to support including the name if available.
* Updated `setup_staging_directory` to handle the case when an extra layer of nesting is required.
* Connected the argument parser to correctly fetch the name and pass it through when `npm link` is run, ensuring that scoped packages are handled correctly.

Tested
-----
* Local testing confirmed that running `npm link` in a scoped package sets up the relative symlink correctly.
* Local testing to ensure that the non-scoped packages are still handled correctly.

### DIFF
--- a/crates/volta-core/src/run/parser.rs
+++ b/crates/volta-core/src/run/parser.rs
@@ -348,7 +348,8 @@ impl<'a> LinkArgs<'a> {
     pub fn executor(self, platform: Platform, project_name: Option<String>) -> Fallible<Executor> {
         if self.tools.is_empty() {
             // If no tools are specified, then this is a bare link command, linking the current
-            // project as a global package. We treat this exactly like a global install
+            // project as a global package. We treat this like a global install except we look up
+            // the name from the current directory first.
             match project_name {
                 Some(name) => PackageInstallCommand::for_npm_link(self.common_args, platform, name),
                 None => PackageInstallCommand::new(self.common_args, platform, PackageManager::Npm),
--- a/crates/volta-core/src/run/parser.rs
+++ b/crates/volta-core/src/run/parser.rs
@@ -345,12 +345,15 @@ pub struct LinkArgs<'a> {
 }
 
 impl<'a> LinkArgs<'a> {
-    pub fn executor(self, platform: Platform) -> Fallible<Executor> {
+    pub fn executor(self, platform: Platform, project_name: Option<String>) -> Fallible<Executor> {
         if self.tools.is_empty() {
             // If no tools are specified, then this is a bare link command, linking the current
             // project as a global package. We treat this exactly like a global install
-            PackageInstallCommand::new(self.common_args, platform, PackageManager::Npm)
-                .map(Into::into)
+            match project_name {
+                Some(name) => PackageInstallCommand::for_npm_link(self.common_args, platform, name),
+                None => PackageInstallCommand::new(self.common_args, platform, PackageManager::Npm),
+            }
+            .map(Into::into)
         } else {
             // If there are tools specified, then this represents a command to link a global
             // package into the current project. We handle each tool separately to support Volta's
--- a/crates/volta-core/src/tool/package/mod.rs
+++ b/crates/volta-core/src/tool/package/mod.rs
@@ -35,7 +35,7 @@ pub struct Package {
 
 impl Package {
     pub fn new(name: String, version: VersionSpec) -> Fallible<Self> {
-        let staging = setup_staging_directory(PackageManager::Npm, false /* needs_scope */)?;
+        let staging = setup_staging_directory(PackageManager::Npm, NeedsScope::No)?;
 
         Ok(Package {
             name,
@@ -123,6 +123,9 @@ impl Display for Package {
 ///
 /// Provides methods to simplify installing into a staging directory and then moving that install
 /// into the proper location after it is complete.
+///
+/// Note: We don't always know the name of the package up-front, as the install could be from a
+/// tarball or a git coordinate. If we do know ahead of time, then we can skip looking it up
 pub struct DirectInstall {
     staging: TempDir,
     manager: PackageManager,
@@ -131,7 +134,7 @@ pub struct DirectInstall {
 
 impl DirectInstall {
     pub fn new(manager: PackageManager) -> Fallible<Self> {
-        let staging = setup_staging_directory(manager, false /* needs_scope */)?;
+        let staging = setup_staging_directory(manager, NeedsScope::No)?;
 
         Ok(DirectInstall {
             staging,
@@ -141,7 +144,7 @@ impl DirectInstall {
     }
 
     pub fn with_name(manager: PackageManager, name: String) -> Fallible<Self> {
-        let staging = setup_staging_directory(manager, name.contains('/'))?;
+        let staging = setup_staging_directory(manager, name.contains('/').into())?;
 
         Ok(DirectInstall {
             staging,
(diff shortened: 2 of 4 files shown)


crates/volta-core/src/tool/package/mod.rs
```diff
@@ -35,7 +35,7 @@ pub struct Package {
 
 impl Package {
     pub fn new(name: String, version: VersionSpec) -> Fallible<Self> {
-        let staging = setup_staging_directory(PackageManager::Npm)?;
+        let staging = setup_staging_directory(PackageManager::Npm, false /* needs_scope */)?;
```
chriskrycho, line 38: The need for the comment is… suggestive… of the high utility we'd get from refactoring this to use an enum (_a la_ [Booleans and Enums](https://thoughtbot.com/blog/booleans-and-enums)):

```rust
#[derive(PartialEq)]
enum NeedsScope {
    Yes,
    No,
}
```

Because of this kind of readability/comprehensibility issue, I've actually gotten to the point where I'll just straight-up introduce new boolean enums anywhere it helps. It makes a *very* small difference in the size of the compiled binary and otherwise no difference perf-wise, so it seems very much worth the tradeoff to me.

```suggestion
        let staging = setup_staging_directory(PackageManager::Npm, NeedsScope::No)?;
```

crates/volta-core/src/run/parser.rs
```diff
@@ -345,12 +345,15 @@ pub struct LinkArgs<'a> {
 }
 
 impl<'a> LinkArgs<'a> {
-    pub fn executor(self, platform: Platform) -> Fallible<Executor> {
+    pub fn executor(self, platform: Platform, project_name: Option<String>) -> Fallible<Executor> {
         if self.tools.is_empty() {
             // If no tools are specified, then this is a bare link command, linking the current
             // project as a global package. We treat this exactly like a global install
```
chriskrycho, line 351: Does this comment still actually hold without modification? Also, can we add to it slightly to explain what we're doing with the `name` handling? At first read, it wasn't obvious which piece of the new code it meant, exactly (because it refers to the old code!) and it *doesn't* explain the `name` handling, so I was left slightly scratching my head.

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -126,13 +126,28 @@ impl Display for Package {
 pub struct DirectInstall {
     staging: TempDir,
     manager: PackageManager,
+    name: Option<String>,
```
chriskrycho, line 129: If I've followed the writeup and the PR correctly, this is the *package* name for the installation? Could we go with that instead, or does it seem clear enough once you have the context of the module loaded up? (I'm happy to defer, here; I do *not* have the context loaded up here, and I'm a big fan of the shortest name possible to get the point across.)

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -211,17 +230,22 @@ impl InPlaceUpgrade {
...
-fn setup_staging_directory(manager: PackageManager) -> Fallible<TempDir> {
+fn setup_staging_directory(manager: PackageManager, needs_scope: bool) -> Fallible<TempDir> {
     // Workaround to ensure relative symlinks continue to work.
     // The final installed location of packages is:
     //      $VOLTA_HOME/tools/image/packages/{name}/
     // To ensure that the temp directory has the same amount of nesting, we use:
     //      $VOLTA_HOME/tmp/image/packages/{tempdir}/
     // This way any relative symlinks will have the same amount of nesting and will remain valid
     // even when the directory is persisted.
+    // We also need to handle the case when the linked package has a scope, which requires another
```
chriskrycho, line 241: It's workarounds all the way down. 😂

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -126,13 +126,28 @@ impl Display for Package {
 pub struct DirectInstall {
     staging: TempDir,
     manager: PackageManager,
+    name: Option<String>,
 }
 
 impl DirectInstall {
     pub fn new(manager: PackageManager) -> Fallible<Self> {
-        let staging = setup_staging_directory(manager)?;
+        let staging = setup_staging_directory(manager, false /* needs_scope */)?;
```
chriskrycho, line 134:
```suggestion
        let staging = setup_staging_directory(manager, NeedsScope::No)?;
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -126,13 +126,28 @@ impl Display for Package {
...
-        Ok(DirectInstall { staging, manager })
+        Ok(DirectInstall {
+            staging,
+            manager,
+            name: None,
+        })
+    }
+
+    pub fn with_name(manager: PackageManager, name: String) -> Fallible<Self> {
+        let staging = setup_staging_directory(manager, name.contains('/'))?;
```
chriskrycho, line 144:
```suggestion
        let staging = setup_staging_directory(manager, name.contains('/').into())?;
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -211,17 +230,22 @@ impl InPlaceUpgrade {
 
 /// Create the temporary staging directory we will use to install and ensure expected
 /// subdirectories exist within it
-fn setup_staging_directory(manager: PackageManager) -> Fallible<TempDir> {
+fn setup_staging_directory(manager: PackageManager, needs_scope: bool) -> Fallible<TempDir> {
```
chriskrycho, line 233:
```suggestion
fn setup_staging_directory(manager: PackageManager, needs_scope: NeedsScope) -> Fallible<TempDir> {
```

crates/volta-core/src/tool/package/mod.rs
```diff
@@ -211,17 +230,22 @@ impl InPlaceUpgrade {
...
     // To ensure that the temp directory has the same amount of nesting, we use:
     //      $VOLTA_HOME/tmp/image/packages/{tempdir}/
     // This way any relative symlinks will have the same amount of nesting and will remain valid
     // even when the directory is persisted.
+    // We also need to handle the case when the linked package has a scope, which requires another
+    // level of nesting
     let mut staging_root = volta_home()?.tmp_dir().to_owned();
     staging_root.push("image");
     staging_root.push("packages");
+    if needs_scope {
```
chriskrycho, line 246:
```suggestion
    if needs_scope == NeedsScope::Yes {
```

chriskrycho (changes requested):
Implementation makes good sense to me! I did notice one thing which I think would be really helpful to tweak for long-term maintainability, but no fundamental issues here!

crates/volta-core/src/tool/package/mod.rs
chriskrycho: Alternatively:

```suggestion
        let staging = setup_staging_directory(manager, NeedsScope::from(name.contains('/')))?;
```

crates/volta-core/src/run/parser.rs
charlespierce: That's a good call out, I'll update it to reflect that we treat it like a global install that we definitely know the name up-front (global installs can be from an arbitrarily named tarball or git link, so we don't always know the name a priori).

crates/volta-core/src/tool/package/mod.rs
charlespierce: I definitely went back and forth on this, I _didn't_ like using a boolean, so yeah I think a boolean Enum is worth it!

crates/volta-core/src/tool/package/mod.rs
charlespierce: That's the truth!

crates/volta-core/src/tool/package/mod.rs
charlespierce: It is the package name, yeah. I think it's pretty clear when you have the context, that we're talking about installing a single package, so this is the name. That said, I'll add a comment about how we don't always know the name up-front which should help make it more clear.

chriskrycho: That makes sense, so I think a comment will close the gap nicely. Thank you!

charlespierce:
Updated with some more comments + a boolean Enum

chriskrycho (approved):
Thanks for the iteration!

PR #957, opened 2021-03-06
Fix pinning of npm@bundled

### LINKED ISSUE #953: Pinning bundled npm does not work
Hi,

If I pin a node version in a project (let's say `15.11.0`), this is added to `package.json`:

```
  "volta": {
    "node": "15.11.0"
  }
```
So far so good.

However, if I want to make sure that the bundled npm is used, and run `volta pin npm@bundled`, _nothing happens_, `package.json` is not updated with anything.

This is problematic, because there's no distinction between two cases:
- I want to use the bundled npm.
- I want to use the global default npm.

And in fact, if I set a different npm version as global default, that's gonna be the one that's executed in the project directory, not the bundled one.

To remedy this, I think there could be two solutions:
- If node is pinned in the project, but npm is not, use bundled npm.
  This would intuitively make sense to me, since there supposedly wouldn't be much value in fixing node, but allowing whatever npm, since the point is to standardize the environment (and also, for example, the `package-lock.json` format can depend on npm version, so you want to keep it aligned between developers).
- Introduce `bundled` as a valid version for npm in `package.json`, like on the command line.
  This is more explicit than the above, but for reasons explained there, I prefer that option to this one.

What do you think?

Sh4rK:
When node is pinned in a project, and `volta pin npm@bundled` is run, nothing is added to `package.json`, that is, `npm@bundled` is pinned by default. However, when actually running npm in a project like this, Volta didn't behave correctly, and used the global default npm.

It would be nice to write a test case for the fixed behavior, and I tried, but I don't understand correctly the Sandbox thing, or how I should go about doing it.

After fixing this, I noticed that the `volta list` output is wrong as well in this case, showing the default npm version instead of the bundled one. Could be fixed in a follow-up commit, will need some advice on how to do that.

Fixes #953

### DIFF
--- a/crates/volta-core/src/platform/mod.rs
+++ b/crates/volta-core/src/platform/mod.rs
@@ -244,34 +244,17 @@ impl Platform {
     ///   pulling Yarn from the user default platform, if available
     /// - If there is no Project platform, then we use the user Default Platform
     pub fn current(session: &mut Session) -> Fallible<Option<Self>> {
-        match session.project_platform()? {
-            Some(platform) => {
-                if platform.yarn.is_none() || platform.npm.is_none() {
-                    if let Some(default) = session.default_platform()? {
-                        let npm = platform
-                            .npm
-                            .clone()
-                            .map(Sourced::with_project)
-                            .or_else(|| default.npm.clone().map(Sourced::with_default));
-                        let yarn = platform
-                            .yarn
-                            .clone()
-                            .map(Sourced::with_project)
-                            .or_else(|| default.yarn.clone().map(Sourced::with_default));
-
-                        return Ok(Some(Platform {
-                            node: Sourced::with_project(platform.node.clone()),
-                            npm,
-                            yarn,
-                        }));
-                    }
-                }
-                Ok(Some(platform.as_project()))
+        if let Some(mut platform) = session.project_platform()?.map(PlatformSpec::as_project) {
+            if platform.yarn.is_none() {
+                platform.yarn = session
+                    .default_platform()?
+                    .and_then(|default_platform| default_platform.yarn.clone())
+                    .map(Sourced::with_default);
             }
-            None => match session.default_platform()? {
-                Some(platform) => Ok(Some(platform.as_default())),
-                None => Ok(None),
-            },
+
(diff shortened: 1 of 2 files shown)


crates/volta-core/src/platform/mod.rs
```diff
@@ -244,34 +244,17 @@ impl Platform {
...
-                        }));
-                    }
-                }
-                Ok(Some(platform.as_project()))
+        if let Some(mut platform) = session.project_platform()?.map(PlatformSpec::as_project) {
+            if platform.yarn.is_none() {
+                platform.yarn = session
+                    .default_platform()?
+                    .and_then(|default_platform| default_platform.yarn.clone())
+                    .map(Sourced::with_default);
```
charlespierce, line 252: This is a nice simplification!

charlespierce (approved):
This looks great, thanks! For the `volta list` issue, I actually think that it's a slightly bigger issue, as I don't think `volta list` handles bundled versions correctly. The solution there may actually be a larger refactor of how list pulls the information.

As for the tests, that's actually another area of tech debt we need to clean up. It looks like there aren't any existing tests that cover the `npm` behavior. If you want to take a look at it, the existing tests for the "merging yarn" behavior are in: https://github.com/volta-cli/volta/blob/main/tests/acceptance/merged_platform.rs

In order to set up similar tests, you'll need to copy over the npm fixture information from one of the other test files, e.g. https://github.com/volta-cli/volta/blob/main/tests/acceptance/volta_install.rs#L150 And then set up test cases analogous to the Yarn ones that confirm the default npm isn't used in a pinned project. I'm happy to leave that for a follow-up issue and merge this as-is, however.

Sh4rK:
Okay, I think you can merge this then.

I tried writing the test again based on your advice, but I'm not familiar enough with the terminology and how stuff works, and I don't want to delay this for long :)

Sh4rK:
Never mind, I figured it out 😁
I also ran the tests without the fix, and indeed one of them fails successfully :)

charlespierce (approved):
Awesome, thanks for adding the tests! And for confirming that they fail without the changes :+1:

PR #969, opened 2021-04-13
Remove package scope from default binary name

### LINKED ISSUE #963: Don't include scope for string-valued `bin`
We support both a hash and a string for the `bin` entry in `package.json`. When using a string, it's interpreted as the path to the binary, while the name for the binary is the name of the package. However, to match the behavior of the package managers, when the package is scoped, we should _not_ include the scope in the binary name, only the unscoped package name.

charlespierce:
Closes #963 

Info
-----
* When the `bin` field in `package.json` is a string (as opposed to a hash), that's interpreted as the _path_ to the binary and the package name is used as the name of the binary.
* However, when the package is scoped (e.g. `@scope/package-name`), then the binary should be named after the package _without_ the scope (`package-name` in this example).
* Including the scope causes problems when creating shims, as the `/` is interpreted as a filesystem directory separator.

Changes
-----
* Updated the logic that handles string-valued `bin` to strip off any scope when setting the name of the binary.

Tested
-----
* Added unit tests to confirm that removing the scope works as expected
* Manual testing to confirm that a scoped package with a string bin can be installed correctly, with the appropriate binary name.

### DIFF
--- a/crates/volta-core/src/tool/package/metadata.rs
+++ b/crates/volta-core/src/tool/package/metadata.rs
@@ -203,9 +203,10 @@ impl PackageManifest {
         // If the bin list contains only an empty string, that means `bin` was a string value,
         // rather than a map. In that case, to match `npm`s behavior, we use the name of the package
         // as the bin name.
+        // Note: For a scoped package, we should remove the scope and only use the package name
         if manifest.bin == [""] {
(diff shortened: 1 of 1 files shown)


dherman (approved):
Very clear and readable! 🚢

PR #972, opened 2021-04-16
Use fish_add_path instead of set -gx

domoritz:


chriskrycho:
Hi @domoritz, thanks for this PR! As the original implementor of our fish support, I thought of this immediately when `fish_add_path` was announced! Unfortunately, using `fish_add_path` requires being on fish 3.2.0+, which is only a month and a half old. Many of our consumers aren't able to upgrade to it yet. For example: anyone who's using Volta on a Linux setup without permissions to install arbitrary versions of external packages will need to wait for their system package manager to update to supporting this version. That's a *very* slow process; at least one distro still didn't have fish v3 support last time I checked in mid-2020!

That said, I think we can do this *conditionally*, by checking the fish version and using `fish_add_path` if we're on a version greater than 3.2.0. If you can update the PR to do it that way, that'd be grand and we'd love to land it!

domoritz:
Ahh good points. I don't have the cycles for this right now and one can always manually update to fish_add_path (which is what I already did). I'll close this pull request and maybe someone else feels inspired enough to make a better one.

chriskrycho:
Totally get that, and thank you again!

Issue #973, opened 2021-04-21
[Blocked] Run CI test suite on Apple Silicon devices


charlespierce:
Currently, our test suite for MacOS only executes in CI on x64 hardware. We should look to get our tests running on Apple Silicon hardware as well, to have maximum confidence that our changes work on all supported platforms.

Unfortunately, GitHub Actions (which we use for CI) currently doesn't have support for hosted runners on M1 machines.

PR #974, opened 2021-04-21
[Medium] Fetch Native Apple Silicon Node builds for Node 16+

### LINKED ISSUE #919: Fetch M1 Native Node builds where possible
Currently, the M1 Native version of Volta fetches the `x64` version of Node, which then runs through Rosetta 2. Once Node begins to ship prebuilt M1 Native binaries, we should fetch those where possible. If there isn't a native build available, we should still fall back to fetching the `x64` Node.

Currently blocked as Node is not yet publishing M1 Native builds and we don't yet know what the `files` identifier will be in https://nodejs.org/dist/index.json

charlespierce:
Closes #919 

Info
-----
* With the release of Node v16, there are now pre-built binaries available for Apple Silicon devices.
* Where possible, we should fetch and use those binaries, so that users get the best possible experience.
* However, for older versions of Node, we will need to continue to fetch the x64 binaries and rely on Rosetta.

Changes
-----
* Updated the constants section in `tool/node/mod.rs` to include the constant values for Apple Silicon architecture
* Ensured that the parsing of the `files` metadata in the Node index will check for both the native and x64 versions of Node.
* Customized `archive_filename` on Apple Silicon builds to check the Node major version and fetch either a native or x64 build of Node depending on whether it is greater than or equal to 16.

Tested
-----
* Our test suite currently doesn't execute on an Apple Silicon machine, so we don't have a way to automatically test these changes (opened #973 to track that issue separately, currently blocked on GitHub Actions support).
* Could someone with access to an M1 machine (cc @chriskrycho ?) run a test to confirm this properly fetches Node 16+ and 15-?

Notes
-----
* Per [this comment](https://github.com/nodejs/build/issues/2474#issuecomment-820312698), Node does not support Apple Silicon (even when building from source) prior to Node 15, and only provides pre-built binaries as of Node 16. So there shouldn't be any LTS releases of Node that have pre-built binaries but a major version less than 16.

### DIFF
--- a/crates/volta-core/src/tool/node/fetch.rs
+++ b/crates/volta-core/src/tool/node/fetch.rs
@@ -30,7 +30,7 @@ cfg_if! {
     }
 }
 
-fn npm_manifest_path(version: &str) -> PathBuf {
+fn npm_manifest_path(version: &Version) -> PathBuf {
     let mut manifest = PathBuf::from(Node::archive_basename(version));
 
     #[cfg(unix)]
@@ -46,7 +46,7 @@ fn npm_manifest_path(version: &str) -> PathBuf {
 pub fn fetch(version: &Version, hooks: Option<&ToolHooks<Node>>) -> Fallible<NodeVersion> {
     let home = volta_home()?;
     let node_dir = home.node_inventory_dir();
-    let cache_file = node_dir.join(Node::archive_filename(&version.to_string()));
+    let cache_file = node_dir.join(Node::archive_filename(version));
 
     let (archive, staging) = match load_cached_distro(&cache_file) {
         Some(archive) => {
@@ -107,7 +107,7 @@ fn unpack_archive(archive: Box<dyn Archive>, version: &Version) -> Fallible<Node
         })?;
 
     // Save the npm version number in the npm version file for this distro
-    let npm_package_json = temp.path().join(npm_manifest_path(&version_string));
+    let npm_package_json = temp.path().join(npm_manifest_path(version));
     let npm = Manifest::version(&npm_package_json)?;
     save_default_npm_version(&version, &npm)?;
 
@@ -115,14 +115,12 @@ fn unpack_archive(archive: Box<dyn Archive>, version: &Version) -> Fallible<Node
     ensure_containing_dir_exists(&dest)
         .with_context(|| ErrorKind::ContainingDirError { path: dest.clone() })?;
 
-    rename(
-        temp.path().join(Node::archive_basename(&version_string)),
-        &dest,
-    )
-    .with_context(|| ErrorKind::SetupToolImageError {
-        tool: "Node".into(),
-        version: version_string,
-        dir: dest.clone(),
+    rename(temp.path().join(Node::archive_basename(version)), &dest).with_context(|| {
+        ErrorKind::SetupToolImageError {
+            tool: "Node".into(),
+            version: version_string,
+            dir: dest.clone(),
+        }
     })?;
 
     progress.finish_and_clear();
@@ -151,8 +149,7 @@ fn load_cached_distro(file: &Path) -> Option<Box<dyn Archive>> {
 
 /// Determine the remote URL to download from, using the hooks if available
 fn determine_remote_url(version: &Version, hooks: Option<&ToolHooks<Node>>) -> Fallible<String> {
-    let version_str = version.to_string();
-    let distro_file_name = Node::archive_filename(&version_str);
+    let distro_file_name = Node::archive_filename(version);
     match hooks {
         Some(&ToolHooks {
             distro: Some(ref hook),
--- a/crates/volta-core/src/tool/node/metadata.rs
+++ b/crates/volta-core/src/tool/node/metadata.rs
@@ -1,6 +1,8 @@
 use std::collections::HashSet;
 
 use super::NODE_DISTRO_IDENTIFIER;
+#[cfg(all(target_os = "macos", target_arch = "aarch64"))]
+use super::NODE_DISTRO_IDENTIFIER_FALLBACK;
 use crate::version::{option_version_serde, version_serde};
 use semver::Version;
 use serde::{Deserialize, Deserializer};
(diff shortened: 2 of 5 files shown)


chriskrycho:
```
chris@Razorback:~/d/t/volta-test  v14.16.1
❯ volta pin node@15
success: pinned node@15.14.0 (with npm@7.7.6) in package.json

chris@Razorback:~/d/t/volta-test  v15.14.0
❯ file (volta which node)
/Users/chris/.volta/tools/image/node/15.14.0/bin/node: Mach-O 64-bit executable x86_64

chris@Razorback:~/d/t/volta-test  v15.14.0
❯ volta pin node@16
success: pinned node@16.0.0 (with npm@7.10.0) in package.json

chris@Razorback:~/d/t/volta-test  took 3s
❯ file (volta which node)
/Users/chris/.volta/tools/image/node/16.0.0/bin/node: Mach-O 64-bit executable arm64
```

🎉

chriskrycho (approved):
I have three comments:

1. This is awesome! And *wow* is Node running native on M1 *fast*.
2. Thank you for all the clear comments along the way!
3. The fact that this also passes around `Version` instead of `&str` is 🤩 

:shipit: !

alcuadrado:
Should this be working now? I'm using volta 1.0.4 and volta install node@16 installs an x86_64 version.

chriskrycho:
Hmm. It should, yes, and I’ve seen it fetch the right thing. 🤔 Can you provide further information on Volta itself, the Node installed, etc.?

alcuadrado:
It may be because I fetched node@16 before upgrading volta.

More info:

```
pato@parm:~% volta --version
1.0.4
pato@parm:~% volta install node@16
success: installed and set node@16.0.0 as default
pato@parm:~% file `which node`
/Users/pato/.volta/bin/node: Mach-O 64-bit executable x86_64
pato@parm:~% uname -a
Darwin parm 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 x86_64
```

alcuadrado:
Also, this didn't help:

```
pato@parm:~% rm -rf ~/.volta
pato@parm:~% curl https://get.volta.sh | bash
pato@parm:~% volta install node@16
pato@parm:~% file `which node`
/Users/pato/.volta/bin/node: Mach-O 64-bit executable x86_64
```

chriskrycho:
HMMMM. A couple more investigatory questions:

- what does <code>file \`which volta`</code> show?
- are you running in a terminal (emulator) which is itself running under Rosetta 2?

alcuadrado:
Np, happy to help.

```
pato@parm:~% file `which volta`
/Users/pato/.volta/bin/volta: Mach-O 64-bit executable x86_64
```

I'm running these commands using Alacritty, which launches zsh, which launches tmux, which launches zsh 😅

```
pato@parm:~% file `which alacritty`
/opt/homebrew/bin/alacritty: Mach-O 64-bit executable x86_64
pato@parm:~% file `which zsh`
/opt/homebrew/bin/zsh: Mach-O 64-bit executable arm64
pato@parm:~% file `which tmux`
/opt/homebrew/bin/tmux: Mach-O 64-bit executable arm64
```

charlespierce:
I'm not at all an expert in how Rosetta works, but I'm wondering if starting with Alacritty (which appears to be an x64 program) puts everything underneath it into "Compat mode", meaning even though `zsh` is a native binary, it runs it with some settings that emulate an x64 environment?

For some context, `which node` should always point to the node shim, which itself is a symlink to the `volta-shim` executable. Since that is installed by the Volta installer, it seems (from your previous testing) that doing `curl https://get.volta.sh | bash` is getting the x64 version instead of the native M1 version. The install script uses `uname -m` to detect the architecture [here](https://github.com/volta-cli/volta/blob/main/dev/unix/volta-install.sh#L140), so I'd be interested to see what the output of that is when you run it in the same terminal environment (alacritty).

Another thing worth checking, since the install involves piping into bash, would be the output of ``file `which bash` ``, to make sure there isn't some weird x64 version of Bash that's breaking the install script.

alcuadrado:
I know almost nothing about Rosetta tbh.

I found some interesting things:


This is run under zsh, which according to activity monitor is running without rosetta:

```
pato@parm:hardhat-ignition (main ✔)% which bash
/bin/bash
pato@parm:hardhat-ignition (main ✔)% file /bin/bash
/bin/bash: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
pato@parm:hardhat-ignition (main ✔)% uname -m
x86_64
```

I opened Terminal.app, running zsh without tmux (both are aarm64), and this worked:

```
pato@parm:~% rm -rf ~/.volta
pato@parm:~% curl https://get.volta.sh | bash
pato@parm:~% volta install node@16
success: installed and set node@16.0.0 (with npm@7.10.0) as default
pato@parm:~%  file `which node`
/Users/pato/.volta/bin/node: Mach-O 64-bit executable arm64
```

Now on Alacritty:

```
pato@parm:~% volta install node@15
success: installed and set node@15.14.0 (with npm@7.7.6) as default
pato@parm:~% file `which node`
/Users/pato/.volta/bin/node: Mach-O 64-bit executable arm64
```

charlespierce:
Okay, great! Yeah, it seems like something in the setup (my guess is alacritty at the start, but I don't have enough expertise to know for sure) was setting things up in a way that misled the Volta installer, causing it to fetch the x64 Mac version, instead of the native one. Now that you _have_ the proper native one, however, Volta should work properly even in the Alacritty environment (since it uses compile-time flags to determine the architecture, rather than detecting it at runtime). So I believe you should be good to go.

One last thing, as I mentioned `which node` should always be pointing at the Volta shim file, which will always be `arm64`, regardless of which Node version is actually selected. If you want to introspect the actual Node binary that will be used, you can take advantage of `volta which` (a command designed to "unwrap" the shims and point you to the actual file that will execute): ``file `volta which node` `` should give you the info about the actual binary. Since it looks like things are set up properly now, I would expect that one to also be `arm64`.

alcuadrado:
Oh, I missed that part! Yeah, it's `arm64` for node@16, and x86_64 for @15.

PR #981, opened 2021-04-30
Ignore non-json files when listing all package configs

### LINKED ISSUE #980: `volta list` - Limit parsing package config files to those that have a `json` extension
Hello 👋 

I'm on Volta 1.04, and when I executed `volta list`, I got:
```
➜  ~ volta list     
[verbose] Found default configuration at '/Users/glenn/.volta/tools/user/platform.json'
error: Could not parse package configuration file.

Please rerun the command that triggered this error with the environment
variable `VOLTA_LOGLEVEL` set to `debug` and open an issue at
https://github.com/volta-cli/volta/issues with the details!
[verbose] Error cause: expected value at line 1 column 1
Error details written to /Users/glenn/.volta/log/volta-error-2021-04-28_13_56_53.370.log
➜  ~ volta --version
1.0.4
➜  ~ 
```

I've set the `VOLTA_LOGLEVEL` to `debug` and also checked the log but it didn't give me more info other than **"Error cause: expected value at line 1 column 1"**. Any insights? 🤔 

Thank you.

charlespierce:
Closes #980 

Info
-----
* Currently, the `inventory::package_configs` function reads all of the files in the package config directory and attempts to parse them as package configs.
* However, sometimes the OS will automatically generate hidden files in a directory (e.g. `.DS_Store` on macOS).
* Those auto-generated files will not be valid package configs, but they shouldn't cause `volta list` to fail, instead they should be ignored.

Changes
-----
* Updated the directory walk within `package_configs` to ignore both directories _and_ any file that doesn't have a `.json` extension.

Tested
-----
* Locally tested to confirm that adding an empty dotfile into the package config directory does _not_ cause `volta list` to fail.

### DIFF
--- a/crates/volta-core/src/inventory.rs
+++ b/crates/volta-core/src/inventory.rs
@@ -2,6 +2,7 @@
 //! of available tool versions.
 
 use std::collections::BTreeSet;
+use std::ffi::OsStr;
 use std::path::Path;
 
 use crate::error::{Context, ErrorKind, Fallible};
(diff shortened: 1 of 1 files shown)


chriskrycho (approved):
🎉 ⚡

PR #990, opened 2021-05-08
Don't update the user's shell config files if VOLTA_HOME and PATH already contains what we need.

dsully:
This will prevent changes when these settings are defined in a different, but compatible way to what volta does.

### DIFF
--- a/src/command/setup.rs
+++ b/src/command/setup.rs
@@ -45,11 +45,14 @@ mod os {
 
         // Don't update the user's shell config files if VOLTA_HOME and PATH already contain what we need.
         let home_in_path = match env::var_os("PATH") {
-            Some(paths) => env::split_paths(&paths).find(|p| p == home.shim_dir()),
-            None => None
+            Some(paths) => env::split_paths(&paths).find(|p| p == home.root()),
+            None => None,
         };
 
         if env::var_os("VOLTA_HOME").is_some() && home_in_path.is_some() {
+            debug!(
+                "{} skipping dot-file modification as VOLTA_HOME is set, and included in the PATH."
+            );
             return Ok(());
         }
 
--- a/src/command/setup.rs
+++ b/src/command/setup.rs
@@ -43,6 +43,16 @@ mod os {
         let home = volta_home()?;
         let formatted_home = format_home(home.root());
 
+        // Don't update the user's shell config files if VOLTA_HOME and PATH already contain what we need.
+        let home_in_path = match env::var_os("PATH") {
+            Some(paths) => env::split_paths(&paths).find(|p| p == home.root()),
+            None => None
+        };
+
+        if env::var_os("VOLTA_HOME").is_some() && home_in_path.is_some() {
+            return Ok(());
+        }
+
         debug!("Searching for profiles to update");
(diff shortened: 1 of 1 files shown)


src/command/setup.rs
```diff
@@ -43,6 +43,16 @@ mod os {
         let home = volta_home()?;
         let formatted_home = format_home(home.root());
 
+        // Don't update the user's shell config files if VOLTA_HOME and PATH already contain what we need.
+        let home_in_path = match env::var_os("PATH") {
+            Some(paths) => env::split_paths(&paths).find(|p| p == home.root()),
```
charlespierce, line 48: I believe this check will almost always fail, since we don't add `~/.volta` to the PATH, but rather `~/.volta/bin`.

```suggestion
            Some(paths) => env::split_paths(&paths).find(|p| p == home.shim_dir()),
```

src/command/setup.rs
```diff
@@ -43,6 +43,16 @@ mod os {
...
+            None => None
+        };
+
+        if env::var_os("VOLTA_HOME").is_some() && home_in_path.is_some() {
+            return Ok(());
```
charlespierce, line 53: Can you add a `debug!` message here before returning, so it will show up in the verbose output why no changes were made?

charlespierce (changes requested):
Thanks @dsully! Looks mostly good, just a couple of small tweaks to make sure things work as expected and are easier to debug later on.

Also, would you mind creating a follow-on issue to add a `--force` or similar flag? It's not necessary for this PR, but would be good to track the need for something like that in case we want to have the ability to force it through even given the environment appears to be properly set up.

charlespierce:
Looks like the changes need to have `cargo fmt` run on them. Also the Windows builds are failing for some unrelated reason, it appears, I'll investigate that separately.

dsully:
Done & done!

src/command/setup.rs
```diff
@@ -43,6 +43,19 @@ mod os {
...
 
+        // Don't update the user's shell config files if VOLTA_HOME and PATH already contain what we need.
+        let home_in_path = match env::var_os("PATH") {
+            Some(paths) => env::split_paths(&paths).find(|p| p == home.root()),
+            None => None,
+        };
+
+        if env::var_os("VOLTA_HOME").is_some() && home_in_path.is_some() {
+            debug!(
+                "{} skipping dot-file modification as VOLTA_HOME is set, and included in the PATH."
```
charlespierce, line 54: It doesn't appear that there's a parameter, so I don't believe we need the `{} ` at the beginning of this (and the formatting will possibly switch with removing it from the string).

charlespierce (changes requested):
One more small change (that may need `cargo fmt` run again afterwards).

src/command/setup.rs
charlespierce: Sorry to keep going round-and-round, but it looks like we may have lost the `shim_dir` change with the follow-up that added the `debug!` statement below.

charlespierce (approved):
LGTM, thanks! The CI failures are unrelated and fixed in #993, once that's merged I'll merge this and then confirm that everything passes post-merge.

PR #1024, opened 2021-09-07
Pass yarn calls through to allow for Corepack where available

### LINKED ISSUE #987: `volta install corepack` doesn't provide corepack's `yarn`
I installed https://github.com/nodejs/corepack.
Normally corepack provide yarn binary, but using volta, yarn is linked to `volta-shim`.
Volta has not compatibility for corepack?

charlespierce:
Closes #987 

Info
-----
* Node now ships with Corepack, which is an alternative way to manage your package manager using a similar shim method to Volta.
* However, we currently do a check internally when running `Yarn` to ensure that there is a Volta-defined version of Yarn available.
* This check means that we don't properly pass through to the Corepack `yarn` executable in cases where Volta is managing Node but Corepack is handling Yarn.
* At the same time, we already have the ability to provide a custom error message in the case that launching a tool fails.
* For better interoperability with Corepack, we should attempt to pass through to `Yarn` in the Node directory and only show our `NoProjectYarn` / `NoDefaultYarn` errors if that isn't available.

Changes
-----
* Updated the logic for running Yarn to determine the appropriate failure `ErrorKind` but use that as the execution context, rather than failing before executing.
* This will allow `Yarn` to run (using the platform PATH which includes the Node binary directory) even when Yarn isn't set in Volta.
* If Corepack is available and set up, it will be called by running `yarn` with the Node binary directory.
* If not, then the execution will fail and show the appropriate error message.

Notes
-----
* Still working through the specifics of how to make the failure case work: With the Volta bin directory still on the PATH, the call will actually always succeed, even if the subsequent call ultimately fails (when it detects the recursion environment variable and tries to call out to the regular system). We need some way to pass the error kind through to that subsequent call, or otherwise detect a possible failure ahead of time.

### DIFF
--- a/crates/volta-core/src/run/yarn.rs
+++ b/crates/volta-core/src/run/yarn.rs
@@ -43,13 +43,13 @@ pub(super) fn execution_context(
 ) -> Fallible<(OsString, ErrorKind)> {
     match platform {
         Some(plat) => {
-            validate_platform_yarn(&plat)?;
+            let on_failure = platform_yarn_error(&plat);
 
             let image = plat.checkout(session)?;
             let path = image.path()?;
             debug_active_image(&image);
 
-            Ok((path, ErrorKind::BinaryExecError))
+            Ok((path, on_failure))
         }
         None => {
             let path = System::path()?;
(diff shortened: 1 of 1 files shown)


avarun42:
any update here?

therynamo:
Would love this feature if there is still appetite for it! It would really help in setting up CI runners so we don't have to worry about `yarn` mirrors in the hooks.json and we can just let `corepack` do all of the work for us.

mikeduminy:
Another consideration is that corepack can be installed globally

PR #1052, opened 2021-11-23
Prebuild binaries for Linux ARM environments

jensmeindertsma:
This pull requests implements support for prebuilding binaries for the Linux ARM environment.

For now this is still very dirty and quick, and I'm open to change requests. I'm just looking to get something working first.

This pull request is still a work in progress.

### To-do
- [ ] Update GitHub Actions workflow
- [x] Update the install script


jensmeindertsma:
@charlespierce any chance you could approve the workflows for this pull request so I can see whether my implementation is correct?

jensmeindertsma:
I think I'll need some more direction on how to approach this 🤔

charlespierce:
Yeah, it's a bit complicated. For the OpenSSL linux builds, we build the different OpenSSL versions in a docker container and then link Volta against the output of those builds (to make sure we are linking against the appropriate versions). In the case of cross-compilation, we likely have to cross-compile OpenSSL as well in order for the dynamic linking to work.

Unfortunately, I'm far from an expert in building OpenSSL, so I don't know if it's possible to do a straightforward cross-compile for ARM Linux. And if it is, I'm not sure how to do it 🙁  It's possible there are docs / references about how to use GitHub actions to compile for ARM Linux.

jensmeindertsma:
@charlespierce I now made it so we cross-compile OpenSSL. Could you approve the workflow? And do you have any idea why I'm getting a `Permission denied` error during the workflow run?

jensmeindertsma:
Going to get a Linux ARM VM up and running on my system to make this process somewhat easier.

jensmeindertsma:
Still getting a "Permission denied". Anyone have any pointers by chance?

jensmeindertsma:
```
- name: Set up docker image
        run: docker build -t volta .
``` What Dockerfile is being built here? The one in `/ci/docker` doesn't seem correct for regular Linux builds?

jensmeindertsma:
So I managed to fix the permissions, now I'd like to run the workflows again to see whether it would work. Any chance you could let them run a few more times @charlespierce ?

jensmeindertsma:
Actually, we need to compile for both musl and gnu. I personally only need musl, but should I add both?

jensmeindertsma:
Actually, how is this currently handled? Is Alpine Linux not supported? It looks like we're just building with the default target on Ubuntu, which builds for GNU.

chriskrycho:
I’ve approved the CI runs, at least. I’m even less of an expert in this corner of things than @charlespierce, though, so will be unable to advise beyond that. 😅

PR #1057, opened 2021-12-01
Fix clippy warnings about unneeded borrows

charlespierce:
Info
-----
* Bumping the Rust version came with some new clippy warnings about unnecessary borrows.
* Those warnings could cause CI to fail since CI always builds on the latest stable.

Changes
-----
* Removed the unnecessary borrows based on Clippy feedback.

Tested
-----
* There should be no functional changes, since the warning is about redundant referencing.
* All tests pass.


chriskrycho (approved):

chriskrycho:
Oh, this is neat.

PR #1059, opened 2021-12-01
Set up a Dependabot config for the project

chriskrycho:
I *think* this is the right setup, and should more or less work. We can tweak the frequency later; I set it to 5/day for the top level and 3/day for the other crates initially just so we don't all die of Dependabot PRs. 😂

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,71 @@
+version: 2
+updates:
+  - directory: /
+    open-pull-requests-limit: 5
+    package-ecosystem: "cargo"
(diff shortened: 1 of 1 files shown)


charlespierce (approved):
I don't have experience with setting up dependabot, but LGTM!

chriskrycho:
I have set it up, but I have never set it up for a project with nested config, so here’s hoping I got it right. 😂

chriskrycho:
Spoilers: it was slightly wrong. #1061 should fix it. 🙃

PR #1088, opened 2021-12-03
Bump 'regex' version to latest

charlespierce:
Info
-----
* The dependabot update to `regex` in #1070 is failing to build for an unknown reason (possibly due to not updating `regex` everywhere?).

Changes
-----
* Bumped `regex` in both `volta-core` and `validate-npm-package-name`

Tested
-----
* Compiling succeeds and all tests pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.10"
+version = "0.7.18"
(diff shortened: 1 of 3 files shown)


charlespierce:
Hmm, doesn't seem like this is working either. I'm going to try bumping `os_info`, which appears to be where the compilation failure is happening, and see if that fixes these updates (`regex` and `console`).

Issue #145, opened 2018-09-10
Acceptance tests should use a real http server for node & yarn archive files


mikrostew:
Right now the requests are being mocked, but the returned file is not a valid archive. This should be updated to use an actual http file server.

charlespierce:
As mentioned in #1099, we should revisit this in light of `mockito` deprecating the `SERVER_URL` constant, as we will need some way to inject the appropriate server URL into the app when running tests.

One solution (not deeply investigated) may be to generate a `hooks.json` file for the tests that includes the `server_url` automatically.

PR #1099, opened 2021-12-03
Bump mockito and suppress deprecation warnings

charlespierce:
Info
-----
* `mockito` appears to be moving away from having a static server address defined in a constant.
* For the time being, the constant still exists and works, however it is marked as deprecated.
* Unfortunately, the suggested `server_url` method as a way forward doesn't work due to how our tests are structured:
    * `server_url` attempts to start the server if it isn't running, then returns the address.
    * This relies on the assumption that the call to `server_url` in the code under test and the creation of the `Mock` objects happen in the same process, since the server status is shared memory.
    * In our case, our acceptance tests call the `volta` and shim binaries as a child process, in order to inspect their return codes and output.
    * Using `server_url` in the `volta_core` area results in the tests setting up a server on the expected local port, and then the app itself trying to set up _another_ server and necessarily picking a different port.
* Since the constants still exist and work, we should prevent unnecessary warning spam on every test run (and CI run!) while we investigate a migration path.
* Long-term, we should come up with a way to inject the server URL into the app, so that we can use the value from the test server (see #145).
    * One solution (which may work almost directly out-of-the-box) would be to generate a `hooks.json` file that includes the `server_url()` as a `prefix` hook for every relevant type of hook.

Changes
-----
* Bumped `mockito` version in the main app.
* Added `#[allow(deprecated)]` to the calls to `SERVER_URL` within the executables themselves.
* Updated calls within the tests to actually use the suggested `server_url()` function.

Tested
-----
* All tests pass with the mocks working as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,12 +36,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "approx"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08abcc3b4e9339e33a3d0a5ed15d84a687350c05689d825e0f6655eef9e76a94"
-
-[[package]]
 name = "archive"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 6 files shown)


chriskrycho (approved):
Funnnnn

Issue #1006, opened 2021-06-24
Please add support for arm64


mainendra:

charlespierce:
Hi @mainendra, can you clarify a bit what you mean by "support for arm64"? Are you referring to Arm64 Linux? Currently we have support in the code itself for that Architecture, but we don't create pre-built binaries. So you should be able to build from source, but the default installer doesn't have binaries to use.

If so, we can definitely investigate creating those binaries as part of our release, however I'm not familiar with the range of distros that support Arm64, so it may take a bit of work to make sure that we compile it correctly to work across a large range of distros.

If you're referring to the M1 MacOS Arm64, that should work out of the box already. If you're seeing problems, then that's a bug.

mainendra:
@charlespierce - thanks. I'm using ubuntu on raspberry pi and default installation throwing an error that arm64 is not supported. But I'll try local build and see if it works. Yes, this is not a bug but feature request. Thanks again.

sirreal:
I ran into this recently. I was able to build Volta from source on a Raspberry PI. The only difficulty I had was that the `os_info` dependency would not build. I solved this by upgrading to the latest version of the next major release `2.0.8`. After that, I was able to install and use Volta on a Raspberry Pi.

@charlespierce is there a reason the pre-built binaries are not provided? I'd be happy to help if it's a question of setting things up.

charlespierce:
Hi @sirreal, the primary reason we don't have pre-built binaries for Linux arm64 is just bandwidth: we haven't gotten around to it. We'd love to have a contribution to add that, if you're interested in taking that on! One thing to consider would be the different OpenSSL versions: Currently we have to create 3 different Linux builds (for x64), since there are different versions of OpenSSL to link against (1.0.x, 1.1.x, and the custom one used by RHEL / CentOS), so there may need to be some testing around that for building the arm64 binaries.

Beyond that, the main steps would be including those builds in the GitHub actions (along side the other production builds) and updating the installer script to support that architecture.

chriskrycho:
I think this is resolved in our v2.0.0 release, @charlespierce?

charlespierce:
Yes, this is resolved now (after making a fix to the installer 😅). The `get.volta.sh` installer will correctly detect Linux ARM and install that version.

PR #1130, opened 2022-01-10
feat: arm64 linux build

vanstinator:
I've been poking on this on and off since early November and didn't realize that in the meantime someone else had opened https://github.com/volta-cli/volta/pull/1052

I've taken a look at that approach and personally prefer the approach I've taken (we build our internal CI arm64 docker images this way at my day job), and I have working builds already that you can pull to your own machine to test and play with. I haven't exhaustively tested the binary yet as my primary focus was on reducing the time it takes to build the binary, but it should work well enough at the moment for testing. I'll do a full pass in the coming days.

#### Why do this at all?
I'm on an M1 MBP. It's a lovely machine. But at my day job we use volta inside debian docker images. Because volta only supports x86_amd64 on Linux I have to run an x86 Docker container locally. It's painfully slow virtualizing a Docker image this way. Native arm64 builds of volta would allow us to build both arm64 and x86 versions of our docker images for local development.

On to the good bits: I've updated the existing linux flow with an additional matrix dimension for architecture. The benefit here is keeping a consistent build process for every linux release. We can leverage Docker's `buildx` and `qemu` actions to build both `x86` and `arm64` binaries in parallel. I needed to bump to CentOS 8 for this to work as 6 does not have an arm64 linux image available. The big caveat: building arm64 binaries on a qemu layer is _very very_ slow. Without any caching enhancements it takes over 2 hours to build. 🤯.

On that note I've refactored how OpenSSL builds work. Rather than build them on every single build I've elected to move the build process inside the docker image, and then cache that docker image. On repeated builds the image is downloaded from the cache, restored into the docker runtime on the GHA node, and you only need to spend the time to build volta itself without spending CPU cycles rebuilding OpenSSL every time.

Additionally I'm caching the rust `target` directory between builds so we can reduce the amount of time spent on building dependencies. That was another source that took exponentially longer on arm64 VMs.

Related https://github.com/volta-cli/volta/issues/1006

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,9 +15,20 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Set up docker image
-        run: docker build -t volta .
-        working-directory: ./ci/docker
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          context: .
+          file: ./ci/docker/Dockerfile.centos6
+          load: true
+          push: false
+          tags: volta
+          target: base
       - name: Compile and package Volta
         run: docker run --volume ${PWD}:/root/workspace --workdir /root/workspace --rm --init --tty volta /root/workspace/ci/build-and-package.sh volta-centos
       - name: Confirm correct OpenSSL Version
@@ -33,34 +44,86 @@ jobs:
   linux:
     strategy:
       matrix:
-        openssl:
-          - 1_1_0
-          - 1_0_1
-    name: Build - OpenSSL ${{ matrix.openssl }}
+        arch: [x86_64, arm64]
+        openssl: [1_0_1, 1_1_1]
+        os: [linux]
+        include:
+        - arch: x86_64
+          dockerfile: Dockerfile.centos6
+          # Docker's BuildX platform for x86_64 is `amd64`. But `uname -m` on linux returns x86_64.
+          # So we'll officially designate this binary as `x86_64` but map it here to `amd64` so that
+          # we build it on the correct docker arch
+          dockerArch: amd64
+        - arch: arm64
+          dockerfile: Dockerfile.centos8
+          dockerArch: arm64
+
+    name: Build - OpenSSL ${{ matrix.openssl }} - Arch ${{ matrix.arch }}
     runs-on: ubuntu-latest
     steps:
+      
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Check out OpenSSL
-        uses: actions/checkout@v2
+
+      # Not stricly speaking necessary for x86 builds, but the QEMU arm64 builds take hours
+      - name: Rust build cache
+        id: cache-openssl
+        uses: actions/cache@v2
         with:
-          repository: openssl/openssl
-          ref: OpenSSL_${{ matrix.openssl }}-stable
-          path: openssl
-      - name: Set up docker image
-        run: docker build -t volta .
-        working-directory: ./ci/docker
+          path: |
+            .cargo/registry
+            .cargo/git
+            target/
+          key: ${{ runner.os }}-${{ matrix.openssl }}-${{ matrix.arch }}-target-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.openssl }}-${{ matrix.arch }}-target-
+
+      - name: Clean .git
+        run: |
+          rm -rf ./openssl/.git
+
+      - name: Debug
+        run: |
+          echo $(ls -la)
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      
+      - name: Build and push
+        uses: docker/build-push-action@v2
+        with:
+          build-args: |
+            OPENSSL_VERSION=${{ matrix.openssl }}
+          cache-from: type=gha,scope=${{ matrix.openssl }}-${{ matrix.arch }}
+          cache-to: type=gha,scope=${{ matrix.openssl }}-${{ matrix.arch }},mode=max
+          context: .
+          file: ./ci/docker/${{ matrix.dockerfile }}
+          load: true
+          platforms: linux/${{ matrix.dockerArch }}
+          push: false
+          tags: volta-${{ matrix.openssl }}-${{ matrix.arch }}
+          target: openssl-${{ matrix.openssl }}
+
       - name: Compile and package OpenSSL & Volta
-        run: docker run --volume ${PWD}:/root/workspace --workdir /root/workspace --rm --init --tty volta /root/workspace/ci/build-with-openssl.sh volta-openssl-${{ matrix.openssl }}
+        run: docker run --volume ${PWD}:/root/workspace --platform linux/${{ matrix.dockerArch }} --workdir /root/workspace --rm --init --tty volta-${{ matrix.openssl }}-${{ matrix.arch }} /root/workspace/ci/build-with-openssl.sh volta-openssl-${{ matrix.openssl }}-${{ matrix.arch }}
+      
+      - name: Set .cargo/ ownership to current user
+        run: sudo chown -R $(id -u):$(id -g) .cargo/
+
       - name: Confirm OpenSSL Version
         run: |
           objdump -p target/release/volta
           readelf -d target/release/volta
+
       - name: Upload release artifact
         uses: actions/upload-artifact@v2
         with:
           name: linux-openssl-${{ matrix.openssl }}
-          path: target/release/volta-openssl-${{ matrix.openssl }}.tar.gz
+          path: target/release/volta-openssl-${{ matrix.openssl }}-${{ matrix.arch }}.tar.gz
 
   macos:
     name: Build - MacOS
@@ -180,15 +243,25 @@ jobs:
         with:
           name: linux-centos
           path: release
-      - name: Fetch OpenSSL 1.0.* artifact
+      - name: Fetch OpenSSL 1.0.*-x86_64 artifact
         uses: actions/download-artifact@v2
         with:
-          name: linux-openssl-1_0_1
+          name: linux-openssl-1_0_1-x86_64
           path: release
-      - name: Fetch OpenSSL 1.1.* artifact
+      - name: Fetch OpenSSL 1.0.*-arm64 artifact
         uses: actions/download-artifact@v2
         with:
-          name: linux-openssl-1_1_0
+          name: linux-openssl-1_0_1-arm64
+          path: release
+      - name: Fetch OpenSSL 1.1.*-x86_64 artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: linux-openssl-1_1_0-x86_64
+          path: release
+      - name: Fetch OpenSSL 1.1.*-arm64 artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: linux-openssl-1_1_0-arm64
           path: release
       - name: Fetch MacOS artifact
         uses: actions/download-artifact@v2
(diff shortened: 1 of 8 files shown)


chriskrycho:
Thanks for opening this, @vanstinator! I've approved the CI run, and I will take a brief look at the updates, but will mostly defer here to @charlespierce, who has much better expertise on especially the OpenSSL part of our build pipeline.

vanstinator:
Thanks! There's definitely some rough edges in here yet. But I wanted to get some feedback before going too far down the polishing rabbit hole 😄

charlespierce:
This is awesome @vanstinator! I'm about as far from a Docker expert as you could get, I definitely couldn't figure out how to get an ARM64 build to work :smile:

I haven't dug too deep into the specifics, but my only concern is with switching to CentOS 8. It's been a while since we implemented the GH Actions builds, but I know there were compatibility issues with `glibc` when we built on more modern images and tried to run on older Linux distros. This was particularly an issue at LinkedIn for internal use at the time, though it also applied to other Linux distros.

Would it be possible to have the existing builds happen on the existing CentOS 6 image, while the ARM64 builds start from the CentOS 8 one?

dev/unix/volta-install.sh
```diff
@@ -134,7 +134,11 @@ parse_os_info() {
         return "$exit_code"
       fi
 
-      echo "linux-openssl-$parsed_version"
+      if [ "$(uname -m)" == "arm64" ]; then
+        echo "linux-openssl-$parsed_version-arm64"
+      else
+        echo "linux-openssl-$parsed_version-amd64"
```
vanstinator, line 140: There _shouldn't_ be an issue with changing the binary name here, right? We could consider following the convention you established with `macOS` where only the arm binary name contains the arch. But the mac case is probably special in that the x86 days are numbered. I'd expect arm64 and amd64 binaries to co-exist in the linux space pretty much indefinitely, so keeping a consistent `linx-openssl-platform` binary name seems like a reasonable approach.

And it makes it trivially easy to scale to something like armv8 for raspberry pi devices without further confusing the naming conventions.

ci/docker/Dockerfile
```diff
@@ -1,13 +1,26 @@
...
-RUN yum -y install gcc curl openssl openssl-devel ca-certificates tar perl perl-Module-Load-Conditional && yum clean all
+RUN yum -y install gcc git curl openssl openssl-devel ca-certificates tar perl perl-Module-Load-Conditional && yum clean all
 
 # Install Rust
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable
 ENV PATH="/root/.cargo/bin:${PATH}"
+
+FROM base as openssl
+
+ENV CARGO_HOME=/root/workspace/.cargo
```
vanstinator, line 17: We set this here so cargo operates from the mapped volume on the host. That's the magic that lets us cache the `target` directory in that GHA cache action.

.github/workflows/release.yml
```diff
@@ -180,16 +232,27 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: linux-openssl-1_0_1-amd64
+          path: release
+      - name: Fetch OpenSSL 1.0.*-arm64 artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: linux-openssl-1_0_1-arm64
+          path: release
+      - name: Fetch OpenSSL 1.1.*-amd64 artifact
```
vanstinator, line 245: I'll likely refactor this to use a matrix as well.

.github/workflows/release.yml
```diff
@@ -33,34 +43,76 @@ jobs:
...
+          push: false
+          tags: volta-${{ matrix.openssl }}-${{ matrix.arch }}
+          target: openssl-${{ matrix.openssl }}
+
       - name: Compile and package OpenSSL & Volta
-        run: docker run --volume ${PWD}:/root/workspace --workdir /root/workspace --rm --init --tty volta /root/workspace/ci/build-with-openssl.sh volta-openssl-${{ matrix.openssl }}
+        run: docker run --volume ${PWD}:/root/workspace --platform linux/${{ matrix.arch }} --workdir /root/workspace --rm --init --tty volta-${{ matrix.openssl }}-${{ matrix.arch }} /root/workspace/ci/build-with-openssl.sh volta-openssl-${{ matrix.openssl }}-${{ matrix.arch }}
+      
+      - name: Set .cargo/ ownership to current user
+        run: sudo chown -R $(id -u):$(id -g) .cargo/
```
vanstinator, line 103: I initially attempted to clone the GHA node user/group info in the Dockerfile so we didn't need to do this. But that proved to be quite a few moving pieces. So instead I decided to just `chown` the directory when the build was done so the cache action can access it and cache the artifacts.

.github/workflows/release.yml
```diff
@@ -33,34 +43,76 @@ jobs:
...
-          - 1_1_0
-          - 1_0_1
-    name: Build - OpenSSL ${{ matrix.openssl }}
+        arch: [amd64, arm64]
+        openssl: [1_0_1, 1_1_1]
+    
+    name: Build - OpenSSL ${{ matrix.openssl }} - Arch ${{ matrix.arch }}
     runs-on: ubuntu-latest
     steps:
+
```
vanstinator, line 52: Personally I find these files are much easier to read and grok if there's a newline between each step. I only modified the areas I was working on but I'm happy to either add this to all the jobs or remove it from here before the PR merges.

vanstinator:
> I haven't dug too deep into the specifics, but my only concern is with switching to CentOS 8. It's been a while since we implemented the GH Actions builds, but I know there were compatibility issues with glibc when we built on more modern images and tried to run on older Linux distros. This was particularly an issue at LinkedIn for internal use at the time, though it also applied to other Linux distros.

Is there a GHI trail somewhere for some context?

> Would it be possible to have the existing builds happen on the existing CentOS 6 image, while the ARM64 builds start from the CentOS 8 one?

It'd be fairly trivial to keep the existing centos 6 Dockerfile for the CentOS build and the amd64 builds while spinning up a new Dockerfile for the arm64 builds. If that's going to happen I'd be tempted to suggest switching to one of the ubuntu-lts base images, but that's probably awkward if we're still keeping the CentOS 6 image. Probably cleaner to keep a consistent-ish OS base where possible.

vanstinator:
@charlespierce I've cleaned a few things up, fixed the tests that were failing, and implemented the requested changes to the centos versions. This should be good for a review!

ci/docker/Dockerfile.centos8
```diff
@@ -0,0 +1,26 @@
+FROM centos:8 as base
+
+# CentOS 6 packages are no longer hosted on the main repository, instead they are in the CentOS Vault
+# RUN sed -i 's/^mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Base.repo && \
+#     sed -i 's/#baseurl=http:\/\/mirror.centos.org\/centos\/$releasever/baseurl=http:\/\/vault.centos.org\/6.10/g' /etc/yum.repos.d/CentOS-Base.repo
```
charlespierce, line 5: I think we can remove these from the file completely, since they only apply to CentOS 6

dev/unix/volta-install.sh
```diff
@@ -133,11 +134,10 @@ parse_os_info() {
       if [ "$exit_code" != 0 ]; then
         return "$exit_code"
       fi
-
-      echo "linux-openssl-$parsed_version"
+      echo "linux-openssl-$parsed_version-$arch_str"
```
charlespierce, line 137: For backwards compatibility, can we handle the `$arch_str` in the same way as the mac case? Where `x86_64` is unadorned, while the `arm64` one gets an extra suffix? That would make sure that the install script still works for installing older versions (where the existing files don't have the `-x86_64` suffix). While I don't think it's super common, we do technically support a user passing `--version 1.0.4` to the script to get an older versions of Volta.

.github/workflows/release.yml
```diff
@@ -15,9 +15,20 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Set up docker image
-        run: docker build -t volta .
-        working-directory: ./ci/docker
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build and push
```
charlespierce, line 21: For my own understanding, this is basically the same as running `docker build ...`, except that it also caches the built docker image so that future runs can just pull it down, rather than having to re-build?

.github/workflows/release.yml
charlespierce: I think updating the areas you touched is reasonable, we can go through the others at some point in the future.

.github/workflows/release.yml
```diff
@@ -33,34 +44,86 @@ jobs:
...
+    name: Build - OpenSSL ${{ matrix.openssl }} - Arch ${{ matrix.arch }}
     runs-on: ubuntu-latest
     steps:
+      
       - name: Check out code
         uses: actions/checkout@v2
-      - name: Check out OpenSSL
-        uses: actions/checkout@v2
+
+      # Not stricly speaking necessary for x86 builds, but the QEMU arm64 builds take hours
```
charlespierce, line 68: Given that the arm64 builds take so long, is there a way we can lock them in the same way as the final `Publish release` job, so that they only run when we're actually publishing a new version? My worry is that currently this build is part of the CI check for PRs and I don't want to run into a spot where every PR is gated on a build that takes ~2 hours.

charlespierce (changes requested):
Sorry for the delay, been busy lately with unrelated work. This looks really nice, and all makes sense (even with my limited understanding of docker)!

jensmeindertsma:
I'm sorry, I totally forgot about #1052. I am really thankful @vanstinator took the time to implement this. I think it would be incredibly valuable to add `aarch64` support to this PR, such that M1 Mac users can use Volta inside Alpine containers. Would you be willing to collaborate on that @vanstinator? Am I missing any steps here that would be necessary for Alpine support (x86_64, arm64, aarch64 are the platforms I need)

.github/workflows/release.yml
jensmeindertsma: Yes, and it also supports other platforms than the host platform

jensmeindertsma:
Currently if you use the following Dockerfile
```dockerfile
FROM alpine
RUN  apk update && apk add curl bash openssl
RUN  curl https://get.volta.sh | bash 
RUN  volta --version
``` And run `docker build .` on an M1 machine you get
```
Error: Sorry! Volta currently only provides pre-built binaries for x86_64 architectures.
``` Which is caused by the lack of `aarch64` builds for Volta on Linux
If you use a compatibility layer for `x86_64` ( `FROM --platform=linux/amd64 alpine` ), you get
```
Installing latest version of Volta (1.0.5)
...
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
``` Which is due to the lack of MUSL builds for Volta.

I think it would be awesome if we could merge this PR and then start on fixing both these problems. I'm more than happy to take the lead, if you could give me some direction on MUSL builds and how to build aarch64 in Docker QEMU. Would it be best to first merge this PR @charlespierce ?

jensmeindertsma:
Just to clarify ( as I'm sure you're aware ), running `uname -m` on my machine yields `arm64`, but inside a Alpine container ( Debian also ) it returns `aarch64`. Do we need to adjust `volta-install.sh` to make things work for M1 machines just like the once you have @vanstinator ?

jensmeindertsma:
Yeah, I think the main oversight here was that `uname -m` will yield `aarch64` on Linux, so you'll want to use `aarch64` instead everywhere you've written `arm64`. I'll fix this.

vanstinator:
I've been super busy but I'll have some time this week to get this over the finish line.

> Yeah, I think the main oversight here was that uname -m will yield aarch64 on Linux, so you'll want to use aarch64 instead everywhere you've written arm64. I'll fix this.

No we definitely want to keep `arm64` here. I vetted most of this in an ubuntu VM and `uname -r` is `arm64`. We need to support both.

jensmeindertsma:
> I've been super busy but I'll have some time this week to get this over the finish line.
> 
> > Yeah, I think the main oversight here was that uname -m will yield aarch64 on Linux, so you'll want to use aarch64 instead everywhere you've written arm64. I'll fix this.
> 
> No we definitely want to keep `arm64` here. I vetted most of this in an ubuntu VM and `uname -r` is `arm64`. We likely need to support both architectures.

Did you mean `uname -m`? That gives me `aarch64` in a Docker container with Debian on my M1 machine

jensmeindertsma:
@vanstinator can you share reproduction steps? I've tried everything, and my Ubuntu / Debian / Alpine Linux Docker container always return `aarch64` from `uname -m` which is what Volta uses to determine the architecture.

jensmeindertsma:
I just confirmed that arm64 and aarch64 on Linux are the very same thing, and that uname will *always* return aarch64 on a ARM system. I therefore think the current implementation is already exhaustive and we don't need to check for arm64 Linux in the install script, as we are already checking for aarch64 which will always be the case on ARM Linux systems

vanstinator:
Hmm ok. Thanks for taking a look and moving this forward in your other PR :). I'll close this one for now.

PR #1163, opened 2022-02-21
Support Alpine

jensmeindertsma:
This pull request implements support for Alpine and the generation of pre-built binaries for Alpine.

This work is not yet finished. Currently OpenSSL is compiled successfully under Alpine aarch64 and x86_64, but when Cargo builds the volta crate it doesn't link against musl correctly for some reason ( `/lib/ld-linux-aarch64.so.1` where we want  `/lib/ld-musl-aarch64.so.1` )

The CI work is finished and the `volta-install.sh` file has also been updated.


jensmeindertsma:
@charlespierce do you have any idea how I can get Rust to stop linking to `/lib64/ld-linux-x86-64.so.2`? It seems to be caused by the `native-tls` dependency? Is there any way we can configure `native-tls` to be built for musl?

jensmeindertsma:
I'll try static linking tomorrow. The only downside would be that we bundle a specific version of OpenSSL. I'll have to make a few changes in the install script to make it stop looking for a OpenSSL install when it detects a musl based system

jensmeindertsma:
Closing this until #1164 is merged and also until a ARM64 PR that I'm yet to write based on #1164 is merged.

PR #1164, opened 2022-02-22
Migrate away from `OpenSSL`

jensmeindertsma:
This pull request swaps out `attohttpc` for `reqwest` with `rustls` and removes soooo much stuff regarding OpenSSL as we don't need it anymore. This is also going to be crucial for ARM64 and MUSL support, as we now don't need to build through Docker anymore.

@charlespierce I know you were thinking about this, what are your thoughts on this PR? Can you leave a review? I think this change makes a lot of sense, especially with the enormous struggles OpenSSL + Rust have presented us with. 

#1161 and #1163 need the changes from this PR


jensmeindertsma:
This PR does break the download URL's though, as we no longer parse the OpenSSL version and also no longer build for a specific OpenSSL version, the download url is now simply suffixed with `-linux`. I don't think this is a major issue

charlespierce:
Hi @jensmeindertsma, we actually intentionally migrated _off_ of `reqwest`, because the async executor bloated both the build time and the compiled binary: See https://github.com/volta-cli/volta/pull/733 for a more thorough description.

`attohttpc` has a `rustls` feature as well, so should we make the switch to use Rustls, we would likely go that route rather than switching back to `reqwest` with all of its extra, unused, features. We have an RFC about that switch open right now, so it's by no means a guarantee, but would be happy to have your thoughts: https://github.com/volta-cli/rfcs/pull/47

Your point about the install script is a good one, I'll make sure to add that to the RFC as well, since we'll need a good plan for that going forward.

jensmeindertsma:
@charlespierce  Ah okay, I'd be happy to use `attohttpc` instead! Let's chat in the RFC and then I can put together a new PR once we've worked out some details.

PR #1165, opened 2022-02-23
Use static TLS implementation

jensmeindertsma:
This pull request implements https://github.com/volta-cli/rfcs/pull/47.

### Tasks
- [ ] Enable `rustls` feature of `attohttpc`
- [ ] Support targets that were previously very hard to support due to the dependency on OpenSSL
  - [ ] `x86_64-unknown-linux-musl`
  - [ ] `aarch64-unknown-linux-gnu`
  - [ ] `aarch64-unknown-linux-musl`
    - [ ] Disable the `bzip2` feature of `zip` as it does not support this target
- [ ] Improve CI
  - [ ] Use matrices in `macos` and `release` jobs
  - [ ] Set up caching of Cargo files
- [ ] Update install script
- [ ] Test `x86_64-unknown-linux-gnu` binary on CentOS 6
- [ ] Test MUSL-based binaries


jensmeindertsma:
This pull request is ready for review, just not ready to be merged yet until I completed the last few tasks.
####  Requesting for review
- @charlespierce 
- @chriskrycho 
- @vanstinator

Reviews from anyone else would be welcome as well 😊

Have a good day!

jensmeindertsma:
As requested by @charlespierce it may be better to switch back to Docker-based builds as that way we can support older Linux distro's. We should base the Docker image off of CentOS 6, this will take some adjustment to the `install-build-tools.sh` script.

jensmeindertsma:
@charlespierce what do you think about the change in upload URL's? There's no longer a version of OpenSSL bundled, so we don't need that in the upload URL anymore. I've removed the OpenSSL version from the upload URL, but does this break installing older Volta versions?

ci/install-build-tools.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if [ -n "$use_clang" ]; then
+  # https://github.com/rustls/rustls/pull/1009 upgraded Rust's LLVM version to
+  # 14
+  llvm_version=14
+  sudo apt-key add ./ci/llvm-snapshot.gpg.key
+  sudo add-apt-repository "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-$llvm_version main"
+  sudo apt-get update
+  install_packages clang-$llvm_version llvm-$llvm_version
+fi
```
jensmeindertsma, line 57: @charlespierce my biggest concern with building in a Docker image based on CentOS 6 is whether these lines here will still work. Surely I can update this to use yum, but a major issue is that LLVM-14 and Clang-14 may not be available for CentOS 6. I can try https://www.vultr.com/docs/how-to-install-llvm-and-clang-on-centos-6 but there may be problems with building from source on such an old Linux distro.

jensmeindertsma:
@charlespierce as you can see here: https://github.com/jensmeindertsma/volta/runs/5311658004?check_suite_focus=true, there's quite some trouble with CentOS 6 package registry. It seems that yum is appending a `:` to the URL which makes the page 404. Is there something wrong with our mirror replacement?
```dockerfile
# CentOS 6 packages are no longer hosted on the main repository, instead they are in the CentOS Vault
RUN sed -i 's/^mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Base.repo && \
    sed -i 's/#baseurl=http:\/\/mirror.centos.org\/centos\/$releasever/baseurl=http:\/\/vault.centos.org\/6.10/g' /etc/yum.repos.d/CentOS-Base.repo
```

jensmeindertsma:
I think it would be acceptable to build the Linux target that we used to build (`x86_64-unknown-linux-gnu`) on CentOS 6, but use a newer image for the new ARM and MUSL builds?

jensmeindertsma:
Reduced the scope of this PR to purely the use of the static TLS implementation, we can figure out ARM and MUSL later in a different PR.

charlespierce:
FYI @jensmeindertsma I just updated our existing Dockerfile to use a mirror of the CentOS vault provided by CERN, rather than the default `vault.centos.org` one, which appears to be more stable.

charlespierce:
I wonder if it's possible to use the Vault versions of the "Software Collections" repo to get a more up-to-date version of GCC to enable `ring` to complete? For reference, I found this issue: https://github.com/briansmith/ring/issues/1417

And related, I found some descriptions of how to update GCC from a different repo:
https://stackoverflow.com/questions/55345373/how-to-install-gcc-g-8-on-centos
https://vault.centos.org/6.10/sclo/

That may enable us to avoid having to build from source (which should significantly speed things up). Also, for ARM / MUSL, since it's a new platform that we didn't support before, I think it's reasonable to base it on a more recent Linux distro version, to simplify that process. Then when we get to releasing Volta 2.0, we can include pulling the build toolchain up to parity with that version, to unify the build steps across the board.

charlespierce:
See also this comment for how to get `ring` to build on CentOS 6: https://github.com/briansmith/ring/issues/673#issuecomment-658459596

jensmeindertsma:
@charlespierce that's a start, but since Rust now requires clang14, and devtools7 doesn't include clang14, we still need to build that from source. Which is where I'm getting stuck.

jensmeindertsma:
If you'd like to learn more about what's required, check the setup and build files used in ring's CI

jensmeindertsma:
Which would be https://github.com/briansmith/ring/blob/main/mk/install-build-tools.sh and the cargo.sh file in the same directory

PR #1179, opened 2022-03-08
Add `Args` event to hold `argv`

mikrostew:
This adds the `argv` info to a new `EventKind::Args` event, so that consumers of the events can differentiate between things like `yarn install` and `yarn --version`, for tracking metrics.

If the `VOLTA_WRITE_EVENTS_FILE` environment variable is set, this also serializes the events to a file, and sets the `EVENTS_FILE` environment variable to the path to that file. I found this much easier to work with in Windows, instead of trying to read events from stdin (which I was not able to get working from a script file).

## Other things

Using `writeln` instead of `write` for the data that is sent over stdin to the executable, so that consumers can read lines instead of characters.

Added some checks for events to the acceptance tests, since we didn't really have any testing around events before.

### DIFF
--- a/crates/volta-core/src/event.rs
+++ b/crates/volta-core/src/event.rs
@@ -102,9 +102,14 @@ impl EventLog {
 
     pub fn add_event_start(&mut self, activity_kind: ActivityKind) {
         let argv = env::args_os()
-            .map(|arg| arg.to_string_lossy().to_string())
-            .collect::<Vec<String>>()
-            .join(" ");
+            .enumerate()
+            .fold(String::new(), |mut result, (i, arg)| {
+                if i > 0 {
+                    result.push(' ');
+                }
+                result.push_str(&arg.to_string_lossy());
+                result
+            });
         self.add_event(EventKind::Start { argv }, activity_kind)
     }
     pub fn add_event_end(&mut self, activity_kind: ActivityKind, exit_code: ExitCode) {
--- a/crates/volta-core/src/event.rs
+++ b/crates/volta-core/src/event.rs
@@ -30,9 +30,7 @@ pub struct ErrorEnv {
 #[derive(Deserialize, Serialize, PartialEq, Debug)]
 #[serde(rename_all = "lowercase")]
 pub enum EventKind {
-    Start {
-        argv: String,
-    },
+    Start,
     End {
         exit_code: i32,
     },
@@ -43,6 +41,9 @@ pub enum EventKind {
     },
     ToolEnd {
         exit_code: i32,
+    },
+    Args {
+        argv: String,
     },
 }
 
@@ -101,16 +102,7 @@ impl EventLog {
     }
 
     pub fn add_event_start(&mut self, activity_kind: ActivityKind) {
-        let argv = env::args_os()
-            .enumerate()
-            .fold(String::new(), |mut result, (i, arg)| {
-                if i > 0 {
-                    result.push(' ');
-                }
-                result.push_str(&arg.to_string_lossy());
-                result
-            });
-        self.add_event(EventKind::Start { argv }, activity_kind)
+        self.add_event(EventKind::Start, activity_kind)
     }
     pub fn add_event_end(&mut self, activity_kind: ActivityKind, exit_code: ExitCode) {
         self.add_event(
@@ -132,6 +124,18 @@ impl EventLog {
             },
             activity_kind,
         )
+    }
+    pub fn add_event_args(&mut self) {
+        let argv = env::args_os()
+            .enumerate()
+            .fold(String::new(), |mut result, (i, arg)| {
+                if i > 0 {
+                    result.push(' ');
+                }
+                result.push_str(&arg.to_string_lossy());
+                result
+            });
+        self.add_event(EventKind::Args { argv }, ActivityKind::Args)
     }
 
     fn add_event(&mut self, event_kind: EventKind, activity_kind: ActivityKind) {
@@ -167,18 +171,7 @@ pub mod tests {
         event_log.add_event_start(ActivityKind::Current);
         assert_eq!(event_log.events.len(), 1);
         assert_eq!(event_log.events[0].name, "current");
-        match event_log.events[0].event {
-            EventKind::Start { ref argv } => {
-                let re = Regex::new("volta_core").unwrap();
-                assert!(re.is_match(argv));
-            }
-            _ => {
-                panic!(
-                    "Expected EventKind::Start {{ argv }}, Got: {:?}",
-                    event_log.events[0].event
-                );
-            }
-        }
+        assert_eq!(event_log.events[0].event, EventKind::Start);
 
         event_log.add_event_end(ActivityKind::Pin, ExitCode::NetworkError);
         assert_eq!(event_log.events.len(), 2);
@@ -198,5 +191,21 @@ pub mod tests {
         assert_eq!(event_log.events.len(), 4);
         assert_eq!(event_log.events[3].name, "install");
         // not checking the error because it has too much machine-specific info
+
+        event_log.add_event_args();
+        assert_eq!(event_log.events.len(), 5);
+        assert_eq!(event_log.events[4].name, "args");
+        match event_log.events[4].event {
+            EventKind::Args { ref argv } => {
+                let re = Regex::new("volta_core").unwrap();
+                assert!(re.is_match(argv));
+            }
+            _ => {
+                panic!(
+                    "Expected EventKind::Args {{ argv }}, Got: {:?}",
+                    event_log.events[4].event
+                );
+            }
+        }
     }
 }
--- a/crates/volta-core/src/event.rs
+++ b/crates/volta-core/src/event.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::error::{ExitCode, VoltaError};
 use crate::hook::Publish;
-use crate::monitor::Monitor;
+use crate::monitor::send_events;
 use crate::session::ActivityKind;
 
 // the Event data that is serialized to JSON and sent the plugin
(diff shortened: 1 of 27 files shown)


src/main.rs
```diff
@@ -25,9 +26,11 @@ pub fn main() {
     Logger::init(LogContext::Volta, verbosity).expect("Only a single logger should be initialized");
 
     let mut session = Session::init();
-    session.add_event_start(ActivityKind::Volta);
+    let volta_argv = env::args().collect::<Vec<String>>().join(" ");
```
charlespierce, line 29: A concern I have here is that `env::args` will [panic](https://doc.rust-lang.org/stable/std/env/fn.args.html#panics) if any of the args are not valid UTF-8. Elsewhere in the app I think we've explicitly used `args_os` to avoid that, and dealt with the potential issue of non-UTF-8 text with the `OsString` type.

tests/acceptance/support/sandbox.rs
```diff
@@ -373,6 +385,17 @@ impl SandboxBuilder {
         self
     }
 
+    /// Add an arbitrary file to the test project within the sandbox,
+    /// give it executable permissions,
+    /// and add its directory to the PATH
+    /// (chainable)
+    pub fn executable_file(mut self, path: &str, contents: &str) -> Self {
```
charlespierce, line 392: This is sweet!

charlespierce (changes requested):
I really like adding some more info to the event logging for users who want that behavior. This winds up adding a fair amount of boilerplate passing the `argv` around to make sure it's available when creating an `EventKind::Start` value. I wonder if it would make sense to build up the `argv` string when actually writing the event, rather than ahead of time and having to pass it around.

Looking at how the events are created, that would probably take some custom (de-)serialization logic since we don't manage the events individually, but would simplify the approach a fair amount.

It would also mess with the testing a bit, since we won't be overriding it directly. A related approach might be to use `LazyCell` and have two constructors: One that takes no arguments and relies on the lazy behavior to load the `argv` from `env::args_os` and another that takes a value for `argv` and we can fill the cell directly. That would let us set a value for testing when we want to override, while not having to pass the arguments around everywhere.

src/main.rs
mikrostew: Oh great callout, I didn't see that when I was adding this. I'll change things to use `env::args_os` instead.

mikrostew:
I think this is good to go now.

crates/volta-core/src/event.rs
```diff
@@ -99,7 +101,11 @@ impl EventLog {
     }
 
     pub fn add_event_start(&mut self, activity_kind: ActivityKind) {
-        self.add_event(EventKind::Start, activity_kind)
+        let argv = env::args_os()
+            .map(|arg| arg.to_string_lossy().to_string())
+            .collect::<Vec<String>>()
```
charlespierce, line 106: Minor, but potentially relevant since this is on the hot path: I think we can avoid the intermediate allocation of `String` values by omitting the `to_string` call here: Map the values to `Cow<str>`, then collect those into a vector and join the vector to a string (`join` should work even when the type is `&str` or `Cow<str>`)

crates/volta-core/src/monitor.rs
```diff
@@ -1,58 +1,79 @@
...
-    /// Returns the current monitor.
-    pub fn new(command: &str) -> Monitor {
-        Monitor {
-            monitor_process: spawn_process(command),
+/// Send event to the spawned command process
+// if hook command is not configured, this is not called
+pub fn send_events(command: &str, events: &[Event]) {
+    match serde_json::to_string_pretty(&events) {
+        Ok(events_json) => {
+            if let Some(tempfile_path) = write_events_file(events_json.clone()) {
```
charlespierce, line 16: May not be a major issue, but since File I/O can be slow, is there a way we can gate the `write_events_file` piece behind an environment variable for use in the tests? The public API for the event hook is still that the data will be written to `stdin`, so we can have the extra bit for use in the tests while keeping the performance of the existing solution.

charlespierce (changes requested):
Apologies for the delay! A couple more thoughts on performance (since this will be on the hot path for all invocations)

crates/volta-core/src/event.rs
mikrostew: I've been trying to get that to work, but I'm not sure how

```rust
let argv = env::args_os()
    .map(|arg| arg.to_string_lossy())
    .collect::<Vec<Cow<str>>>()
    .join(" ");
```

That gives me this error, and I haven't been able to work around it:

```
error[E0515]: cannot return reference to function parameter `arg`
   --> crates/volta-core/src/event.rs:108:24
    |
108 |             .map(|arg| arg.to_string_lossy())
    |                        ^^^^^^^^^^^^^^^^^^^^^ returns a reference to data owned by the current function
```

crates/volta-core/src/monitor.rs
mikrostew: Good point - updated that in 9c857a289a511a202a5ccfc20f06212d527f5194

crates/volta-core/src/event.rs
charlespierce: Ohh, shoot, I missed that `args_os` iterates over _owned_ `OsString` values, so each iteration consumes the value. We may still be able to avoid the extra allocation by essentially implementing what `join` does but at the iterator level:

```rust
let argv = env::args_os()
    .enumerate()
    .fold(String::new(), |mut result, (i, arg)| {
        if i > 0 {
            result.push(' ');
        }
        result.push_str(&arg.to_string_lossy());
        result
    });
```

mikrostew: I just asked about this in the Rust Discord to see if anyone there can help. Will update this discussion with anything I learn from that...

mikrostew: We could also just make non-unicode arguments pay a little extra cost, where the happy path for 99% of cases should be fast:

```rust
let argv = env::args_os()
    .map(|arg| {
        arg.into_string()
            .unwrap_or_else(|s| s.to_string_lossy().into_owned())
    })
    .collect::<Vec<String>>()
    .join(" ");
```

There's also this, that someone suggested in the Rust Discord: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c6b91f154c5b21f8288e0345dcb9c5d5> (no way do I want to go `unsafe` just to make this a little faster 🤣 )

mikrostew: I like your option better though, it doesn't have the suboptimal non-unicode path that mine does.

I updated that in fa4186cf3eb5e20d92dc1d80f60f19b9c60c8bee

mikrostew:
After discussing offline, I changed this to add the `argv` string to a new `Args` event, instead of `Start`. This is only created once, right before the events are published before exiting, so it not affecting the startup hot path.

crates/volta-core/src/session.rs
```diff
@@ -158,12 +160,14 @@ impl Session {
...
-    pub fn exit(self, code: ExitCode) -> ! {
+    pub fn exit(mut self, code: ExitCode) -> ! {
+        self.event_log.add_event_args();
         self.publish_to_event_log();
         code.exit();
     }
 
-    pub fn exit_tool(self, code: i32) -> ! {
+    pub fn exit_tool(mut self, code: i32) -> ! {
+        self.event_log.add_event_args();
```
charlespierce, line 170: Super minor: We can probably do this as part of `publish_to_event_log` right before calling `self.event_log.publish`. Would need to update that method to take `mut self`, but that would be it, I think.

crates/volta-core/src/monitor.rs
```diff
@@ -1,58 +1,84 @@
...
-    /// Returns the current monitor.
-    pub fn new(command: &str) -> Monitor {
-        Monitor {
-            monitor_process: spawn_process(command),
+/// Send event to the spawned command process
+// if hook command is not configured, this is not called
+pub fn send_events(command: &str, events: &[Event]) {
+    match serde_json::to_string_pretty(&events) {
+        Ok(events_json) => {
+            let tempfile_path = match env::var("VOLTA_WRITE_EVENTS_FILE") {
```
charlespierce, line 17: Uber nit: I think elsewhere we've used `env::var_os` when checking for the existence of an environment variable.

charlespierce (approved):
Awesome! Thanks for all the iteration on this :+1: Left a few super minor suggestions but LGTM

crates/volta-core/src/session.rs
mikrostew: I had tried to add this to `publish_to_event_log` in that way, but `self.hooks()` is immutably borrowing, and the call to `add_event_args` is also trying to mutably borrow, and I don't see a way to reconcile that.

charlespierce: Ahh! Took a quick look and I think we can do it with some destructuring, since `publish_to_event_log` takes ownership of `self`:

```rust
    fn publish_to_event_log(self) {
        let Self {
            project,
            hooks,
            mut event_log,
            ..
        } = self;

        let plugin_res = project
            .get()
            .and_then(|p| hooks.get(p))
            .map(|hooks| hooks.events().and_then(|e| e.publish.as_ref()));

        match plugin_res {
            Ok(plugin) => {
                event_log.add_event_args();
                event_log.publish(plugin);
            }
            Err(e) => {
                debug!("Unable to publish event log.\n{}", e);
            }
        }
    }
```
Basically since we know that the `plugin_res` is only holding references to `self.hooks` and `self.project`, we split the components out into separate variables, then the borrow checker can correctly understand that we aren't crossing any references.

crates/volta-core/src/monitor.rs
mikrostew: I've seen both in the code, but using `var_os` does make it easy to use `and_then` with this, so I updated that in fd3c6de8c6a532eefdcf45d11da727fc90d0cd89

crates/volta-core/src/session.rs
mikrostew: Oh, great idea! Didn't even think about destructuring things. That totally works, and I updated to use that in fd3c6de8c6a532eefdcf45d11da727fc90d0cd89

charlespierce (approved):
LGTM! 🚢

PR #1193, opened 2022-03-31
Fetch and resolve Yarn3 using @yarnpkg/cli-dist

### LINKED ISSUE #1195: Support for fetching and resolving Yarn3

mikrostew:
Implementing fetch and resolve for Yarn3. Yarn3 functionality is gated behind the `VOLTA_FEATURE_YARN_3` environment variable, while Yarn1 continues to work same as before.

```
$ VOLTA_FEATURE_YARN_3=1 ../volta/target/debug/volta pin yarn@3
success: pinned yarn@3.2.0 in package.json

$ yarn --version
3.2.0
```

closes #1195 

## Details

It uses the [`@yarnpkg/cli-dist`](https://www.npmjs.com/package/@yarnpkg/cli-dist) package, which is a distribution of the `yarn.js` code from the `@yarnpkg/berry` repo (using [this script](https://github.com/yarnpkg/berry/blob/master/scripts/release/03-release-npm.sh)). In that package, the `bin/yarn` file is not executable as-is, and needs to be given exec permissions.

One thing that's not great in general with supporting Yarn1 and Yarn3 is that we now have two registries for the same tool. So for this implementation I'm searching the Yarn2+ registry first, and if a matching version is not found there, trying the Yarn1 registry:

```
$ VOLTA_FEATURE_YARN_3=ok ./target/debug/volta --verbose fetch yarn@1
[verbose] Did not find yarn matching requirement '>= 1.0.0 < 2.0.0' from https://registry.npmjs.org/@yarnpkg/cli-dist
[verbose] Found yarn@1.22.18 matching requirement '>= 1.0.0 < 2.0.0' from https://registry.npmjs.org/yarn
[verbose] Acquiring lock on Volta directory: /Users/mistewar/.volta/volta.lock
[verbose] Downloading yarn@1.22.18 from https://registry.npmjs.org/yarn/-/yarn-1.22.18.tgz
[verbose] Unpacking yarn into '/Users/mistewar/.volta/tmp/.tmp1Ntmna'
[verbose] Installing yarn in '/Users/mistewar/.volta/tools/image/yarn/1.22.18'
[verbose] Unlocking Volta Directory
success: fetched yarn@1.22.18
```

**Yarn 2.x is not supported**

Skipping Yarn2.x support is recommended in https://github.com/volta-cli/rfcs/pull/48. There are 2 versions of Yarn2 published in `@yarnpkg/cli-dist`, so this change checks if a user attempts to installing Yarn2, and returns an error.

```
VOLTA_FEATURE_YARN_3=1 ../volta/target/debug/volta pin yarn@2
error: Yarn@2 is not recommended for use, and not supported by Volta.

Please use Yarn@3 instead.
```

I am checking that the user's version request only matches versions of Yarn 2.x, which adds some complication, but I don't think we can just check the input for a pattern like 2.x, because a crafty user could do something like

```
$ VOLTA_FEATURE_YARN_3=1 ../volta/target/debug/volta pin "yarn@>1 <3"
success: pinned yarn@2.4.2 in package.json
```

**Tests**

I added some tests mirroring the yarn 1.x testing that already exists, along with the fixtures and mocking support for those tests.

I also added some smoke tests, and updated the use of Node v12 to Node v14, since v12 is [EOL as of April 30](https://nodejs.org/en/about/releases/) (and some of those smoke tests were getting 500 ISE errors from nodejs.org when running locally).


crates/volta-core/src/tool/registry.rs
```diff
@@ -30,11 +30,13 @@ cfg_if! {
     }
 }
 
-pub fn public_registry_package(package: &str, version: &str) -> String {
+// need package and filename for namespaced tools like @yarnpkg/cli-dist, which is located at
+//   https://registry.npmjs.org/@yarnpkg/cli-dist/-/cli-dist-1.2.3.tgz
+pub fn public_registry_package(package: &str, filename: &str, version: &str) -> String {
```
charlespierce, line 35: Would it possibly make sense to split this into two functions? Something like the existing `public_registry_package` and then `scoped_public_registry_package` That does this:

```rust
pub fn scoped_public_registry_package(scope: &str, package: &str, version: &str) -> String {
    format!(
        "{}/{}/-/{}-{}.tgz",
        public_registry_index(scope),
        package,
        package,
        version
    )
}
```

crates/volta-core/src/tool/yarn/fetch.rs
```diff
@@ -122,7 +126,19 @@ fn determine_remote_url(version: &Version, hooks: Option<&ToolHooks<Yarn>>) -> F
             let distro_file_name = Yarn::archive_filename(&version_str);
             hook.resolve(version, &distro_file_name)
         }
-        _ => Ok(public_registry_package("yarn", &version_str)),
+        _ => {
+            let matches_yarn_berry = VersionReq::parse(">=2").unwrap();
```
charlespierce, line 130: Since we're only looking for a major version difference and we have a `Version` object, I think we can check the [`major`](https://docs.rs/semver/latest/semver/struct.Version.html#structfield.major) field of the `Version` directly, rather than creating a `VersionReq` and verifying if it matches. Something like `version.major >= 2`

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -69,8 +71,8 @@ fn resolve_semver(matching: VersionReq, hooks: Option<&ToolHooks<Yarn>>) -> Fall
     }
 }
 
-fn fetch_yarn_index() -> Fallible<(String, PackageIndex)> {
```
charlespierce, line 72: Super minor, but do we have a similar function in the `npm` or `package` resolve modules? It may make sense to consolidate them into a single function and then add the `registry_fetch_error` context when calling.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -84,8 +86,25 @@ fn fetch_yarn_index() -> Fallible<(String, PackageIndex)> {
...
 fn resolve_custom_tag(tag: String) -> Fallible<Version> {
-    let (url, mut index) = fetch_yarn_index()?;
+    if env::var_os("VOLTA_FEATURE_YARN_3").is_some() {
+        // first try yarn2+, which uses "@yarnpkg/cli-dist" instead of "yarn"
+        let (url, mut index) = fetch_yarn_index("@yarnpkg/cli-dist")?;
+
+        let matches_yarn_2 = VersionReq::parse("2.*").unwrap();
+        if let Some(version) = index.tags.remove(&tag) {
+            debug!("Found yarn@{} matching tag '{}' from {}", version, tag, url);
+            if matches_yarn_2.matches(&version) {
```
charlespierce, line 96: Similarly here, we can probably check `version.major == 2` directly, rather than needing to create a `VersionReq` to compare.

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -109,7 +128,41 @@ fn resolve_latest_legacy(url: String) -> Fallible<Version> {
...
+    if env::var_os("VOLTA_FEATURE_YARN_3").is_some() {
+        // first try yarn2+, which uses "@yarnpkg/cli-dist" instead of "yarn"
+        let (url, index) = fetch_yarn_index("@yarnpkg/cli-dist")?;
+        let matching_entries: Vec<PackageDetails> = index
+            .entries
+            .into_iter()
+            .filter(|PackageDetails { version, .. }| matching.matches(version))
+            .collect();
+
+        if matching_entries.len() > 0 {
```
charlespierce, line 140: As suggested by Clippy:
```suggestion
        if !matching_entries.is_empty() {
```

crates/volta-core/src/tool/yarn/resolve.rs
```diff
@@ -109,7 +128,41 @@ fn resolve_latest_legacy(url: String) -> Fallible<Version> {
...
+            let matches_yarn_3 = VersionReq::parse(">=3").unwrap();
+            let details_opt = matching_entries
+                .iter()
+                .find(|PackageDetails { version, .. }| matches_yarn_3.matches(version));
```
charlespierce, line 144: Another spot where we can likely check against the `major` field directly.

crates/volta-core/src/error/kind.rs
```diff
@@ -1361,6 +1376,12 @@ to {}
                 "Could not write Path environment variable.
 
 Please ensure you have permissions to edit your environment variables."
+            ),
+            ErrorKind::Yarn2NotSupported => write!(
+                f,
+                "Yarn@2 is not recommended for use, and not supported by Volta.
+
+Please use Yarn@3 instead."
```
charlespierce, line 1384: My quick intuition here would be to make this slightly more verbose with writing out `version` as opposed to using the `@` syntax, since we're referring to `Yarn` capitalized, not the lower-case package name.

We also should probably make the CTA a little more future proof with something like `Please use version 3 or greater instead.`

charlespierce (changes requested):
This is awesome, thanks @mikrostew! The solution of searching both package registries, while potentially inefficient in some cases, is really clean and easy to follow.

I left a few small style / approach comments, but nothing major.

crates/volta-core/src/tool/registry.rs
mikrostew: Totally makes sense. I split that up like you suggested in db9e6d1d0e1a2ca910231af66ce65b8f9416f382

crates/volta-core/src/tool/yarn/fetch.rs
mikrostew: 🤦  I didn't realize we could just check the major field directly. That's so much nicer! Fixed in db9e6d1d0e1a2ca910231af66ce65b8f9416f382

crates/volta-core/src/tool/yarn/resolve.rs
mikrostew: There's also a `fetch_npm_index` function that overlaps with a lot of the same fetching code, although it has some hook logic first. So I refactored the duplicated code into a new `fetch_public_index` function, which is called by both of the fetch_*_index` functions, in db9e6d1d0e1a2ca910231af66ce65b8f9416f382

crates/volta-core/src/tool/yarn/resolve.rs
mikrostew: Fixed in db9e6d1d0e1a2ca910231af66ce65b8f9416f382

crates/volta-core/src/error/kind.rs
mikrostew: Good points. I updated this error message to be more verbose and future-proof in db9e6d1d0e1a2ca910231af66ce65b8f9416f382

crates/volta-core/src/tool/npm/resolve.rs
```diff
@@ -41,16 +36,7 @@ fn fetch_npm_index(hooks: Option<&ToolHooks<Npm>>) -> Fallible<(String, PackageI
...
-    let metadata: RawPackageMetadata = attohttpc::get(&url)
-        .header(ACCEPT, NPM_ABBREVIATED_ACCEPT_HEADER)
-        .send()
-        .and_then(Response::error_for_status)
-        .and_then(Response::json)
-        .with_context(registry_fetch_error("npm", &url))?;
-
-    spinner.finish_and_clear();
-    Ok((url, metadata.into()))
+    fetch_public_index(url, "Npm")
```
charlespierce, line 39: Super tiny nit: My understanding is that `npm` should always be lowercase as it's part of their brand.

```suggestion
    fetch_public_index(url, "npm")
```

charlespierce (approved):
LGTM with one tiny nit!

Out of curiosity, is the plan to remove the gating around `VOLTA_FEATURE_YARN_3` once the behavior is stable and covers all of the use-cases?

mikrostew:
> Out of curiosity, is the plan to remove the gating around `VOLTA_FEATURE_YARN_3` once the behavior is stable and covers all of the use-cases?

Yeah that's the plan, to leave that undocumented so that users don't accidentally try to use incomplete Yarn3 features, and file issues or get upset on twitter :D. Then once all of the use-cases are covered, remove that and document that we support Yarn3 on the website and such.

crates/volta-core/src/tool/npm/resolve.rs
mikrostew: Oh oops, didn't realize I changed the casing of that error. Good catch!

PR #1204, opened 2022-04-18
Yarn3 Plug'n'Play

### LINKED ISSUE #1197: Support project local binaries with Yarn3 PnP
From [the RFC](https://github.com/volta-cli/rfcs/pull/48):

> Beyond installation, the other blocker for fully supporting Yarn Berry is that our global package delegation does not work with Yarn Plug'n'Play. The root issue is that we expect any project-local binaries to be inside of `node_modules/.bin`, which isn't the case for Plug'n'Play installations.
>
> To work around that, I believe we can detect the presence of the Plug'n'Play include file (named `.pnp.js` or `.pnp.cjs` depending on the Yarn version) and instead of calling the binary ourselves, delegate to it using `yarn run <binary>`. This will make launching a tool slightly slower in those scenarios, however there is no way (without tightly coupling to internal Yarn implementation details) to resolve the path to a binary in Plug'n'Play without invoking `yarn`.

mikrostew:
Draft of support for Yarn3 Plug'n'Play

There were minimal changes needed to get this working, in `crates/volta-core/src/project/mod.rs` and `crates/volta-core/src/run/binary.rs`.

## Tests

The majority of changes were for testing. Previously we had no testing around binary execution, so I added some, with additional tests to verify PnP.

## TODO

* [ ] Make sure this works and is tested for Windows

Closes #1197

### DIFF
--- a/crates/volta-core/fixtures/plug-n-play/pnp-js/package.json
+++ b/crates/volta-core/fixtures/plug-n-play/pnp-js/package.json
@@ -1,0 +1,11 @@
+{
+  "name": "plug-n-play-js",
+  "version": "2.0.0",
+  "description": "Testing that Plug-n-Play things work",
+  "license": "To Ill",
+  "volta": {
+    "node": "6.11.1",
+    "npm": "3.10.10",
+    "yarn": "2.4.0"
+  }
+}
--- a/crates/volta-core/src/run/binary.rs
+++ b/crates/volta-core/src/run/binary.rs
@@ -20,6 +20,17 @@ pub(super) fn command(exe: &OsStr, args: &[OsString], session: &mut Session) -> 
     if let Some(project) = session.project()? {
         // Check if the executable is a direct dependency
         if project.has_direct_bin(exe)? {
+            if project
+                .platform()
+                .map_or(false, |platform| platform.yarn.is_some())
+                && project.is_plug_n_play()
+            {
+                debug!("Project uses Yarn PnP, calling {} with 'yarn'", bin);
+                let platform = Platform::current(session)?;
+                let mut exe_and_args = vec![exe.to_os_string()];
+                exe_and_args.extend_from_slice(args);
+                return Ok(ToolCommand::new("yarn", exe_and_args, platform, ToolKind::Yarn).into());
+            }
             let path_to_bin =
                 project
                     .find_bin(exe)
--- a/crates/volta-core/fixtures/plug-n-play/pnp-cjs/.pnp.cjs
+++ b/crates/volta-core/fixtures/plug-n-play/pnp-cjs/.pnp.cjs
@@ -1,0 +1,1 @@
+// plug and play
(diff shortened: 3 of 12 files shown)


crates/volta-core/fixtures/plug-n-play/pnp-js/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "plug-n-play-js",
+  "version": "2.0.0",
+  "description": "Testing that Plug-n-Play things work",
+  "license": "To Ill",
```
charlespierce, line 5: 😂

crates/volta-core/src/run/binary.rs
```diff
@@ -20,6 +20,17 @@ pub(super) fn command(exe: &OsStr, args: &[OsString], session: &mut Session) ->
     if let Some(project) = session.project()? {
         // Check if the executable is a direct dependency
         if project.has_direct_bin(exe)? {
+            if project
```
charlespierce, line 23: Since this is on the hot path and will require some extra file stat checks, would it be possible to make this lazy with the below `find_bin` call? That call will return `None` if the expected `node_modules/.bin/my_binary` doesn't exist, so we can check _then_ for the PnP behavior. If it's not around, we can error out (in the same way that we do now). That way we only check for the `.pnp.js` file iff we know the binaries aren't already in the `.bin` folder.

Thinking something like:

```rust
match project.find_bin(exe) {
    Some(path_to_bin) => {
        // Existing code for a project binary
    }
    None => {
        if project.platform().map_or(false, |platform| platform.yarn.is_some()) && project.is_plug_n_play() {
            // This block executing `yarn <command>`
        } else {
            // No binary _and_ no PnP
            return Err(ErrorKind::ProjectLocalBinaryNotFound {
                command: exe.to_string_lossy().to_string(),
            }
            .into());
        }
    }
}
```

crates/volta-core/src/run/binary.rs
```diff
@@ -20,6 +20,17 @@ pub(super) fn command(exe: &OsStr, args: &[OsString], session: &mut Session) ->
...
+                .platform()
+                .map_or(false, |platform| platform.yarn.is_some())
+                && project.is_plug_n_play()
```
charlespierce, line 26: Thoughts on rolling the check for `platform.yarn` into the `is_plug_n_play` method? Perhaps renaming to `is_yarn_pnp` or similar?

charlespierce (changes requested):
This is super awesome, thanks @mikrostew! I left a couple of small suggestions for organization and making sure we maintain the hot-path performance.

crates/volta-core/src/run/binary.rs
mikrostew: Makes sense - I updated that to work like how you suggested above

crates/volta-core/src/run/binary.rs
mikrostew: Good idea! I combined that into `is_yarn_pnp`

charlespierce (approved):
LGTM!

PR #1211, opened 2022-04-22
Build Volta for OpenSSL 3 / Ubuntu 22.04

### LINKED ISSUE #881: Download release fails if OpenSSL 3.0.0 installed
The `curl https://get.volta.sh | bash` command fails with 404 Not Found error message if OpenSSL 3.0.0 is installed on Ubuntu.

I uninstalled OpenSSL 3.0.0 and the Volta installation succeeded.

The error message from the failed install was strange, it should be caught in https://github.com/volta-cli/volta/blob/main/dev/unix/volta-install.sh#L353 when you parse the openssl_version or in the parse_os_info method.

The fetching archive for Ubuntu shows release 0.9.2 the current release of Volta. So it took some time for me to figure out why the install failed.

### LINKED ISSUE #1150: Openssl 3.0 support
Openssl 3.0 is probably coming out on archlinux and fedora in the next few months, so building against openssl 3.0 is a good idea when there is a chance to do so.

Additionally, fedora 36 and archlinux will probably make openssl executable the 3.0 version, meaning that the current detection mechanism will detect openssl 3.0, even if openssl 1.1 library files are also installed. This already happens on Fedora rawhide, and I had to manually change the sh script to force 1.1 to make it work. Volta should actually look for the latest supported openssl version using soname rather than the output of `openssl version`.

### LINKED ISSUE #1209: curl https://get.volta.sh | bash doesn't work on ubuntu 22.04
Hello. 

The installation command gives me an error on ubuntu 22.04
`curl https://get.volta.sh | bash`

>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100 12319  100 12319    0     0  21955      0 --:--:-- --:--:-- --:--:-- 21959
>   Installing latest version of Volta (1.0.6)
>     Checking for existing Volta installation
>     Fetching archive for Linux, version 1.0.6
> curl: (22) The requested URL returned error: 404                              
> 
> Error: Could not download Volta version '1.0.6'. See https://github.com/volta-cli/volta/releases for a list of available releases

charlespierce:
Closes #1209 
Closes #1150 
Closes #881 

Info
-----
* With the release of Ubuntu 22.04, OpenSSL 3.0 is packaged by default and Volta is not currently built to link against that version.
* As a short-term fix while we work out the blockers for statically linking, we should build a version of Volta linked with OpenSSL 3 on Linux
* Since we haven't previously supported OpenSSL 3, this is an _additive_ version and we don't need to maintain the backwards compatibility with older versions of `glibc`

Changes
-----
* Added a new docker file that sets up an Ubuntu 22.04 environment to build Volta (with OpenSSL 3)
* Added a new job to the CI configuration that builds Volta in the above docker environment and uploads the release artifact
* Updated the release CI job to pull in and publish the new OpenSSL binaries as well.
* Bumped the `native-tls` and `openssl` versions in `Cargo.lock` to versions that support OpenSSL 3.
* Also fixed a typo in the `content-type` we were sending for the release assets

Tested
-----
* Tested locally in a docker environment built by the docker image
* Additional testing will depend on the CI runs.


.github/workflows/release.yml
```diff
@@ -229,7 +255,16 @@ jobs:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_path: ./release/volta-centos.tar.gz
           asset_name: volta-${{ steps.release_info.outputs.version }}-linux-openssl-rhel.tar.gz
-          asset_content_type: applictaion/gzip
+          asset_content_type: application/gzip
```
chriskrycho, line 258: Yeeeeeesh.

chriskrycho (approved):
Thank you so much for doing this!

Issue #1049, opened 2021-11-23
Provide prebuilt binaries for `alpine` `arm64` environments


jensmeindertsma:
When working in a Docker context on an Apple Silicon Mac, it is common to use the `alpine` image. When I want to install Node.js on this image, I may want to use `volta` to get my image size as small as possible. Currently this doesn't work because Volta fails to install with the error message `Sorry! Volta currently only provides pre-built binaries for x86_64 architectures.`. Is it possible to provide a prebuilt binary for `alpine` `arm64` environments?

jensmeindertsma:
I would love to help out with implementing this. Seems like I just need to add some stuff to the GitHub release workflow?

charlespierce:
@jensmeindertsma We'd absolutely love a contribution to add more pre-built binaries! There are a few pieces that would be needed:

1. Add compilation to the GitHub actions workflow that creates the production builds. Since GH actions runners only provide a few hosts, this will likely have to happen with `cargo cross`. We already do this for the MacOS ARM builds, so it should be analogous.
2. Update the install script `dev/unix/volta-install.sh` to know how to check for that build (and not hard error on other architectures). Again we already did this for MacOS, so I would expect it to be analogous.

One possible source of complication is that on Linux, we currently compile against different versions of OpenSSL in order to support each. I'm not sure how the dynamic linking works with `cargo cross`, so there will likely need to be some investigation into which version is needed and if there is anything special we have to do with the compilation.

jensmeindertsma:
Okay, let me see how far I get and I'll report back 😊

satackey:
Any update on this?
I'm going to try to do this.

jensmeindertsma:
> Any update on this? I'm going to try to do this.

@satackey there's a work in progress PR at #1165 that removes the need to bundle OpenSSL with Volta. If we can get those binaries to compile, we can easily add support for ARM and MUSL environments if we keep the `./ci/cargo.sh` and `./ci/install-build-tools.sh` files which are needed to build properly for MUSL and ARM. The main blocker on that PR is how do we install LLVM 14 and Clang 14 from source? You can see the origin of `./ci/install-build-tools.sh` here: https://github.com/briansmith/ring/blob/32b2c6c39ef459aad47f476fe9139b8d57532ab1/mk/install-build-tools.sh#L82

[...]

PR #1214, opened 2022-04-22
Statically link to Rustls instead of using OpenSSL

charlespierce:
Info
-----
* Implements the changes described in https://github.com/volta-cli/rfcs/pull/47
* As discussed in the RFC, while using a statically-linked TLS implementation reduces our need to build for the unique CentOS flavor of OpenSSL, we still need to use an older version of Linux to link against an older `glibc` version.
* However, the build tools on CentOS 6 (the baseline for the `glibc` support that we've maintained) are too old to support building `ring` (a dependency of `rustls`).
* To work around that, I discovered a docker image for [Scientific Linux CERN 6](https://linux-archive.web.cern.ch/scientific6/), which is based off of RHEL 6 (the same as CentOS).
    * The CERN distro includes its own mirror of the package repositories (in fact, we are already using the cern mirrors for CentOS 6 in our CI right now)
    * It also includes the ability to install more recent build tools, which allows us to compile `ring`

Changes
-----
* Updated the dependency on `attohttpc` to enable the use of `rustls` instead of `native-tls` (OpenSSL on Linux)
* Created a `Dockerfile` that builds off of Scientific Linux CERN 6, installing updated build tools
* Updated the `release` CI workflow to build a single Linux version using the above Docker container.
* Updated the install script to look for the new single version in the case of Linux

Tested
-----
* Confirmed locally using this Docker image that we are able to compile and run Volta
* Verified that the single Linux build artifact from CI runs correctly works on old CentOS Linux, recent Ubuntu, and latest Ubuntu (3 different OpenSSL versions).

Notes
-----
* This PR changes the file name for the `linux` build artifact to be `volta-{version}-linux.tar.gz`, without the `openssl-X` suffix. This means we probably will need to roll it out in stages:
    * Pin the redirect for `get.volta.sh` to a commit hash, so that it isn't affected when this is merged. (Done in https://github.com/volta-cli/get/pull/7) (_Completed_)
    * Merge this and prepare the release (along with any other features that we might want to include)
    * Test by fetching the new script directly to confirm that it works as expected.
    * Cut the release, then un-pin the `get.volta.sh` redirect so that it points at the updated script.
* We'll also need to update the "Installing Old Versions" section of the [Installers](https://docs.volta.sh/advanced/installers) docs to include another special-cased version of the install script, for people who want to install versions prior to the install script change.


alvarlagerlof:
Would this mean that you wouldn't need openssl to install volta?

charlespierce:
@alvarlagerlof Yes, exactly! That's actually one of the main motivations for this suggested change (see the linked RFC): Dealing with all of the different OpenSSL versions—trying to detect the appropriate one, fetch the correct binary, etc.—is a major pain point in our Linux installer. This would simplify all of that by statically linking to a Rust-based TLS/SSL implementation and avoiding all of the headaches that come with dynamically-linked OpenSSL.

ZauberNerd:
@charlespierce I guess this PR would also allow https://github.com/volta-cli/volta/issues/1049 to move forward, right?
I'm using volta on an EC2 with the Graviton2 ARM CPU and at the moment have to compile it myself, but would much prefer volta to release prebuilt binaries.

charlespierce:
@ZauberNerd Yes, I believe this would allow that change to move forward as well, since it would remove the OpenSSL issues that ran into. The shift to Scientific Linux CERN 6 may also help alleviate the toolchain issues.

ZauberNerd:
@charlespierce that's awesome! Is there something I could do to help get this PR out of the draft state?

charlespierce:
@ZauberNerd I don't think there's anything left on this PR directly (other than resolving the merge conflicts). I'm leaving this in draft for now mostly to make sure that we don't accidentally merge it :smile: There's some infrastructure work we'll need to do before merging to make sure we don't break existing installs while we work towards the actual release of this change.

chriskrycho:
@charlespierce what (if anything) can I do to unblock this over this coming week? Would love to help get it landed!

charlespierce:
@chriskrycho I think the only thing remaining is landing this PR (and then some testing when we cut the release). My initial thoughts were being a little wary of version churn with the in-progress pnpm implementation, but I think there's enough significant changes between this and Yarn 3 support, plus there are some follow-ups needed for pnpm, so I'd lean towards getting this merged (along with any smaller outstanding PRs), and then cutting it all as version 1.1.

I'll update this to resolve the merge conflicts and also put out a PR to clean up some additional warnings we seem to be getting.

chriskrycho:
That seems reasonable to me! Let's do it!

dev/unix/volta-install.sh
```diff
@@ -120,21 +120,19 @@ upgrade_is_ok() {
...
     Linux)
-      parsed_version="$(parse_openssl_version "$openssl_version")"
-      exit_code="$?"
-      if [ "$exit_code" != 0 ]; then
-        return "$exit_code"
+      if [ "$arch" == "x86_64" ]; then
+        echo "linux"
+      else
+        error "Releases for non x64 architectures are not currently supported."
+        return 1
```
chriskrycho, line 134: Just double-checking—this isn't a regression, is it? I know we had had folks *interested* in doing Arm builds on Linux, and AFAIK we had never *shipped* those but also hadn’t done things which prevented people from doing it themselves… but don't remember if that was the extent of it?

chriskrycho (approved):
Other than the one question, this looks good to me.

dev/unix/volta-install.sh
charlespierce: Correct, though double-checking we don't actually need that. There is already a `check_architecture` function that prints an error message if the wrong architecture is used.

There was some work to try to get ARM builds for Linux, but I believe it stalled on CI build issues (that I think the approach in this PR should unblock)

charlespierce:
itshappening.gif

chriskrycho:
That's precisely the plan—this and Yarn 3 support will be the headliner features! (With, hopefully, a 1.2 with pnpm support landing *relatively* soon after that. 🤞🏼)

dev/unix/volta-install.sh
vladfrangu: Hey! Would like to add that I this this *does* cause a regression for arm64 users of Linux trying to build and install volta locally (conversation from https://canary.discord.com/channels/413753499835301908/413753499835301917/1026109337313222736 lower) 😅 
Basically this change prevents `./dev/unix/volta-install.sh --release` from being ran on arm64.

charlespierce: Ah, thanks for the call-out @vladfrangu! Will get that reverted (since as noted above it isn’t actually necessary anyway).

PR #1230, opened 2022-05-10
Add "format" to yarn.index hook

### LINKED ISSUE #1196: Implement Hooks for Yarn3
From [the RFC](https://github.com/volta-cli/rfcs/pull/48):

> One outstanding question is how we want to handle the hooks for installs. Since the Hooks format was created before we switched Yarn to use the npm registry, the logic is rather complicated:
> 
> * If we aren't using hooks, then we fetch from npm and expect the data to match the npm package format
> * If we _are_ using hooks, then we fetch from the location specified by the hook and we expect the data to be in the old GitHub API format
>
> With the addition of Yarn 3 pulling from a different package (though still using the npm registry), this could get more complicated. My current suggestion is that we keep the hooks format the same even for Yarn Berry releases. This will mean that teams using hooks will have to make sure the correct package is available at the appropriate URLs themselves. Then, in a future change where we update the hooks format, we can migrate all of Yarn to use the npm format and normalize the behavior.

I commented that:

> [there] could be an optional `"format"` field in the `"yarn"` section of hooks.json, which would take `"npm" | "github"`, like:

```json
    "yarn": {
        "index": {
            "template": "http://example.com/yarn/{{os}}/{{arch}}/yarn-{{version}}.tgz"
        },
        "latest": {
            "prefix": "http://example.com/yarnpkg/"
        },
        "distro": {
            "bin": "~/yarn-distro"
        },
        "format": "npm"
    }
```

> That would specify the format Volta should expect. Leaving this out would default to the current behavior of using the Github API format.

@charlespierce agreed:

> Great suggestion! That would make the different formats more clear / explicit in a backwards-compatible way. It would also enable teams to use the `npm` format for all versions of Yarn, even the older 1.x ones.

mikrostew:
This adds support for user-specified registry formats.

The `"index"` is the only part of the Yarn hooks that can use either Github (legacy) or Npm registry formats.

In [the RFC](https://github.com/volta-cli/rfcs/pull/48), I left a comment suggesting an optional `"format"` field in the `"yarn"` section of the hooks, but ended up implementing this under '"yarn"."index"`, since that is the only hook that is affected by this configuration.

```json
    "yarn": {
        "index": {
            "template": "http://example.com/yarn/{{os}}/{{arch}}/yarn-{{version}}.tgz",
            "format": "npm"
        },
        "latest": {
            "prefix": "http://example.com/yarnpkg/"
        },
        "distro": {
            "bin": "~/yarn-distro"
        }
    }
```

Closes #1196

### DIFF
--- a/crates/volta-core/src/hook/serial.rs
+++ b/crates/volta-core/src/hook/serial.rs
@@ -3,8 +3,9 @@ use std::marker::PhantomData;
 use std::path::Path;
 
 use super::tool;
+use super::RegistryFormat;
 use crate::error::{ErrorKind, Fallible, VoltaError};
-use crate::tool::{Node, Npm, Tool, Yarn};
+use crate::tool::{Node, Npm, Tool};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize)]
@@ -12,6 +13,14 @@ pub struct RawResolveHook {
     prefix: Option<String>,
     template: Option<String>,
     bin: Option<String>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct RawIndexHook {
+    prefix: Option<String>,
+    template: Option<String>,
+    bin: Option<String>,
+    format: Option<String>,
 }
 
 #[derive(Serialize, Deserialize)]
@@ -75,6 +84,26 @@ impl RawResolveHook {
     }
 }
 
+impl RawIndexHook {
+    pub fn into_index_hook(self, base_dir: &Path) -> Fallible<tool::YarnIndexHook> {
+        // use user-specified format, or default to Github (legacy)
+        let format = if let Some(format_str) = self.format {
+            RegistryFormat::from_str(&format_str)?
+        } else {
+            RegistryFormat::Github
+        };
+        Ok(tool::YarnIndexHook {
+            format,
+            metadata: RawResolveHook {
+                prefix: self.prefix,
+                template: self.template,
+                bin: self.bin,
+            }
+            .into_metadata_hook(base_dir)?,
+        })
+    }
+}
+
 impl TryFrom<RawPublishHook> for super::Publish {
     type Error = VoltaError;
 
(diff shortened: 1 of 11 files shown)


crates/volta-core/src/hook/serial.rs
```diff
@@ -75,6 +84,26 @@ impl RawResolveHook {
...
 }
 
+impl RawIndexHook {
+    pub fn into_index_hook(self, base_dir: &Path) -> Fallible<tool::YarnIndexHook> {
+        // use user-specified format, or default to Github (legacy)
+        let format = if let Some(format_str) = self.format {
+            RegistryFormat::from_str(&format_str)?
+        } else {
+            RegistryFormat::Github
+        };
```
charlespierce, line 94: Tiny nit: I typically find it a little more readable to use `match` instead of `if let { } else { }`, something like:

```rust
let format = match self.format {
    Some(format_str) => RegistryFormat::from_str(&format_str)?,
    None => RegistryFormat::Github,
};
```

Another option would be to use the more "fluent" methods on `Option`, but given the need to early return with an error, that may wind up looking _less_ readable / maintainable:

```rust
let format = self.format
    .map(|format_str| RegistryFormat::from_str(&format_str))
    .transpose()?
    .unwrap_or(RegistryFormat::Github);
```

charlespierce (approved):
This looks great, thanks @mikrostew! Left one small style comment, but nothing that blocks the core logic here.

Out of curiosity, what are your thoughts on having the hooks themselves handle the version resolution into some common internal data structure, regardless of the format? This is well outside the scope of this PR (and something I've thought about for the hooks system in general), but it feels a little strange to have the hooks store the "format" value, but then have the `yarn::resolve` method be the one that actually looks at it.

mikrostew:
> Out of curiosity, what are your thoughts on having the hooks themselves handle the version resolution into some common internal data structure, regardless of the format? This is well outside the scope of this PR (and something I've thought about for the hooks system in general), but it feels a little strange to have the hooks store the "format" value, but then have the `yarn::resolve` method be the one that actually looks at it.

That's something I was thinking about while working on this PR, that there should be a more consistent way that resolution is handled. Maybe a common data structure like what you're suggesting. The hooks do need some changes (possibly enough for an RFC).

PR #1242, opened 2022-06-06
fix: renames `release.sh` to `dev/unix/release.sh`

EverlastingBugstopper:
when you pass `--release` or `--dev` to the installer provided at `get.volta.sh`, you get the following error:

```console
$ curl -sSL https://get.volta.sh | bash -s -- --release
Installing Volta locally after compiling with '--release'
    Checking for existing Volta installation
main: line 346: /home/a/work/volta/release.sh: No such file or directory
```

i _believe_ that this PR will fix this by pointing to the correct `release.sh` path, but i'm not entirely sure!

### DIFF
--- a/dev/unix/volta-install.sh
+++ b/dev/unix/volta-install.sh
@@ -343,7 +343,7 @@ compile_and_package() {
 
   # call the release script to create the packaged archive file
   # '2> >(tee /dev/stderr)' copies stderr to stdout, to collect it and parse the filename
-  release_output="$( "$DIR/release.sh" "--$dev_or_release" 2> >(tee /dev/stderr) )"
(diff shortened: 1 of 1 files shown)


EverlastingBugstopper:
on second thought - this might not actually be desired usage of `get.volta.sh` since it does not clone the repository for you. perhaps you are meant to refer to it directly w/`./dev/unix/volta-install.sh --release` and that just works.

EverlastingBugstopper:
fixed w/this approach instead: https://github.com/apollographql/federation-rs/pull/127/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R291-R297

Issue #737, opened 2020-05-08
Support for pnpm


willsoto:
Seems to be a fairly popular alternative to `npm` and `yarn`. Apologies if this has already been suggested but I couldn't find an existing issue. It would be great to be able to `volta pin pnpm`.

https://pnpm.js.org/en/

charlespierce:
Agreed, this is definitely something we’ve discussed on Discord, I’m surprised there wasn’t an issue already either. Thanks!

arxpoetica:
What would need to take place for this to happen? Point me to the right place in the code, and I'll see if I can get around to a PR.

charlespierce:
Hi @arxpoetica, we would definitely welcome that! Adding support for `pnpm`—or really any unsupported package manager—is a bit involved, though hopefully not too difficult. There are 3 main areas that need to be updated:

* Fetching / installing / pinning the version, which is all contained within the `tool` module in `crates/volta-core` (https://github.com/volta-cli/volta/tree/main/crates/volta-core/src/tool)
    * The changes here can mostly be based on what is done for `npm`, the only caveat I'm aware of is that instead of editing the launcher files, we need to create them (so that the bare command `pnpm` works as expected).
* Reading / writing `pnpm` settings, contained in the `platform`, `project`, and `toolchain` modules.
    * The changes here should mostly be adding the `pnpm` key alongside `npm` and `yarn`
* Running `pnpm` through the shim, contained in the `run` module.
    * Here the logic should be almost identical to the `yarn` logic.

That's a very high-level overview of the needed changes, but I'll take a look at write some more this week about any corner cases / difficulties I can predict. I'm also happy to pair on this for a while if that would help.

arxpoetica:
@charlespierce this ended up being something I just haven't had time to get around to yet. It's definitely still on my radar, but if someone feels like giving it a head start (or even attacking totally)...

...I think you'll find it happening more quickly. Sorry for my lack of involvement (re: time).

charlespierce:
@arxpoetica No worries at all, thanks for updating!

willsoto:
@charlespierce slightly related to this, has any thought been given to pinning arbitrary dependencies that would useful to have available on the command-line? For example, it might be useful to have `mocha` or `tsc` installed and available directly on the command line at a specific version without telling users to `volta install` it themselves. Another example is something like `lerna` or `rush` which are the primary entry-points for repos that are managed by those tools.

[...]

PR #1273, opened 2022-07-30
feat: Initial pnpm support

chawyehsu:
This PR is going to implement the first-class support for the pnpm package manager.

Thanks to @mikrostew for the well-written https://github.com/volta-cli/rfcs/pull/46, I've been able to implement this feature initially. Most of the critical functionalities have been implemented, except for the global package management. Due to a lack of further details on how to intercept pnpm to relocate its global package installs to fit Volta's current layout.

#### Implemented

- pnpm shim
- pnpm fetch / install / pin
- pnpm hooks
- per project pnpm manipulation
- volta run with `--pnpm` and `--no-pnpm`

#### Not Implemented

- global package manipulation/management via pnpm

In fact the argument parser `for_pnpm()` has been introduced to try to intercept pnpm args, but I found that there is no proper way to make pnpm's global installs whose original location is `<path to pnpm home>/global/5/node_modules` fit Volta's current global package image layout `<path to volta home>/tools/image/packages`. Hence I intercepted all `--global` pnpm commands and marked them as unimplemented.

#### Trackers

RFC: https://github.com/volta-cli/rfcs/pull/46
Tracking issue: https://github.com/volta-cli/volta/issues/737

This is currently a draft PR, please leave your review or comment if you have any ideas to improve the PR to push forward.

#### Demo

```
❯ .\target\debug\volta.exe list
⚡️ Currently active tools:

    Node: v16.16.0 (default)
    npm: v8.15.1 (default)
    pnpm: v7.7.1 (default)
    Yarn: v1.22.19 (default)
    Tool binaries available:
        cleancss (default)
        yarn-deduplicate (default)

See options for more detailed reports by running `volta list --help`.
DESKTOP in volta on  feature/pnpm [?] is 📦 v1.0.8 via 🦀 v1.62.0
❯ .\target\debug\pnpm.exe -h
Version 7.7.1
Usage: pnpm [command] [flags]
       pnpm [ -h | --help | -v | --version ]

Manage your dependencies:
      add                  Installs a package and any packages that it depends on. By default, any new package is installed as a prod dependency
      import               Generates a pnpm-lock.yaml from an npm package-lock.json (or npm-shrinkwrap.json) file
   i, install              Install all dependencies for a project
  it, install-test         Runs a pnpm install followed immediately by a pnpm test
  ln, link                 Connect the local project to another one
      prune                Removes extraneous packages
  rb, rebuild              Rebuild a package
  rm, remove               Removes packages from node_modules and from the project's package.json
      unlink               Unlinks a package. Like yarn unlink but pnpm re-installs the dependency after removing the external link
  up, update               Updates packages to their latest version based on the specified range

Review your dependencies:
      audit                Checks for known security issues with the installed packages
  ls, list                 Print all the versions of packages that are installed, as well as their dependencies, in a tree-structure
      outdated             Check for outdated packages

Run your scripts:
      exec                 Executes a shell command in scope of a project
      run                  Runs a defined package script
      start                Runs an arbitrary command specified in the package's "start" property of its "scripts" object
   t, test                 Runs a package's "test" script, if one was provided

Other:
      pack
      publish              Publishes a package to the registry
      root

Manage your store:
      store add            Adds new packages to the pnpm store directly. Does not modify any projects or files outside the store
      store prune          Removes unreferenced (extraneous, orphan) packages from the store
      store status         Checks for modified packages in the store

Options:
  -r, --recursive          Run the command for each project in the workspace.
DESKTOP in volta on  feature/pnpm [!] is 📦 v1.1.0-alpha.1 via 🦀 v1.62.0
❯ .\target\debug\pnpm.exe dlx create-vite vite-project --template vanilla
.../../../.pnpm-store/v3/tmp/dlx-31720   |   +6 +
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: D:\.pnpm-store\v3
  Virtual store is at:             ../../../../../.pnpm-store/v3/tmp/dlx-31720/node_modules/.pnpm
.../../../.pnpm-store/v3/tmp/dlx-31720   | Progress: resolved 6, reused 6, downloaded 0, added 6, done

Scaffolding project in D:\workspace\repos\chawyehsu\volta\vite-project...

Done. Now run:

  cd vite-project
  pnpm install
  pnpm run dev

DESKTOP in volta on  feature/pnpm [!?] is 📦 v1.1.0-alpha.1 via 🦀 v1.62.0 took 5s
❯ cd vite-project
DESKTOP in volta\vite-project on  feature/pnpm [!?] via ⬢ v16.16.0
❯ ..\target\debug\pnpm.exe i
Packages: +14
++++++++++++++
Packages are copied from the content-addressable store to the virtual store.
  Content-addressable store is at: D:\.pnpm-store\v3
  Virtual store is at:             node_modules/.pnpm
node_modules/.pnpm/esbuild@0.14.51/node_modules/esbuild: Running postinstall script, done in 318ms
Progress: resolved 34, reused 0, downloaded 14, added 14, done

devDependencies:
+ vite 3.0.4
DESKTOP in volta\vite-project on  feature/pnpm [!?] via ⬢ v16.16.0
❯ ..\target\debug\volta.exe pin node pnpm
success: pinned node@16.16.0 (with npm@8.11.0) in package.json
success: pinned pnpm@7.7.1 in package.json
DESKTOP in volta\vite-project on  feature/pnpm [!?] via ⬢ v16.16.0 took 3s
❯ cat .\package.json
{
  "name": "vite-project",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "vite build",
    "preview": "vite preview"
  },
  "devDependencies": {
    "vite": "^3.0.0"
  },
  "volta": {
    "node": "16.16.0",
    "pnpm": "7.7.1"
  }
}
DESKTOP in volta\vite-project on  feature/pnpm [!?] via ⬢ v16.16.0
❯ ..\target\debug\pnpm.exe --version
7.7.1
DESKTOP in volta\vite-project on  feature/pnpm [!?] via ⬢ v16.16.0
❯
```


hustcer:
Thanks for the PR

hustcer:
Any update on this?

chriskrycho:
@chawyehsu this is *excellent* to see! 👏🏼  I approved CI to run. @charlespierce or @mikrostew if either of you has time to look and see what needs to be done for the global intercept, that would be fantastic. 🤩

chawyehsu:
@chriskrycho Thanks! I fixed tests on Unix according to the CI result.

chriskrycho:
The GH setting is such that I’ll have to reapprove every time, so I'll try to keep an eye on this to keep rerunning it – sorry about the hassle on that front!

crates/volta-core/src/tool/package/manager.rs
```diff
@@ -71,7 +72,7 @@ impl PackageManager {
             // On Windows, npm leaves the binaries at the root of the `prefix` directory
             PackageManager::Npm => package_root,
             // On Windows, Yarn still includes the `bin` subdirectory
-            PackageManager::Yarn => {
+            PackageManager::Pnpm | PackageManager::Yarn => {
```
charlespierce, line 75: small: Can we also update the comment to include `pnpm`?

crates/volta-core/src/tool/package/metadata.rs
```diff
@@ -13,7 +13,7 @@ use semver::Version;
 
 /// Configuration information about an installed package
 ///
-/// Will be stored in <VOLTA_HOME>/tools/user/packages/<package>.json
+/// Will be stored in <VOLTA_HOME>/tools/user/packages/\<package>.json
```
charlespierce, line 16: Is this added `\` intentional?

chawyehsu:
I didn't notice there is a `--features mock-network` flag to turn on all tests including sandbox smoke tests. I was only using `cargo test --all` until I saw the CI arguments.

I'll try to fix them today when I have a couple of hours to figure it out.

crates/volta-layout/src/v3.rs
```diff
@@ -19,11 +19,13 @@ layout! {
             "inventory": inventory_dir {
                 "node": node_inventory_dir {}
                 "npm": npm_inventory_dir {}
+                "pnpm": pnpm_inventory_dir {}
```
charlespierce, line 22: I'm not 100% sure, it may be okay that this is added to the existing layout, rather than included in a migration. We'll need to be sure that it's backwards-compatible, otherwise we'll have to create a migration (which is one of the more awkward areas of the code base). So I'm hopeful that it works and doesn't have issues 🤞🏻

crates/volta-migrate/src/v3/config.rs
```diff
@@ -18,6 +18,8 @@ pub struct LegacyPackageConfig {
 pub struct LegacyPlatform {
     pub node: NodeVersion,
     #[serde(with = "option_version_serde")]
+    pub pnpm: Option<Version>,
+    #[serde(with = "option_version_serde")]
```
charlespierce, line 22: I think we can actually omit this, since the `LegacyPlatform` necessarily won't have a `pnpm` value, as we didn't support it until this point. So we can default the value to `None` for those packages that are being migrated.

crates/volta-core/src/project/mod.rs
```diff
@@ -211,6 +212,22 @@ impl Project {
...
+    pub fn pin_pnpm(&mut self, version: Option<Version>) -> Fallible<()> {
+        if let Some(platform) = self.platform.as_mut() {
+            update_manifest(&self.manifest_file, ManifestKey::Pnpm, version.as_ref())?;
+
+            platform.pnpm = version;
+
+            Ok(())
+        } else {
+            Err(ErrorKind::NoPinnedNodeVersion {
+                tool: "Pnpm".into(),
```
charlespierce, line 225: minor: We should make sure to be consistent (both internally and with `pnpm` branding) in our capitalization. I believe there are some cases that use `Pnpm` while others use `pnpm`. From their website, it looks like `pnpm` uses the same all-lowercase branding that `npm` does, so we should probably match that everywhere.

src/command/run.rs
```diff
@@ -92,6 +96,11 @@ impl Run {
             },
         };
 
+        let pnpm = match &self.pnpm {
+            None => InheritOption::None,
+            Some(version) => InheritOption::Some(pnpm::resolve(version.parse()?, session)?),
+        };
```
charlespierce, line 102: Since `pnpm` isn't bundled with Node the way `npm` is, I believe this should mirror the `yarn` branch more than the `npm` one. Specifically, we should have an explicit `--no_pnpm` flag in the same way that disables the resolution inheritance, while omitting the value entirely leads to `InheritOption::Inherit`.

charlespierce (commented):
Thanks for this! Apologies for taking a bit longer to get to it than I expected, it's been a very busy couple of weeks at work 😅 

I haven't finished reading through the meatiest parts (resolving, parsing, etc.), but left a few minor style comments on the surrounding infrastructure. Will follow-up later in the week with a review on the core parts, as well as some potential thoughts on the `FIXME` comments.

Super excited for this to land!

crates/volta-core/src/tool/package/metadata.rs
chawyehsu: It needs to be escaped, otherwise:
image

I updated it to use inline code block syntax.

crates/volta-layout/src/v3.rs
chawyehsu: This is implemented via https://github.com/volta-cli/volta/pull/1273/commits/1ef45ec9e09b67354d015a2cc2d92f3143a811cd#diff-19d1763ea1b8954788fa4de2fd0d62975d8334c65ca927ddad712d33ef738831R168 thanks to it we don't need to update to layout.v4 to achieve the goal.

ScreamZ:
Any news on this boyz ? Need it

chawyehsu:
@ScreamZ Sorry no ETA, let's wait for the team back from their main work.

crates/volta-core/src/tool/pnpm/resolve.rs
```diff
@@ -0,0 +1,72 @@
...
+            Ok(details.version)
+        }
+        None => Err(ErrorKind::PnpmVersionNotFound {
+            matching: matching.to_string(),
+        }
+        .into()),
+    }
+}
+
+fn fetch_npm_index(hooks: Option<&ToolHooks<Pnpm>>) -> Fallible<(String, PackageIndex)> {
```
charlespierce, line 59: nit: Should be `fetch_pnpm_index` :smile:

crates/volta-core/src/tool/pnpm/fetch.rs
```diff
@@ -0,0 +1,184 @@
...
+) -> Fallible<Box<dyn Archive>> {
+    debug!("Downloading {} from {}", tool_version("pnpm", version), url);
+    Tarball::fetch(url, staging_path).with_context(download_tool_error(
+        tool::Spec::Pnpm(VersionSpec::Exact(version.clone())),
+        url,
+    ))
+}
+
+/// Create executable launchers for the pnpm and pnpx binaries
+fn overwrite_launcher(base_path: &Path, tool: &str) -> Fallible<()> {
```
charlespierce, line 152: small naming nit: It appears that `pnpm` doesn't include launchers at all by default in the package tarball, so rather than "overwrite_[cmd_]launcher", would it make sense for these to be named "write_[cmd_]launcher"?

crates/volta-layout/src/v3.rs
charlespierce: Makes sense, thanks!

charlespierce (commented):
Thank you again! Took a look through the meat of the code and it's all reasonable. Also points to a lot of things it would be good for us to clean up re: Adding package support (lots of boilerplate).

Left a few (mostly minor) suggestions. For the ones in `parser.rs`, I think it's totally reasonable to resolve them with `FIXME` comments, as I see the overall global interception isn't totally wired up yet anyway.

tests/acceptance/support/sandbox.rs
```diff
@@ -622,6 +674,7 @@ impl SandboxBuilder {
         ok_or_panic! { fs::create_dir_all(node_cache_dir()) };
         ok_or_panic! { fs::create_dir_all(node_inventory_dir()) };
         ok_or_panic! { fs::create_dir_all(package_inventory_dir()) };
+        ok_or_panic! { fs::create_dir_all(pnpm_inventory_dir()) };
```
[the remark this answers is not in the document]
chawyehsu: This is just for the sandbox tests. The inventory directory is ensured to exist by the _fetch_ function in production, https://github.com/volta-cli/volta/pull/1273/files#diff-cecf323599e6ac86e26655f65f1935e8d8435f00e05003ad7e228a9dc29d056cR92-R94

crates/volta-core/src/run/parser.rs
```diff
@@ -126,6 +126,71 @@ impl<'a> CommandArg<'a> {
...
+            .into_iter()
+            .map(AsRef::<OsStr>::as_ref)
+            .collect::<Vec<&OsStr>>();
+
+        let (flags, positionals): (Vec<&OsStr>, Vec<&OsStr>) =
+            args.into_iter().partition(|arg| is_flag(arg));
+
+        if positionals.len() < 2 {
+            return CommandArg::Standard;
+        }
```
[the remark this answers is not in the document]
chawyehsu: 1. Improved.
2. Improved.
3. I've updated this part of the implementation, please take another review when you are available.

Currently, the interception parser is dead code and will not be used, since all global commands are captured by https://github.com/volta-cli/volta/pull/1273/files#diff-b72f8a187559437e8f61843775ed13a80245c48032061c623123992d89770e44R17-R26 with the _unimplemented note_ early return.

crates/volta-core/src/tool/package/manager.rs
```diff
@@ -87,6 +88,10 @@ impl PackageManager {
         if let PackageManager::Yarn = self {
             command.env("npm_config_global_folder", self.source_root(package_root));
         }
+
+        // FIXME: Find out if there is a good way to redirect pnpm global installs
```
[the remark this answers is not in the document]
chawyehsu: From my research, I think there will be much work to implement global commands manipulation for pnpm. pnpm relies on absolute path symlinks heavily (binaries generation, symlinked `node_modules` structure), this makes it very difficult for volta to install and update packages via pnpm without a layout change. Related annotation in https://github.com/volta-cli/volta/pull/1273/commits/e830d3dcbc81d41bb4c3af1f4772891d7e5efd18#diff-80dfff59af9a37069dea094c900c30eafd02dd51c2dea21d150a017d1a86c9d3R90.

mikrostew (changes requested):
This will also need to add the `pnpm` and `pnpx` shims to the `get_shim_list_deduped()` fn in `crates/volta-core/src/shim.rs`, for Unix systems to get the new shims (looks like Windows is fine though).

Otherwise I agree with @charlespierce, it seems reasonable not to block on pnpm globals and leave that unimplemented for now.

Sorry there is so much boilerplate needed to onboard a new package manager.

tests/acceptance/support/sandbox.rs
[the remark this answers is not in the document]
mikrostew: It looks like this is not needed - I removed this line and the tests still pass (for me locally on OSX).

chawyehsu:
`pnpx` has been [deprecated](https://pnpm.io/6.x/pnpx-cli), hence I didn't implement the shimming for it.

> it seems reasonable not to block on pnpm globals and leave that unimplemented for now.

It's easy to unblock global commands by simply commenting out the global commands guard in `crates/volta-core/src/run/pnpm.rs`, this will leave pnpm global packages out of volta's control, however. Blocking global commands may bring inconvenience while allowing the unimplemented to be executed outside volta's layout may get package migration issues in the future.

avaly:
Could we get this PR merged as-is with a disclaimer that `pnpm` support is experimental for now, until the global commands can be properly implemented?

We'd really like to use this in our workflows for local development and Docker builds, so missing global support is not really impacting us.

runspired:
I second that: pnpm ships often enough that running `volta install pnpm@<version-project-uses>` many times a day has become annoying :D

chawyehsu:
Hi folks, thanks for all your interest, response, and emoji reactions to this. Undoubtedly, I hope we could make this happen as possible, Volta and pnpm are both great tools for me to be involved in the ecosystem, and I'm happy with them. 

Therefore I made the PR to try to improve these everyday tooling things a little bit more productive and efficient, with a micro goal of hunting some bounties because of my state of out-of-work. The waiting time was a little longer than I expected, expected two months at most for rolling this out. But it's definitely up to the Volta team, I'm not one of its members hence I ain't able to push it to go further. The team is so busy I guess, let's wait a bit more time and we will see.

If you have a strong demand for it right now, you may build a nightly version based on my branch. It's also the base of the local build that I'm currently using.

crates/volta-core/src/tool/package/manager.rs
```diff
@@ -11,72 +12,71 @@ use crate::fs::read_dir_eager;
...
     /// Given the `package_root`, returns the root of the source directory. This directory will
     /// contain the top-level `node-modules`
     #[cfg(unix)]
-    pub fn source_root(self, package_root: PathBuf) -> PathBuf {
-        // On Unix, the source is always within a `lib` subdirectory, with both npm and Yarn
-        let mut path = package_root;
-        path.push("lib");
-
-        path
+    pub fn source_root<P: AsRef<Path>>(self, package_root: P) -> PathBuf {
```
charlespierce, line 29: Is there a change elsewhere that is requiring the switch to accepting an `AsRef<Path>` here? It's minor, but this approach requires an extra copy of the `Path` value to convert it into a `PathBuf`, which the original didn't.

Could we instead continue to use `PathBuf` even for the pnpm case?

crates/volta-core/src/run/pnpm.rs
```diff
@@ -0,0 +1,73 @@
...
+                }
+                .into());
+            }
+
+            if let CommandArg::Global(cmd) = CommandArg::for_pnpm(args) {
+                // For globals, only intercept if the default platform exists
+                if let Some(default_platform) = session.default_platform()? {
+                    return cmd.executor(default_platform);
+                }
+            }
```
charlespierce, line 33: Since we're guarding against global commands above anyway, could we remove this block for now? If necessary, we can add a `#[allow(dead_code)]` directive to the `CommandArg::for_pnpm` method, to prevent warnings about it being unused.

crates/volta-core/src/tool/package/manager.rs
```diff
@@ -11,72 +12,71 @@ use crate::fs::read_dir_eager;
...
+            // possible to retrieve it from pnpm dynamically.
+            PackageManager::Pnpm => package_root.join("5"),
+        }
     }
 
     /// Given the `package_root`, returns the root of the source directory. This directory will
     /// contain the top-level `node-modules`
     #[cfg(windows)]
-    pub fn source_root(self, package_root: PathBuf) -> PathBuf {
+    pub fn source_root<P: AsRef<Path>>(self, package_root: P) -> PathBuf {
```
charlespierce, line 45: Similar question here (and in the below methods as well) about accepting an `AsRef<Path>` and having to do an extra copy.

charlespierce (changes requested):
Exciting! A couple of small changes to make sure we're not impacting the performance of the existing flows with unused code or extra data copying.

To the release strategy, I agree with the plan of merging this and then releasing experimental support. Then we can take another look at the global command interception and look to get that working to remove the experimental label.

chawyehsu:
Changes are applied.

A side note here: When a user updates to a version of Volta that has first-class pnpm support, if they have installed pnpm (as a package) before they might need to uninstall it manually by `npm uninstall -g pnpm` in order to make the new pnpm shim be available.

charlespierce:
Thank you @chawyehsu!

chawyehsu:
Glad that! CI artifacts are available https://github.com/volta-cli/volta/actions/runs/3383641697

ScreamZ:
Just tried, seems to work, now we need a way to auto-update volta ahah v1.1.0 released

smblee:
How can you use this? Is it only available on nightly?

PindaPixel:
Does this take [pnpm/node version compatibility](https://github.com/pnpm/pnpm.github.io/blob/main/docs/installation.md#compatibility) into account?

chawyehsu:
> Does this take [pnpm/node version compatibility](https://github.com/pnpm/pnpm.github.io/blob/main/docs/installation.md#compatibility) into account?

@PindaPixel  Didn't have any extra implementation for the node version compatibility, the compatibility check will be passed to pnpm.

smblee:
Hey @charlespierce just wanted to check on when this PR will be officially released

Issue #1053, opened 2021-11-26
Recreate symlink on `volta install`


felipecrs:
I was using volta from Homebrew (`brew install volta`), but now I'm just using the volta binary at `~/.volta/bin/volta`. When I made this switch, all my previously installed packages stopped to work.

I found out that their symlinks were still pointing to the old Homebrew's volta which I uninstalled.

So, my first attempt was to `volta install npm` and expected that it would recreate the symlink. But it seems that volta skips the creation of it, if it is detected there already.

It could more smarter, by either recreating the symlink even when it already exists, or by checking if the symlink is pointing to the correct `volta-shim`.

chriskrycho:
Thanks for the report. This is definitely an interesting case. We try really hard to avoid having Volta do unnecessary work (especially unnecessary I/O), so we’d have to think a fair bit about this. That said, it’s in a flow where we’re already doing a bunch of *other* I/O which would very much overwhelm the check for symlink targets.

The other (and indeed the *main*) reason I would be hesitant to implement that kind of checking is: having symlinks pointing to the “wrong” place like this could be a signal that you have multiple conflicting installations or otherwise are in a semi-broken state, and trying to fix it could just make things worse. 🤔

svenjacobs:
It seems that Volta installed via [Homebrew](https://brew.sh/) is generally not well supported, am I right? I just updated Volta installed via `brew` from version `1.0.5` to `1.0.6` and all symlinks in `.volta/bin` were broken, still pointing to `/opt/homebrew/Cellar/volta/1.0.5`. However I just learned about the `volta-migrate` command, which seems to fix broken installations? Anyway, I did not find any reference to the `volta-migrate` command in the documentation. Maybe `volta-migrate` should be called in the [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/volta.rb)?

felipecrs:
A relative symlink instead of an absolute one would prevent such issues. Example:

```
# This is how it's being done
❯ ls -la node
lrwxrwxrwx 1 felipecrs felipecrs 37 Feb 28 13:39 node -> /home/felipecrs/.volta/bin/volta-shim

# This is how it could be done
❯ ln -svf volta-shim node
'node' -> 'volta-shim'

❯ ls -la node
lrwxrwxrwx 1 felipecrs felipecrs 10 Apr  6 12:35 node -> volta-shim
```

In the latter example, `node` would always point to a `volta-shim` in the same directory of itself.

[...]

PR #1292, opened 2022-08-12
Check volta shims for reachability during tool install

### LINKED ISSUE #1268: Suggestion: `volta install node` should error if `node` is being shadowed by another binary in PATH
`volta setup` takes care of volta being at the front of the path. However:

1) I hand-maintain my `.profile`
2) [What if two programs did this?](https://devblogs.microsoft.com/oldnewthing/20050607-00/?p=35413)

Today I found out the answer to question 2. `PATH` contained additional binaries by other "shim" programs, and as a result volta's shim `node` was shadowed by another `node`

I believe `volta`, on *every* run of `volta install`, should check `$PATH` and ensure that the just-installed command actually resolves to the shim binary. The `which` crate can be used for this. If that isn't the case, `volta` should at least emit a warning, or even error. From a user POV, nothing was installed.

untitaker:
Fix #1268


crates/volta-core/src/shim.rs
```diff
@@ -89,6 +89,23 @@ pub fn create(shim_name: &str) -> Fallible<ShimResult> {
     }
 }
 
+pub fn check_reachable(shim_name: &str) -> Fallible<()> {
```
charlespierce, line 92: Rather than make the public method fallible (and need to create a new error type), since this is intended to be purely informational, can we make this infallible and only show a message if we're able to successfully determine both the shim and the resolved and compare them?

charlespierce: Separately, also a slight preference for moving this into the `tool/mod.rs` file, rather than holding it here in `shim.rs`. The main reason for that is we can then use the `tool` helper functions for writing the message out (e.g. `note_prefix()` and similar that are used for other notes).

Alternatively, we could have this method be something like `is_shadowed(shim_name: &str) -> ShadowStatus {}` with an enum:

```rust
pub enum ShadowStatus {
    NotShadowed,
    Shadowed {
        expected: PathBuf,
        found: PathBuf,
    }
}
```

And then only build the message inside of the `tool` module.

crates/volta-core/src/shim.rs
```diff
@@ -89,6 +89,23 @@ pub fn create(shim_name: &str) -> Fallible<ShimResult> {
...
+    let resolved = which::which(shim_name).map_err(|e| {
+        VoltaError::from_source(
+            e,
+            ErrorKind::ShimResolveError {
+                name: shim_name.to_string(),
+            },
+        )
+    })?;
+    if resolved != shim {
+        warn!("{} is shadowed by another binary of the same name at {}. Please ensure that {} is at the front of your PATH.", shim_name, resolved.display(), shim.display());
```
charlespierce, line 103: Given where this is used, rather than `warn!`, I would lean a little towards using `info!` and `note_prefix`, as that's the approach we've taken for a similar message when you install a global version of Node but the local version is different. In both cases, the message exists to let you know that while you just installed something, what you get when you run the command may not be what you expect.

crates/volta-core/src/error/kind.rs
```diff
@@ -429,6 +429,11 @@ pub enum ErrorKind {
         name: String,
     },
 
+    /// Thrown when Volta is unable to find the shimmed command.
+    ShimResolveError {
```
charlespierce, line 433: As mentioned below, if we make the lookup infallible by handling (i.e. ignoring) the error directly, we shouldn't need a new error type.

tests/acceptance/support/sandbox.rs
```diff
@@ -460,6 +460,20 @@ impl SandboxBuilder {
         self.add_exec_dir_to_path()
     }
 
+    /// Add an arbitrary file to the test project within the sandbox,
+    /// give it executable permissions,
+    /// and add its directory to the *front* of PATH, shadowing any volta binaries.
```
charlespierce, line 465: nit: Can we update this doc comment to more accurately reflect what the method does? It appears that it doesn't work with any files, rather it only does the prepending. Can we also document the constraint that `prepend_exec_dir_to_path` is exclusive of `add_exec_dir_to_path`?

charlespierce (changes requested):
Thanks for this @untitaker, and apologies for the delay in reviewing! I've unfortunately been increasingly busy with my day job and haven't had the time to dedicate to keeping up with reviews.

Left a couple of suggestions for code organization, but the overall approach is sound and I think will be a big help :tada:

felipecrs:
I think this also fixes https://github.com/volta-cli/volta/issues/1053.

untitaker:
@charlespierce i think i addressed all comments. initially i thought it would make sense to make volta hard-crash with exitcode 1 if the PATH can't be resolved to the shim (i think that would be useful when volta is used in CI), but when going back to this code i saw that  only some error conditions were treated as fatal.

crates/volta-core/src/tool/mod.rs
```diff
@@ -192,3 +193,27 @@ fn registry_fetch_error(
...
+        Ok(home) => home,
+        Err(_) => return,
+    };
+
+    let shim = home.shim_file(shim_name);
+    let resolved = match which::which(shim_name) {
+        Ok(resolved) => resolved,
+        Err(_) => {
+            info!(
+                "{} cannot find command {}. Please ensure that {} is at the front of your PATH.",
```
charlespierce, line 208: Apologies for multiple rounds of iteration, but for an error / info message, we try to take special care to be as helpful as possible. Three relatively minor things that jump out here:

* Would it make sense to use the `home.shim_dir()` value for the CTA part? Using the shim path itself would mean that we would suggest putting `~/.volta/bin/node` on the `PATH`, when what we're actually suggesting is having `~/.volta/bin` on the `PATH`
* Since Volta works on Windows as well (and we do these same checks), we should match Windows' casing for `Path` on those systems. So we should have `PATH` on Unix and `Path` on Windows.
* Also related to Windows, with how the Environment variables are handled, I'm not sure "at the front of your Path" makes the most sense. For this case (the command couldn't be found at all), we can probably say something like "Please ensure that {} is available on your Path".

crates/volta-core/src/tool/mod.rs
```diff
@@ -192,3 +193,27 @@ fn registry_fetch_error(
...
+                note_prefix(),
+                shim_name,
+                shim.display()
+            );
+            return;
+        }
+    };
+    if resolved != shim {
+        info!("{} {} is shadowed by another binary of the same name at {}. Please ensure that {} is at the front of your PATH.", note_prefix(), shim_name, resolved.display(), shim.display());
```
charlespierce, line 217: Similar comments as above for this message re `shim_dir` and `PATH`/`Path`. I also wonder if it would make sense to expand the CTA somewhat, to make it clear why we're suggesting it? Perhaps something like:

```
To ensure your commands work as expected, please move {} to the start of your PATH.
```

charlespierce (changes requested):
Thanks for continuing to iterate on this! A couple of small suggested tweaks to the language, to make sure that we are providing the best UX that we can in these situations.

There also appears to be a merge conflict with dependency updates.

untitaker:
@charlespierce I think I fixed both comments

untitaker:
@charlespierce could you take a look?

chriskrycho:
Sorry about the delay here, @untitaker – will see if we can get this reviewed and merged here shortly (at the start of the new year, if nothing else)!

charlespierce (approved):
LGTM! 🎉

chriskrycho:
I'll merge in `main` to make sure CI is still ✅ and we'll merge it! 🎉

chriskrycho:
Ah, I messed up the merge commit; I will re-create it and force push the correct version in a bit!

untitaker:
I can try to fix the test

chriskrycho:
@untitaker that would be great! I fixed up the one build issue I saw, so you should be unblocked. Let us know when it's fixed and we'll get this merged. Thanks again, and thanks for your patience!

untitaker:
Running this locally:

```
RUST_BACKTRACE=1 cargo test --all --features mock-network
```

gives me this output for most tests:

```

thread 'volta_run::inherited_pnpm' panicked at /Users/untitaker/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mockito-0.31.1/src/lib.rs:1184:72:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/result.rs:1653:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/result.rs:1077:23
   4: mockito::Mock::with_body_from_file
             at /Users/untitaker/.cargo/registry/src/index.crates.io-6f17d22bba15001f/mockito-0.31.1/src/lib.rs:1184:52
   5: acceptance::support::sandbox::SandboxBuilder::distro_mock
             at ./tests/acceptance/support/sandbox.rs:457:25
   6: acceptance::support::sandbox::SandboxBuilder::distro_mocks
             at ./tests/acceptance/support/sandbox.rs:470:20
   7: acceptance::volta_run::inherited_pnpm
             at ./tests/acceptance/volta_run.rs:475:13
   8: acceptance::volta_run::inherited_pnpm::{{closure}}
             at ./tests/acceptance/volta_run.rs:474:20
   9: core::ops::function::FnOnce::call_once
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/ops/function.rs:250:5
  10: core::ops::function::FnOnce::call_once
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

untitaker:
@chriskrycho can you or somebody else help me figure out what is broken with my devenv? I can't figure out the broken tests before that. I am following the instructions in CONTRIBUTING

chriskrycho:
Huh. That *should* have been fixed by some of the work @rwjblue and I did a little while back, but it's possible that we missed something there *or* that I did the merge above wrong *or* that we caused a regression. The issue is that the test mock is asking for a path named `"tests/fixtures/node-v10.99.1040-darwin-arm64.tar.gz"` (and so on for the other failing tests), and there is no `darwin-arm64` fixture for Node v10.99, because that would be a nonsense thing to ask for. What I am trying to figure out now is whether that error came from, because it should certainly be entirely unrelated to your changes here! My current best bet is that I messed up merging into your branch somehow!

untitaker:
> there is no darwin-arm64 fixture for Node v10.99, because that would be a nonsense thing to ask for. 

I switched to working on this PR on an x86 laptop, and everything worked fine. darwin-arm64 does make sense to me, that's just any new MacBook.

Fixed the remaining test failure, so I think this PR should be ready to merge.

chriskrycho:
Aggghhh. I am merging `main` in and merging this PR! Your note got lost in my notifications somehow, @untitaker – so sorry about the delay.

chriskrycho:
Ugh, and now there’s a failure in the Windows CI. I am going to kick it off again in case it was spurious. I will check back in on it tomorrow.

chriskrycho:
Oh, and to answer this bit:

> darwin-arm64 does make sense to me, that's just any new MacBook. So I suppose the devenv does not work on any M1 or M2 macbook? Not sure

The issue there was that the Node version it’s asking for, Node v10.99, does not have an Apple Silicon build (that shipped *several* versions later than that). The build works locally on my M1 machine as well, so that wasn't the issue.

If you or @charlespierce or someone with a Windows box is able to get to the bottom of the most recent test failure here, I still would love for us to get this merged.

untitaker:
@chriskrycho I want to be able to debug this in CI but I am unable to do so due to approved workflows. I think I have found the issue.

untitaker:
If that doesn't fix it, I can also just disable the test on Windows. It's too much fuzz for such a simple change.

chriskrycho:
Just triggered CI again, so hopefully it’ll be green. Assuming we get to ✅ we’ll merge it!

chriskrycho:
We did it! Huzzah!

untitaker:
thanks!

PR #1305, opened 2022-09-02
Remove the feature flag to fully enable Yarn3

mikrostew:
This removes the `VOLTA_FEATURE_YARN_3` feature flag, so that Yarn3 is fully supported.

https://github.com/orgs/volta-cli/projects/1/views/1

I slightly modified some of the logic, because we are checking the `@yarnpkg/cli-dist` registry first for Yarn3, before the `yarn` registry for Yarn1, so that errors with Yarn3 don't impact Yarn1.


charlespierce (approved):
LGTM!

PR #1325, opened 2022-09-19
wfix: grafana and astro images

emersonlaurentino:
Grafana and Astro images are broken. So, this PR fixes it.

Screen Shot 2022-09-19 at 14 46 50


charlespierce (approved):
Awesome, thank you!

PR #1351, opened 2022-10-17
Cache cargo artifacts to speed up CI runs

mikrostew:
Use https://github.com/Swatinem/rust-cache to cache CI artifacts from cargo and speed things up


charlespierce (approved):
LGTM!

PR #1372, opened 2022-11-03
Add multiplatform linux build to CI

abihf:
Fix https://github.com/volta-cli/volta/issues/1049

* Add linux-aarch64 build and assets
* Update `volta-install.sh` scripts to support aarch64 platform on linux


avaly:
Shouldn't this [install script](https://github.com/volta-cli/volta/blob/main/dev/unix/volta-install.sh#L129-L135) also be updated with the support for this architecture?

charlespierce:
Hi @abihf, thanks for the contribution! I'm a little worried this is conflating two related but not identical issues, making it a little more risky. Creating an arm build in the existing Linux pipeline is something we've been looking at for a while (which should be a lot easier now that we dropped the OpenSSL dependency). On the other hand, switching to a fully statically-linked libc is a bigger change that we haven't fully investigated. For that, we should have an RFC so that we can be reasonably confident that we've thought through the impact of that change.

Given that, would you be open to breaking this into two parts? The first being creating a pre-built binary for ARM Linux (using the existing pipeline) and the second being an RFC for switching to statically-linked libc. Totally reasonable to do one or the other (or neither if you don't have the bandwidth!), but I don't think we're in a state right now where we can take on both of these changes at once.

abihf:
I've disabled static link & updated `volta-install.sh`

avaly:
How can we help moving this PR along?

My team is blocked on using the new `pnpm` support from #1273, because we also need a linux-aarch64 binary.

nevir:
Also interested in this

ditachev:
Our team is also in need of a pre-built ARM64 linux binary. Is there any progress on this or is there a plan of action on this going forward?

avarun42:
> Hi @abihf, thanks for the contribution! I'm a little worried this is conflating two related but not identical issues, making it a little more risky. Creating an arm build in the existing Linux pipeline is something we've been looking at for a while (which should be a lot easier now that we dropped the OpenSSL dependency). On the other hand, switching to a fully statically-linked libc is a bigger change that we haven't fully investigated. For that, we should have an RFC so that we can be reasonably confident that we've thought through the impact of that change.
> 
> Given that, would you be open to breaking this into two parts? The first being creating a pre-built binary for ARM Linux (using the existing pipeline) and the second being an RFC for switching to statically-linked libc. Totally reasonable to do one or the other (or neither if you don't have the bandwidth!), but I don't think we're in a state right now where we can take on both of these changes at once.

Hi @charlespierce, it seems that this PR no longer uses a fully statically-linked libc. Would it be possible to get this merged now?

nevir:
Also very interested in this

this-is-tobi:
Hi, any news about this one ?

.github/workflows/release.yml
```diff
@@ -12,27 +12,35 @@ jobs:
...
+          toolchain: 1.63
+          override: true
+      - name: Cargo Cache
+        uses: Swatinem/rust-cache@v2
+      - name: Install cross
+        run: cargo install cross
+      - name: Build
+        run: cross build --release --target ${{ matrix.target }}
+      - name: Create artifact
+        run: tar -czvf volta-${{ matrix.name }}.tar.gz volta volta-shim volta-migrate
```
charlespierce, line 37: I'm not on a machine right now where I can cleanly check, but does this change the `glibc` dependency for the standard linux build? The Docker system setup right now is built intentionally to support running on an old version of `glibc`.

For the ARM build, since we haven't supported it in the past, it's not a big deal to use a more recent version, but for the existing linux build we've been committed to backwards compatibility.

charlespierce (commented):
Overall looks good! Just one open question about the `glibc` compatibility for the existing Linux build.

.github/workflows/release.yml
this-is-tobi: I have never done any development in Rust so it would take me a long time to do a good analysis but I can try to make a proposal.

Could [QEMU](https://github.com/docker/setup-qemu-action) be used in Github actions with a matrix to build 2 linux builder images (`linux/amd64` and `linux/arm64` platform) with differents tags and then use these images with the `docker run --platform` flag to build volta for both architectures?

abihf:
> Overall looks good! Just one open question about the `glibc` compatibility for the existing Linux build.

the `cross` build uses ubuntu:20.04  as base image [link](https://github.com/cross-rs/cross/blob/main/docker/Dockerfile.x86_64-unknown-linux-gnu)

I have tried running it on ubuntu 18.04 and it still running 
```sh
❯ docker run -i -t --rm -v $(pwd):/volta -w /volta ubuntu:18.04 ./volta -v
Updating your Volta directory. This may take a few moments...
1.1.1
```

But unfortunately it can't run on centos:7
```sh
❯ docker run -i -t --rm -v $(pwd):/volta -w /volta centos:7 ./volta -v
./volta: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by ./volta)
```

will try to investigate it later

dev/unix/volta-install.sh
```diff
@@ -129,6 +129,8 @@ parse_os_info() {
     Linux)
       if [ "$arch" == "x86_64" ]; then
         echo "linux"
+      elif [ "$arch" == "aarch64" ]; then
+        echo "linux-aarch64"
       else
         error "Releases for non x64 architectures are not currently supported."
```
Fydon, line 135: Thank you for this. Might be worth updating the message as well

Fydon:
Will using [cargo clean](https://doc.rust-lang.org/cargo/commands/cargo-clean.html) resolve the GLIBC issue here, as [recommended here](https://github.com/cross-rs/cross/issues/711#issuecomment-1146225548)?

Fydon:
I assume that a similar change to [this](https://github.com/volta-cli/volta/pull/1567) for Linux would be ideal, in another pull request?

dev/unix/volta-install.sh
abihf: Hi @Fydon would you point me which message I need to update?

Fydon: Sure. For macOS there is no error text, so perhaps the same is true for Linux when ARM64 is supported. Otherwise the error message should be something like "Only x64 and ARM64 are currently supported". I believe that x64 is x86_64 and not all architectures ending in 64.

charlespierce:
I believe this is handled by #1696. Thanks again for the inspiration!

PR #1375, opened 2022-11-06
Upgrade attohttpc and use native root certificates

### LINKED ISSUE #1364: `volta` not respecting system SSL/TLS certificate settings on macOS
Hi! I'm a huge fan of Volta and went to use it at work with our corporate proxy only to discover that it seems `volta` doesn't respect system certificate settings on macOS. I have the proper certs in my system keychain as well as referenced with the usual `SSL_CERT_FILE` and `SSL_CERT_PATH` variables as appropriate. Is there something I'm doing wrong?

Some info about my system:

```bash
$ uname -a
Darwin HOSTNAME 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64 x86_64

$ volta --version
1.1.0
```

Is there any other info I can provide or something I can do to help test?

Thanks,
+Jonathan

charlespierce:
Closes #1364 

Info
-----
* Prior to version 0.24.0, when using the `tls-rustls` feature flag, `attohttpc` would use WebPKI root certificates instead of the native certificate store for the appropriate system.
* As a result, when we switched in Volta 1.1.0 to use `rustls` instead of `native-tls`, we regressed in our certificate handling, causing issues with corporate proxies.
* In version 0.24.0, `attohttpc` [added the ability](https://github.com/sbstp/attohttpc/pull/134) to use native root certificates with rustls via the `rustls-native-certs` crate.
* This allows the root certificate finding to match that of the `native-tls` implementation.

Changes
-----
* Updated our imports of `attohttpc` to use version 0.24 and activate the `tls-rustls-native-roots` feature flag, to ensure that we are using the expected certificate stores for our network connections.

Tested
-----
* I don't have an easy way to test the native certificates themselves, however I did run the smoke tests locally to ensure that our connections continue to work in the public case.
* Will likely need to confirm with the original issue reporter to verify that this change resolves the issue (once it's released).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,19 +75,19 @@ dependencies = [
 
 [[package]]
 name = "attohttpc"
-version = "0.23.1"
+version = "0.24.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "12489f6c5525b06c7e5d689f0efeeae5b9efc38824478e854da9f4f33f040cf5"
+checksum = "b85f766c20e6ae766956f7a2fcc4e0931e79a7e1f48b29132b5d647021114914"
 dependencies = [
  "flate2",
  "http",
  "log",
  "rustls",
+ "rustls-native-certs",
  "serde",
  "serde_json",
  "url",
  "webpki",
- "webpki-roots",
 ]
 
 [[package]]
@@ -121,6 +121,12 @@ checksum = "0b25d992356d2eb0ed82172f5248873db5560c4721f564b13cb5193bda5e668e"
 dependencies = [
  "byteorder",
 ]
+
+[[package]]
+name = "base64"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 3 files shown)


chriskrycho (approved):

Purexo:
Hello, I just seen this PR. I will take time tomorrow to test it.

Because I originally done this one https://github.com/volta-cli/volta/pull/1365

I can't wait to have native ssl store support in this tool, with this, I will migrate with my teams from nvm-windows to volta at the next release !

Purexo:
Seems OK, after full uninstall volta (from windows params + remove appdata volta folder and program files volta folder) and

```
git pull upstream main
cargo wix --nocapture --package volta --output target\wix\volta-windows.msi
```

Then run the msi

my node / npm commands is working perfectly from volta proxy, with the installation of rights versions of node and npm from package.json volta key.

Yes i'm behind a MITM Proxy enterprise so the CA certs needs to be loaded from the native ssl store.

@charlespierce, can I ask when is plan the next release ?

dryprogrammer:
This breaks in macOS 12.7.6 with error: "Error cause: Io Error: The Trust Settings Record was corrupted." hooks.json mirror will not help.
Unfortunately, GPT recommends "Reset Default Keychains" catastrophically - to anyone seeing this, don't do that.
Solved by downgrading to volta version 1.1.0.

PR #1388, opened 2022-11-24
fix(yarn) detect yarn berry when running bin

xg-wang:
## Summary

In Yarn 3, binaries should be run via `yarn run <bin-name>` instead of the assumption of `node_modules/.bin/<bin-name>`. This is required not only with pnp mode.

## Note
I chose to detect .yarnrc.yml because this is how yarn uses `yarnPath` setup; this may not be required when using corepack, but volta does not support corepack yet.

Ref yarnpkg.com:

> Starting from the v2, they must be written in valid Yaml and have the right extension (simply calling your file .yarnrc won't do).

I am very new to Rust, please correct me if there are things that are wrong or can be improved.

## Test Plan

Built dev volta and verified it works on one of my yarn berry repo.

Note: `cargo test --all --features mock-network` seems to be failing because I don't have `mockito` locally


crates/volta-core/src/project/mod.rs
```diff
@@ -171,13 +171,15 @@ impl Project {
         })
     }
 
-    /// Does this project use Yarn Plug'n'Play?
-    // (project uses Yarn, and either of the files '.pnp.js' or '.pnp.cjs' exist)
-    pub fn is_yarn_pnp(&self) -> bool {
+    /// Does this project use Yarn Berry?
+    // (project uses Yarn berry as detected by the files 'yarnrc.yml', '.pnp.js' or '.pnp.cjs' exist)
+    pub fn is_yarn_berry(&self) -> bool {
```
charlespierce, line 176: very minor: Since this also applies to using PnP in Yarn 1.0, not just Yarn Berry, would it make sense to make this name more generic? Perhaps something like `needs_yarn_run`?

charlespierce (approved):
A small naming suggestion, but otherwise this look good, thank you!

crates/volta-core/src/project/mod.rs
xg-wang: Updated

charlespierce:
Oh boo, sorry about that! It looks like GitHub requires us to manually approve the CI on follow-up pushes. And there's a small formatting issue failing the format check CI. Can you run `cargo fmt` to resolve the formatting?

xg-wang:
@charlespierce oops sorry I missed the message. I pushed the fmt commit.

PR #1393, opened 2022-12-03
fix(yarn) pnp workspace support

mufasa71:
Yarn berry works great with single mode, but does not works if inside workspace directory. I found that this line does not respect root monorepo:
https://github.com/volta-cli/volta/blob/7827adefcb57ffe3beb7896c7ed483b24e619556/crates/volta-core/src/project/mod.rs#L176

as it expect `.pnp.js` on same level as root dir. But workspace dir does not have `.pnp.js` hence it works fine with yarn berry. like `yarn why packagename`

### DIFF
--- a/crates/volta-core/src/project/mod.rs
+++ b/crates/volta-core/src/project/mod.rs
@@ -177,9 +177,18 @@ impl Project {
         self.platform()
             .map_or(false, |platform| platform.yarn.is_some())
             && self.manifest_file.parent().map_or(false, |base_dir| {
+<<<<<<< HEAD
                 base_dir.join(".yarnrc.yml").exists()
                     || base_dir.join(".pnp.js").exists()
                     || base_dir.join(".pnp.cjs").exists()
+=======
+                base_dir.join(".yarnrc.yml").exists()
+		    && (base_dir.join(".pnp.js").exists()
+                    || base_dir.join(".pnp.cjs").exists()
+                    || self
+                        .workspace_roots()
+                        .any(|x| x.join(".pnp.cjs").exists() || x.join(".pnp.js").exists()))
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/project/mod.rs
```diff
@@ -177,7 +177,11 @@ impl Project {
         self.platform()
             .map_or(false, |platform| platform.yarn.is_some())
             && self.manifest_file.parent().map_or(false, |base_dir| {
-                base_dir.join(".pnp.js").exists() || base_dir.join(".pnp.cjs").exists()
+                base_dir.join(".pnp.js").exists()
+                    || base_dir.join(".pnp.cjs").exists()
+                    || self
+                        .workspace_roots()
```
charlespierce, line 183: I think this winds up duplicating work: `self.workspace_roots()` includes as the first entry the parent of `self.manifest_file`, so we're re-checking that. Instead, I think this check that you have is already exactly what we want (and can replace the whole `self.manifest_file...` check)

```rust
self
    .workspace_roots()
    .any(|dir| dir.join(".pnp.cjs").exists() || dir.join(".pnp.js").exists())
```

charlespierce (changes requested):
Thanks for taking this on! A suggestion to avoid duplicating work. Also, there may be some merge conflicts with #1388 which is adding another file to the checks here, so if that is merged first, it will need to be resolved.

crates/volta-core/src/project/mod.rs
mufasa71: Thanks I applied changes, will wait until #1388 is merged to resolve conflicts.

mufasa71: Looks like #1388 is merged, can we merge this now?

charlespierce (approved):

PR #1394, opened 2022-12-03
Gate pnpm support behind a feature flag to avoid compatibility issues

charlespierce:
Info
-----
* #1273 Added support for the core pieces of `pnpm`: Instaling, pinning, and running.
* Global package installation / interception are still TODO as we need to work out the specifics of how to implement those.
* As noted in that PR, this change will break the behavior for anyone who was using `pnpm` as a global package in previous versions of Volta.
* To maintain backwards compatibility, we want to make sure that those users aren't impacted by default.
* We tackled a related issue previously with Yarn 3 support by adding a feature flag environment variable, which should work for this case as well.

Changes
-----
* Added a feature flag environment variable `VOLTA_FEATURE_PNPM` that will turn on the native pnpm support.
* Updated the install, uninstall, run, and platform checkout flows to gate on the feature flag when handling `pnpm`
    * If the feature flag is off, we default to the package behavior, which is exactly how `pnpm` is handled in previous builds.
    * If the feature flag is on, then we use the new behavior that properly supports `pnpm`

Tested
-----
* Updated the tests to set the environment variable for any pnpm tests
* Manually tested that with the feature flag off, things continue to treat `pnpm` as a global package
* Manually tested that with the feature flag on, we use the pinned version of `pnpm` with the new behavior.
* Confirmed that previous builds of Volta will still parse `package.json` if there is a `pnpm` value set - though it will be ignored since older versions don't know anything about that key.

Notes
-----
* Once `pnpm` support is completed with the global package controls, we will likely need some sort of migration to handle users that have `pnpm` installed as a global package and update their default platform with the appropriate pnpm version.
* We will also remove the feature flag gating at that time and bump the minor version as it represents a new feature.

cc @chawyehsu Who did the initial pnpm implementation as well.

### DIFF
--- a/crates/volta-core/src/lib.rs
+++ b/crates/volta-core/src/lib.rs
@@ -20,3 +20,5 @@ pub mod sync;
 pub mod tool;
 pub mod toolchain;
 pub mod version;
+
+const VOLTA_FEATURE_PNPM: &str = "VOLTA_FEATURE_PNPM";
--- a/crates/volta-core/src/platform/mod.rs
+++ b/crates/volta-core/src/platform/mod.rs
@@ -1,8 +1,10 @@
+use std::env;
 use std::fmt;
 
 use crate::error::{ErrorKind, Fallible};
 use crate::session::Session;
 use crate::tool::{Node, Npm, Pnpm, Yarn};
+use crate::VOLTA_FEATURE_PNPM;
 use semver::Version;
 
 mod image;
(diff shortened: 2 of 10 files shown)


smblee:
When do we think this will be merged/released?

mikrostew (approved):
LGTM

dherman (approved):
LGTM2!! 🚢

chawyehsu:
Great to see it gets merged. I guess for releasing the next release, there is only #1372 (https://github.com/volta-cli/volta/pull/1372#issuecomment-1310242774) blocker left now?

PR #1417, opened 2023-01-11
fixed #1311

### LINKED ISSUE #1311: Make it clear that `pin` is needed in addition to `npm i npm`
I recently got a bit confused when installing an old version of npm for a project that needs it.

1. I didn't know prior to this that volta had taken ownership of npm installs. As I was using as a drop in replacement for nvm this additional behaviour was surprising. No objection to it, but just highlighting that it'd not obvious that volta will do this
2. When running `npm i -g npm@6` I intended on the current project using npm@g, but was told :

```
   note: using Volta to install npm
success: installed and set npm@6.14.17 as default
   note: you are using npm@8.14.0 in the current project
```
This does not help me achieve my goal of using npm@6.

So I think the output of `npm i -g npm@X` shodul include an additional line

```
  note: to use npm@6.14.17 in this project run `volta pin npm@6`
```

gautamprikshit1:
Fixes #1311

### DIFF
--- a/crates/volta-core/src/tool/mod.rs
+++ b/crates/volta-core/src/tool/mod.rs
@@ -31,7 +31,7 @@ fn debug_already_fetched<T: Display + Sized>(tool: T) {
 
 #[inline]
 fn info_installed<T: Display + Sized>(tool: T) {
-    info!("{} installed and set {} as default", success_prefix(), tool);
+    info!("{} installed and set {} as default\n{} to use {tool} in this project run `volta pin {tool}`", success_prefix(), tool, note_prefix());
 }
 
 #[inline]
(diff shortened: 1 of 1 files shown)


crates/volta-core/src/tool/mod.rs
```diff
@@ -31,7 +31,7 @@ fn debug_already_fetched<T: Display + Sized>(tool: T) {
 
 #[inline]
 fn info_installed<T: Display + Sized>(tool: T) {
-    info!("{} installed and set {} as default", success_prefix(), tool);
+    info!("{} installed and set {} as default\n{} to use {tool} in this project run `volta pin {too;
```
chriskrycho, line 34: Looks like this is missing some of the necessary bits to actually present the right message here!

```suggestion
    info!("{} installed and set {} as default", success_prefix(), tool);
    info!("{} to use {} in this project run `volta pin {}`, success_prefix(), tool);
```

chriskrycho (changes requested):
Thanks! Needs a small tweak to work as suggested in #1311, suggested as a revision.

chriskrycho (approved):
Thanks! I've kicked off CI (which should pass) and we can get this merged.

PR #1496, opened 2023-04-21
Update Cargo.lock after dependency upgrades

charlespierce:


chriskrycho:
Just checked with @charlespierce and this is safe to close now, superseded by other fixes!

PR #1508, opened 2023-05-05
chore: replace atty with is-terminal

tottoto:
Replaces `atty` with `is-terminal` as the former is unmaintained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,7 +50,7 @@ dependencies = [
 name = "archive"
 version = "0.1.0"
 dependencies = [
- "attohttpc 0.24.1",
+ "attohttpc",
  "cfg-if 1.0.0",
  "flate2",
  "fs-utils",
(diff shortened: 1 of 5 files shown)


chriskrycho:
Thanks for opening this! A couple notes:

1. I don't see any actual explicit indication [on the repo](https://github.com/sunfishcode/is-terminal) that `is-terminal` is unmaintained, only a recommendation that folks use the Rust built-in post-1.70. Do you have some further info?
2. We're currently evaluating our MSRV, and if we bump it to a more recent version it's very likely we will just do exactly what `is-terminal` recommends and use [the thing Rust ships](https://doc.rust-lang.org/1.74.1/std/io/trait.IsTerminal.html).

chriskrycho:
Oh! I see, I read this backwards; you're wanting to move this *from* `atty` *to* `is-terminal`. I think the same basic question applies, and in general I would not be worried by something having gone without an update for a while… but seeing as `atty` [has a known soundness issue](https://github.com/softprops/atty/issues/50), I'm strongly in support here. I will rebase this and get it merged. Thanks!

PR #1545, opened 2023-08-15
Detect when the installer is running via Rosetta

nihalgonsalves:
This fixes an edge case where the installer itself runs under Rosetta on ARM64 macOS, for example when running the installer via a subprocess in another install script that uses an x86 binary, or if you run your terminal via Rosetta.

I'm not sure how to add a test given that it would have to run on an macOS ARM64 CI runner. Here's a manual verification:

```bash
#!/usr/bin/env bash
echo "uname: $(uname -m)"
echo "proc_translated: $(sysctl -n sysctl.proc_translated)"
```

Output:

```
$ ./test.sh
uname: arm64
proc_translated: 0

$ arch -x86_64 ./test.sh
uname: x86_64
proc_translated: 1
```

I don't have an x86_64 Mac to test it, but it returns `0` on native x86_64. This is the relevant information from Apple: https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment#3616845

### DIFF
--- a/dev/unix/volta-install.sh
+++ b/dev/unix/volta-install.sh
@@ -135,7 +135,8 @@ parse_os_info() {
       fi
       ;;
     Darwin)
-      if [ "$(uname -m)" == "arm64" ]; then
+      # If running via Rosetta, uname -m returns x86_64, but sysctl reports proc_translated=1
+      if [ "$(uname -m)" == "arm64" ] || [ "$(sysctl -n sysctl.proc_translated)" == "1" ]; then
         echo "macos-aarch64"
       else
         echo "macos"


chriskrycho:
Thanks for the PR, @nihalgonsalves. It is not immediately obvious to me what problem this solves. That does *not* imply there is no problem here, just that I am not following what the edge case you name is?

nihalgonsalves:
@chriskrycho I had a setup script that among other things, ran a Volta command. 

At the bare minimum, this is something like this:

```python3
import subprocess

subprocess.run("curl https://get.volta.sh | bash", shell=True)
```

If you run this from an emulated Rosetta environment, e.g. an x86 Python interpreter on macOS ARM64, `uname -m` in the setup script results in `x86_64`, which then causes the installer to install x86 Volta, and subsequently x86 Node binaries.

A rather rare edge case I'd imagine, but still nice to fix. In a perfect world nobody would still be using x86 Python on ARM64 macOS, but there are often cases where there's some library with no native bindings built for arm64 and you have to do this to run a project.

chriskrycho:
Got it: so the host terminal instance etc. is all running in native Arm mode, but the subprocess which in turn executes Volta is running in Rosetta, leading to an installation where you *could* and *want to* be running natively but end up running in Rosetta needlessly because of how the script is running? 🤔 

If I am understanding correctly, I see why you would want to tweak this, but am not sure this is *quite* the right behavior. In the case where someone *is* running in a Rosetta terminal setup (or similar) on purpose, they would now end up installing native (non-Rosetta) binaries even though they *want* to be installing via Rosetta. Or am I missing something?

nihalgonsalves:
> In the case where someone is running in a Rosetta terminal setup (or similar) on purpose, they would now end up installing native (non-Rosetta) binaries even though they want to be installing via Rosetta. Or am I missing something?

Correct, also a good point. I guess ideally there'd be a way to specify the architecture you want to install.

On a tangent, it would be nice if you could choose the architecture of Node to install independent of the Volta binary. Today you're tied to what Volta build is installed. I'm not sure how configuring that would look but maybe that's the real solution here?

chriskrycho:
Yeah, that makes sense to me for folks who need it. As a stopgap, I think just documenting the behaviors would be good:

- Volta itself will be installed in Rosetta mode if you run the installer in Rosetta mode.
- Volta will install Rosetta binaries when running in Rosetta mode (at least if I recall correctly!).

There might be more nuance there; it has been a few years since we implemented that (and @charlespierce did the lion's share of that work so I might be missing things!).

PR #1552, opened 2023-09-19
No more symlinks on Windows

### LINKED ISSUE #1397: Dismiss the use of symlinks on Windows
I think it would be a huge security enhancement if Volta didn't make use of symlinks on Windows. [Microsoft discourages the use of symlinks](https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links#vulnerability), for security reasons, and Developer Mode (which is recommended by Volta) enables much more than just symlinks, many of them are undesirable.

It looks like the main Volta logic is implemented by code and the symlink could be easily replaced by a call to `volta-shim`. Actually, I made a test replacing `rimraf`'s symlink by a renamed copy of `volta-shim` and it worked.

If this issue makes sense for the Volta team, I can start working on a PR with this implementation.

PS: this idea came from #964

rcsilva83:
Replace file symlink by shims and dir symlinks on Windows by junctions

closes #1397


crates/volta-core/src/fs.rs
```diff
@@ -118,17 +118,17 @@ pub fn create_staging_dir() -> Fallible<TempDir> {
...
-    #[cfg(windows)]
-    return std::os::windows::fs::symlink_file(src, dest);
-
-    #[cfg(unix)]
-    return std::os::unix::fs::symlink(src, dest);
+    let result = std::fs::copy(src, dest);
+    match result {
+        Ok(_v) => Ok(()),
+        Err(e) => Err(e),
+    }
```
charlespierce, line 131: I'm not sure I totally follow the intent behind redoing this helper to always be a `copy`, can you expand on the reasoning a bit?

My sense is that what we want to do instead is to handle this within the `volta_core::shim::create` method, which would then let us maintain the symlinking behavior on macOS / Linux, where it's a normal and supported tool.

crates/volta-core/src/shim.rs
```diff
@@ -73,7 +73,7 @@ pub fn create(shim_name: &str) -> Fallible<ShimResult> {
     #[cfg(windows)]
     windows::create_git_bash_script(shim_name)?;
 
-    match symlink_file(executable, shim) {
+    match copy_shim(executable, shim) {
```
charlespierce, line 76: As mentioned above, I suspect what we want here are two separate implementations: One for `#[cfg(windows)]` and one for `#[cfg(not(windows))]`, rather than changing the helper that's used. We've already got some custom Windows logic (creating the git bash script), so it's probably the most straightforward at this point to completely separate them.

crates/volta-layout/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "volta-layout"
-version = "0.1.1"
+version = "0.1.2"
```
charlespierce, line 3: Very minor nit: Since this is a purely internal crate, there's not much need to bump the version on changes 😄

crates/volta-layout/src/v4.rs
charlespierce, line 1: 😅 The layout / migration stuff is definitely some of the most complicated / annoying to work with in the code base. Is this layout identical to the `v3` one except for the layout file name?

crates/volta-migrate/src/v4.rs
```diff
@@ -0,0 +1,165 @@
...
+use volta_core::tool::{Package, PackageConfig};
+use volta_core::version::VersionSpec;
+use volta_layout::{v3, v4};
+use walkdir::WalkDir;
+
+mod config;
+
+use config::LegacyPackageConfig;
+
+/// Represents a V4 Volta layout (used by Volta v0.9.0 and above)
```
charlespierce, line 21: nit: We should update this comment to reflect the expected version (1.1.2) rather than 0.9.0

crates/volta-migrate/src/v4.rs
charlespierce, line 1: Is this new layout & migration needed to get things to work with the changed junction / file copy behavior? I'm not sure I totally follow as it looks like it's the same as the V2 -> V3 migration, which is complete at this point.

src/volta-shim.rs
```diff
@@ -1,35 +1,9 @@
...
-            session.exit_tool(code);
-        }
-        Err(Error::Volta(err)) => {
-            report_error(env!("CARGO_PKG_VERSION"), &err);
-            session.add_event_error(ActivityKind::Tool, &err);
-            session.add_event_end(ActivityKind::Tool, err.exit_code());
-            session.exit(ExitCode::ExecutionFailure);
-        }
+        Ok(_) => ExitCode::SUCCESS,
+        Err(_) => ExitCode::FAILURE,
```
charlespierce, line 7: I think this logic may be swallowing the actual exit code: If the `status` call returns `Ok`, that doesn't necessarily mean the command returned success, rather that it was successfully executed.

I believe in the `Ok` case, we probably need to look at the [`ExitStatus::code`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code) method and forward along the underlying exit code.

charlespierce (changes requested):
Thanks for tackling this and many apologies on the delayed response! Left some comments / suggestions on the implementation. At a higher level, I'm a little wary of the overhead from this approach in general - it's an extra indirection of calling a child command which isn't free. This is the hottest path for Volta and one of the design goals of the project is to have as low startup as possible, so that users ideally can't tell that Volta is intercepting & redirecting their commands.

For Unix especially, we should likely continue to take advantage of symlinks, as there aren't the same set of security / usability concerns on those platforms.

Finally, for the failing tests, it appears that there are a couple of files in the `tests` directory that import the `symlink_file` method, which no longer exists. This is causing a compile failure when trying to compile the tests.

crates/volta-core/Cargo.toml
```diff
@@ -45,11 +45,16 @@ once_cell = "1.17.0"
 dunce = "1.0.3"
 ci_info = "0.14.9"
 hyperx = "1.4.0"
-attohttpc = { version = "0.25", default-features = false, features = ["json", "compress", "tls-rustls-native-roots"] }
+attohttpc = { version = "0.25", default-features = false, features = [
+    "json",
+    "compress",
+    "tls-rustls-native-roots",
+] }
```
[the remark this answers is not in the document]
rcsilva83: Yes, it was made automatically... Sorry, I didn't notice. I'll revert it

crates/volta-core/src/fs.rs
rcsilva83: I just created this to maintain the previous design but I'll change it.

crates/volta-layout/src/v4.rs
rcsilva83: Yes, the v4 layout is the same as v3. Any suggestions on this?

crates/volta-migrate/src/v4.rs
rcsilva83: It is needed to recreate the packages shims. If Volta don't recreate, it'll continue to work as before as far as I know. Correct me if I'm mistaken.

rcsilva83:
@charlespierce is there anything else I need to do?

Xstoudi:
Any update? Could be awesome to merge it if ready to allow company-constrained environment to run Volta.

crates/volta-core/src/fs.rs
```diff
@@ -120,14 +120,12 @@ pub fn create_staging_dir() -> Fallible<TempDir> {
...
-    S: AsRef<Path>,
-    D: AsRef<Path>,
+    where
+        S: AsRef<Path>,
+        D: AsRef<Path>,
 {
-    #[cfg(windows)]
     return std::os::windows::fs::symlink_file(src, dest);
 
-    #[cfg(unix)]
```
charlespierce, line 130: I'm not totally sure I follow these changes. If I'm reading the diff correctly (definitely not a guarantee), it looks like this function now unconditionally returns the windows symlink call.

If we're now only calling this on Unix, we can probably remove this function entirely and call `std::os::unix::fs::symlink` directly where we would otherwise call this function. It was originally intended as a helper to simplify dispatching to the correct standard library function based on the OS.

crates/volta-migrate/src/v4.rs
charlespierce: I believe `volta setup` (which is called by the installer) should recreate the shims (if I remember correctly). If we remove the new layout version and migration, does it work and set up the shims correctly?

src/volta-shim.rs
```diff
@@ -1,35 +1,9 @@
...
-            session.add_event_tool_end(ActivityKind::Tool, code);
-            session.exit_tool(code);
-        }
-        Err(Error::Volta(err)) => {
-            report_error(env!("CARGO_PKG_VERSION"), &err);
-            session.add_event_error(ActivityKind::Tool, &err);
-            session.add_event_end(ActivityKind::Tool, err.exit_code());
-            session.exit(ExitCode::ExecutionFailure);
-        }
+        Ok(_) => ExitCode::code,
```
charlespierce, line 6: I think we need to look at the [`code`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code) method on `ExitStatus` directly, I don't believe `code` is attached to `ExitCode` at all.

src/volta-shim.rs
```diff
@@ -1,35 +1,9 @@
...
-pub fn main() {
-    Logger::init(LogContext::Shim, LogVerbosity::Default)
-        .expect("Only a single Logger should be initialized");
-    setup_signal_handler();
-
-    let mut session = Session::init();
-    session.add_event_start(ActivityKind::Tool);
-
-    let result = ensure_layout().and_then(|()| execute_shim(&mut session).into_result());
+pub fn main() -> ExitCode {
```
charlespierce, line 3: Could add some info to the description about the flow that execution goes through on both Windows and Unix after this change? I think I understand it, but I want to make sure I'm tracking it correctly.

For example, the status quo before this change is:

### Unix
node -(symlink)-> volta_shim -> (resolve version) -> `node`

### Windows
node -(symlink)-> volta_shim -> (resolve version) -> `cmd.exe /C node`

charlespierce (changes requested):
At a high-level, do you have any details about the runtime cost of this on Windows? Adding an extra indirection should be relatively quick, but it might not be negligible. What I've done in the past to test this is to use a tool like [hyperfine](https://github.com/sharkdp/hyperfine) to do a bunch of runs calling `node -v` (which necessarily goes through the shim) to get a sense for the overhead that Volta imposes. We don't need super significant results, just a sense of how the startup time is impacted.

crates/volta-core/src/fs.rs
Xstoudi: I removed the function as advised.

src/volta-shim.rs
Xstoudi: Done, let me know if it's not the right way to do it, I'm not the best at Rust syntax.

Xstoudi:
Running hyperfine, this is the result on the current Volta official release on my Windows 10 virtual machine:
```
Benchmark 1: node -v
  Time (mean ± σ):      47.7 ms ±  10.7 ms    [User: 13.1 ms, System: 25.8 ms]
  Range (min … max):    40.0 ms …  93.9 ms    57 runs
```

And this is the result on the release exec compiled from this pull request, again on my Windows 10 virtual machine:
```
Benchmark 1: node -v
  Time (mean ± σ):      46.7 ms ±   3.6 ms    [User: 12.8 ms, System: 29.7 ms]
  Range (min … max):    42.8 ms …  65.0 ms    55 runs
```

I'm not really sure how to interpret this and it's the first time I ever use hyperfine.

crates/volta-migrate/src/v4.rs
Xstoudi: I confirm `volta setup` regenerates the shims: https://github.com/volta-cli/volta/blob/main/src/command/setup.rs#L18
v4 migration is not needed, I'll remove it.

src/volta-shim.rs
rcsilva83: On Windows would be:

node -(runner)-> volta_shim -> (resolve version) -> cmd.exe /C node

charlespierce (commented):
One remaining issue with unilaterally changing the behavior for all platforms when the change is intended to impact Windows.

Separately, reviewing this and continuing to think things through, I wonder if we could leverage `volta run` for all of the Windows use-cases? It is already set up to work similarly to what we're building in this change: Reading the command from the later arguments rather than the name of the executable. We're also already building a helper script as part of setting up Windows shims. I suspect we could make the "shims" for Windows be `.cmd` files with a short script that calls `volta run {arguments}` directly, which would then have the same effect as this change while requiring less overall churn.

I'm going to rough up that approach, which combined with the change here to use junctions rather than directory symlinks may achieve the same goals.

src/volta-shim.rs
```diff
@@ -1,35 +1,12 @@
...
+    let result = Command::new("volta-runner").args(std::env::args()).status();
```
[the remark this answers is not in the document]
rcsilva83: Hi Charles!

Thank you for the feedback.

So you prefer this change is changed only for Windows, right? We can make this condition, no problem.

However, I'm not sure if the `volta run`  proposal would solve the issues. The main use case for the file symlink is the global installs. I made a test with `volta run npm install -g ng` and `volta run ng` and it didn't work. I don't know if I misunderstood something.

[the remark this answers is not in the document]
charlespierce: Yeah, if we move forward here, we should limit the scope to only for Windows, so that we have the smallest possible blast radius.

For the `volta run` idea, I want to do a quick prototype to see if it's feasible, then we can look at which approach makes the most sense.

And thank you again for persisting with this change, despite our busy schedules making the lead times really long! It will be a big improvement in Windows behavior when we don't need the "Developer Mode" warning 💥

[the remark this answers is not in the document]
rcsilva83: Great! So I'll wait for you before making both changes.

chriskrycho:
Closing now, as we landed #1755. Thank you again for getting the ball rolling and doing a ton of research and implementation work, @rcsilva83! This will be a big win for our upcoming 2.0 release!

PR #1600, opened 2023-12-20
Update to zip 0.5 to resolve bzip2 issue

### LINKED ISSUE #1599: Upgrade to latest version of zip
Although we should not have any direct vulnerability as a result of it, we use `zip` and have a transitive dependency on a vulnerable version of `bzip2`, and should update accordingly. This is only an issue because we are on an old enough version of `zip` that we *cannot* get a transitive dependency bump on `bzip2` to a safe range.

chriskrycho:
This only gets us part of the way to fully updated, but our MSRV and the transitive MSRV of one of zip's dependencies means we cannot yet upgrade all the way to v0.6 (latest). It is sufficient to get the vulnerability eliminated, though!

Fixes #1599

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,7 +51,7 @@ name = "archive"
 version = "0.1.0"
 dependencies = [
  "attohttpc",
- "cfg-if 1.0.0",
+ "cfg-if",
  "flate2",
  "fs-utils",
  "hyperx",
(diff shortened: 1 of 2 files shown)


chriskrycho:
Note: this is failing two smoke tests locally:

- `volta_fetch::fetch_node`
- `volta_install::install_node`

It's unclear why that would be, and it's possible it's not a real issue. Will dig in further at some point later (or just be happy if CI is ✅!).

chriskrycho:
Hooray, they pass! I do need to dig into why they failed locally, though, so that we don't end up in flakey test hell.

Issue #1618, opened 2023-12-21
Switch to actions-rust-lang/setup-rust-toolchain


chriskrycho:
- [Action](https://github.com/marketplace/actions/setup-rust-toolchain-for-github-ci)
- [Repo](https://github.com/actions-rust-lang/setup-rust-toolchain)

This will let us use the `rust-toolchain.toml` file as a single source of truth for nearly everything.

PR #1635, opened 2024-01-04
Build universal binary for Mac

scottjasso:
This is a proof-of-concept of how volta can publish universal binaries, rather than needing separate arm64 and x86_64 artifacts. This is a bit easier for distribution, and allows easier switching between x86 and arm (`arch -x86_64 volta run blah`), at the cost of the executable size being 2x.

This is probably not safe to push as-is, since this deletes the arm-only artifact.


chriskrycho:
Thanks for this! We all agreed it seems like a good enhancement and just want to do some testing to confirm that it has no perf impact (we certainly expect it does not; universal binaries are really well optimized, but given how perf-sensitive Volta is, we want to double check!). Assuming all is good, we will probably bundle it into a couple related tweaks to our release process for the upcoming 2.0 release (which, for anyone else reading along, is *solely* about updating our supported OSes; see #1611).

chriskrycho (approved):
I did some basic benchmarking to validate the results here. I tried a bunch of different variations on `hyperfine` invocations and they all ended up looking about the same:

```sh
$ hyperfine \
  "./target/aarch64-apple-darwin/release/volta --help" \
  "./target/universal-apple-darwin/release/volta --help"
Benchmark 1: ./target/aarch64-apple-darwin/release/volta --help
  Time (mean ± σ):       1.9 ms ±   0.3 ms    [User: 1.1 ms, System: 0.5 ms]
  Range (min … max):     1.6 ms …   6.0 ms    639 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./target/universal-apple-darwin/release/volta --help
  Time (mean ± σ):       1.9 ms ±   0.1 ms    [User: 1.1 ms, System: 0.5 ms]
  Range (min … max):     1.7 ms …   2.8 ms    638 runs

  Warning: Command took less than 5 ms to complete. Note that the results might be inaccurate because hyperfine can not calibrate the shell startup time much more precise than this limit. You can try to use the `-N`/`--shell=none` option to disable the shell completely.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  ./target/aarch64-apple-darwin/release/volta --help ran
    1.05 ± 0.17 times faster than ./target/universal-apple-darwin/release/volta --help
```

The universal binary is about 5% slower, but in a way where the delta of ~0.1ms is going to be completely dominated by the shell prompt. For reference, on my M1 Max MBP with clean (literally untouched) default profiles for `bash` and `zsh` and my relatively lightweight profile for `fish`:

- `hyperfine "bash -i -c exit"` → 1.9ms
- `hyperfine -i "zsh -i -c exit"` → 8.1ms
- `hyperfine "fish -i -c exit"` → 6.4ms

Net, I think this is probably fine to merge as part of our release of v2.0.0, as long as we document the change to shipping a universal binary clearly in the release notes!

chriskrycho:
The macOS build failure seems unrelated. I have seen it before but not figured out why it sometimes happens. Trying a rerun.

chriskrycho:
Update: I thought it was unrelated, but it is actually somehow related to switching to having a single build for both… but I do not know why it was not failing *before* I merged in `main`!

chriskrycho:
The one notable difference is that we are using `actions-rust-lang/setup-rust-toolchain` rather than `dtolnay/rust-toolchain@master` since #1618. 🤔

chriskrycho:
This worked perfectly fine locally after a `cargo clean`; I wonder if there is a cache busting issue. 🤔

chriskrycho:
In a classic case of “I should have read *all* of the build logs”, the (first) problem is actually fairly straightforward: see d794b7b12511!

scottjasso:
Wow, thanks so much for merging this in so quickly! Thanks for fixing the merge issues too!

chriskrycho:
@scottjasso no problem – glad I had the time in which to do it! Thanks for getting all the actually hard bits done. 😅

PR #1638, opened 2024-01-10
Fix all recent Clippy lints

chriskrycho:
We run `clippy lint` as part of CI, and Rust has introduced a number of lints since we last updated our Rust version. It is not clear to me why these did not fail on #1611, but they immediately started causing issues for all PRs once that was merged. This PR resolves all outstanding Clippy lints, in most cases with very simple mechanical transformations. The vast majority of those are removing references in contexts where they have become redundant courtesy of improvements to the borrow checker (all dating back to Rust 2021, if I recall correctly.) The only “interesting” change is in 12976b23, which reworks one loop in test support from a loop with the actual functionality in the condition for a `while let` binding to using `loop` with an explicit `break` instead.


chriskrycho:
Looks like I need to tweak the semantics of that `loop` slightly. Will do that tomorrow!

PR #1641, opened 2024-01-11
chore: replace is-terminal crate with std::io::IsTerminal

tottoto:
As the Rust version used by `volta` has been updated to 1.75, `volta` can use `std::io::IsTerminal`.


chriskrycho (approved):
🎉 Thank you!

PR #1646, opened 2024-01-12
chore: replace term_size with terminal_size

tottoto:
`term_size` announces that it is not maintained although it works fine, and proposes using `terminal_size` crate instead.

https://github.com/clap-rs/term_size-rs/pull/31

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1352,13 +1352,13 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


chriskrycho (approved):
Thanks for helping us with all of these! I've pushed one change, to put the two `terminal_width` bits in a `use`, but this looks great and I’ll merge it as soon as CI is ✅!

PR #1650, opened 2024-01-16
Remove code signing and update installer icon

charlespierce:
## Info

* It looks like our Installer already works properly with Winget - Several versions are already set up.
* However, there are a few pieces that we need to clean up:
    * The Installed app icon is the old Notion logo, not the Volta logo
    * The CI for releasing the build will not use code signing any more, given the hurdles / expense of code signing in CI.

## Changes

* Updated the `.ico` file that's included with the Windows installer to properly use the Volta logo.
* Removed the now-expired code signing certificate from the CI build.

## Testing

* We will likely need to do some testing and set up automated releasing of new versions once a Release is published. However, since Volta 1.1.1 is already in Winget, we will need to wait until 2.0.0 to test and verify that process.


chriskrycho (approved):
My basic take: Oh, we can just simplify our process a whole bunch, neat!

PR #1649, opened 2024-01-15
chore: update to clap 4

tottoto:
Updates to `clap` 4.

PR #1651, opened 2024-01-17
Add colors back to help output

chriskrycho:
We lost this with the Clap v4 upgrade, and it is a small nicety that users may miss. Happily, the capability exists, and is just opt-in rather than the default now.

<details><summary>Volta v1</summary>

volta-v1

</details>

<details><summary>Volta v2 <em>without</em> this change</summary>

volta-v2-plain

</details>

<details><summary>Volta v2 <em>with</em> this change</summary>

volta-v2

</details>

Things to notice:

- The Clap output made a couple changes going to v4 (all of which already landed as part of #1649):
    - It now puts flags after commands.
    - It renamed "Subcommands" to "Commands".
    - It switched from all caps to title case.
    - It no longer prints `volta-completions` or similar as the name of subcommands (which is a behavior change we have wanted!).
- I made a couple design choices:
    - I restored the color choices we had before for the headers, options, and flags.
    - I added bright blue for the options and flags in the usage message.
    - I chose to italicize as well as bold the headings, but not to underline them.

Input welcome!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -279,6 +279,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+ "terminal_size",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


rwjblue:
👏👏🎉

PR #1652, opened 2024-01-18
Remove Swatinem/rust-cache from CI configs

chriskrycho:
We get this transitively via actions-rust-lang/setup-rust-toolchain@v1 now, and we are not using any of its configuration options, so we can simply drop it for now.

I am opening this on top of #1635 to see if it resolves the build issue there; I do not expect it to, as I think the underlying caching issue there may be a bit more finicky than this, but it is worth checking.


chriskrycho:
This does indeed not fix the issue, because it was actually d794b7b12511 that was the cause. However, this will still be useful, so I will still land it once #1635 lands.

PR #1657, opened 2024-01-20
feat: use zshenv over zshrc in setup

### LINKED ISSUE #1537: Target `.zshenv` over `.zshrc`
Is there any specific reason why Volta adds her variables to `.zshrc` and not `.zshenv`? Seems a bit more apt.

I could PR if there is conseus.

gautamprikshit1:
Closes #1537 

@chriskrycho

### DIFF
--- a/crates/volta-core/src/error/kind.rs
+++ b/crates/volta-core/src/error/kind.rs
@@ -948,7 +948,7 @@ Use `volta pin yarn` to select a version (see `volta help pin` for more info)."
             ErrorKind::NoShellProfile { env_profile, bin_dir } => write!(
                 f,
                 "Could not locate user profile.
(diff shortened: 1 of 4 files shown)


chriskrycho (approved):
🎉 Thank you!

chriskrycho (changes requested):
So what we have here is good, *and* we will also need to update the `install.sh.in` script, which only checks for `.zshrc` to identify the current shell:

https://github.com/volta-cli/volta/blob/aace0e5463239fde35e343fb3d203ccf43814478/dev/unix/install.sh.in#L118-L156

If you can knock that out as well, we can get this merged!

chriskrycho (approved):
Excellent! Much obliged!

Issue #327, opened 2019-03-27
Support `notion uninstall` for node and yarn


mikrostew:
Similar to #319 - since we have `notion install` for Node and Yarn, we should also support `notion uninstall`.

chriskrycho:
Per conversation in Discord today (see esp. [1], [2], [3], [4]), this is probably worth designing and implementing alongside the logic for `volta install npm`. I've marked it for now as [<b>needs RFC</b>][label] accordingly.

[1]: https://discordapp.com/channels/413753499835301908/413753499835301917/576061595814461450
[2]: https://discordapp.com/channels/413753499835301908/413753499835301917/576062745779109908
[3]: https://discordapp.com/channels/413753499835301908/413753499835301917/576086087072612363
[4]: https://discordapp.com/channels/413753499835301908/413753499835301917/576088824132927498
[label]: https://github.com/notion-cli/notion/issues?q=is%3Aissue+is%3Aopen+label%3A%22needs+RFC%22

BProg:
In a little discord [discussion](https://discordapp.com/channels/413753499835301908/413753499835301917/751459790453997570) around the possibility to _uninstall_ yarn and node, a little idea came up: 
> I was thinking that maybe they will get deleted automatically after some time of not using them xD

maybe worth trying 🤷

dwiyatci:
Hi all 👋

is there an alternative way to do this manually while waiting for the RFC to be done? For example, if I'm not using `yarn@1.22.10` anymore and want to "uninstall" it, would it be enough to `rm -rf` the followings?

- `~/.volta/tools/image/yarn/1.22.10`
- `~/.volta/tools/inventory/yarn/yarn-v1.22.10.tar.gz`

Thank you.

charlespierce:
@dwiyatci Yep, that's exactly what you'd need to do to reclaim the space: The first directory is the unpacked install, while the 2nd is the cached tarball.

Possibly important caveat: After deleting, if you come across that version again in a project, it will be automatically re-downloaded (in the same way that they're automatically downloaded the first time they are seen).

migs35323:
Recently node upped from 18.0.0 to 18.1.0 and i just "updated" it so i have no great reason to keep the  8.flat version, so i was looking to unninstall it just to don't have a massive list of leftovers, maybe based on "volta uninstall" we can also get a "volta prune"? so it deletes the minor releases?
should work with the tools and package managers as well
examples:
installed list:
- node 18.1.0
- node 18.0
[...]

PR #1658, opened 2024-01-20
poc: Enable `volta uninstall node@x.x.x`

morishin:
This pull request is for a PoC.

As per https://github.com/volta-cli/volta/issues/327, the `volta uninstall` command does not allow you to uninstall `node` or `yarn`.
However, I have many patch versions of node installed on my machine, and I wanted to delete older versions to free up unnecessary disk space.
Therefore, it is inconvenient that `volta uninstall node@x.x.x` is not possible.

I looked through the issues and Discord but couldn't understand why this has not been implemented for so many years, so I created a simple implementation for discussion.
I would like to know whether this approach is acceptable or if there are any concerns.

(The core of the implementation is here. The rest of the changes are not significant.)

```rust
fn uninstall(self: Box<Self>, _session: &mut Session) -> Fallible<()> {
    let home = volta_home()?;
    // Acquire a lock on the Volta directory, if possible, to prevent concurrent changes
    let _lock: Result<VoltaLock, crate::error::VoltaError> = VoltaLock::acquire();

    let node_dir = home.node_image_root_dir().join(self.version.to_string());
    if node_dir.exists() {
        remove_dir_if_exists(&node_dir)?;
        info!("{} 'node@{}' uninstalled", success_prefix(), self.version);
    } else {
        warn!("No version 'node@{}' found to uninstall", self.version);
    }

    Ok(())
}
```

### DIFF
--- a/crates/volta-core/src/tool/node/mod.rs
+++ b/crates/volta-core/src/tool/node/mod.rs
@@ -5,12 +5,14 @@ use super::{
     info_project_version, FetchStatus, Tool,
 };
 use crate::error::{ErrorKind, Fallible};
+use crate::fs::remove_dir_if_exists;
 use crate::inventory::node_available;
+use crate::layout::volta_home;
 use crate::session::Session;
-use crate::style::{note_prefix, tool_version};
+use crate::style::{note_prefix, success_prefix, tool_version};
 use crate::sync::VoltaLock;
 use cfg_if::cfg_if;
-use log::info;
+use log::{info, warn};
 use node_semver::Version;
 
 mod fetch;
@@ -243,6 +245,21 @@ impl Tool for Node {
         } else {
             Err(ErrorKind::NotInPackage.into())
         }
+    }
+    fn uninstall(self: Box<Self>, _session: &mut Session) -> Fallible<()> {
+        let home = volta_home()?;
+        // Acquire a lock on the Volta directory, if possible, to prevent concurrent changes
+        let _lock: Result<VoltaLock, crate::error::VoltaError> = VoltaLock::acquire();
+
+        let node_dir = home.node_image_root_dir().join(self.version.to_string());
+        if node_dir.exists() {
+            remove_dir_if_exists(&node_dir)?;
+            info!("{} 'node@{}' uninstalled", success_prefix(), self.version);
+        } else {
+            warn!("No version 'node@{}' found to uninstall", self.version);
+        }
+
+        Ok(())
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/command/uninstall.rs
```diff
@@ -15,10 +14,8 @@ impl Command for Uninstall {
     fn run(self, session: &mut Session) -> Fallible<ExitCode> {
         session.add_event_start(ActivityKind::Uninstall);
 
-        let version = VersionSpec::default();
-        let tool = tool::Spec::from_str_and_version(&self.tool, version);
-
-        tool.uninstall()?;
```
morishin, line 21: This `uninstall` is a function of `Spec`. ↓

https://github.com/volta-cli/volta/blob/17763872574f6c67841282551bc38609a11a9f00/crates/volta-core/src/tool/mod.rs#L120-L124

What I want to do now is to perform an uninstall with a specified version like `volta uninstall node@x.x.x`, so `Spec::uninstall` is not appropriate. Therefore, I considered defining `Tool::uninstall`. (Since `Tool` is the return type of `Spec::resolve`, I believe that "Resolved" mentioned in this comment refers to `Tool`).

But I'm not sure that I can replace `Spec::uninstall` with the `Tool::uninstall` completely.

src/command/uninstall.rs
```diff
@@ -15,10 +14,8 @@ impl Command for Uninstall {
...
+        let tool = tool::Spec::try_from_str(&self.tool)?;
```
karthest, line 17: How about 
```rust
        for tool in Spec::from_strings(&self.tools, "uninstall")? {
            tool.resolve(session)?.uninstall(session)?;
        }
```
just the same way as install which enable to install/uninstall multiple node/package once.

crates/volta-core/src/tool/node/mod.rs
```diff
@@ -244,6 +246,21 @@ impl Tool for Node {
...
     }
+    fn uninstall(self: Box<Self>, _session: &mut Session) -> Fallible<()> {
+        let home = volta_home()?;
+        // Acquire a lock on the Volta directory, if possible, to prevent concurrent changes
+        let _lock: Result<VoltaLock, crate::error::VoltaError> = VoltaLock::acquire();
+
+        let node_dir = home.node_image_root_dir().join(self.version.to_string());
+        if node_dir.exists() {
+            remove_dir_if_exists(&node_dir)?;
+            info!("{} 'node@{}' uninstalled", success_prefix(), self.version);
```
karthest, line 257: I think maybe remove node-version.tar.gz in home.tmp_dir is also needed

PR #1659, opened 2024-01-20
chore: replace hyperx with headers

tottoto:
Replaces `hyperx` crate with `headers` crate to modernize dependencies regarding to HTTP headers. `hyperx` was released as an extraction of the `hyper` 0.11 typed header module. Today, `hyper` community decoupled the header module as `headers` crate and maintains it.

### DIFF
--- a/crates/archive/src/tarball.rs
+++ b/crates/archive/src/tarball.rs
@@ -115,9 +115,11 @@ impl Archive for Tarball {
 /// more efficient than simply downloading the entire file up front.
 fn fetch_isize(url: &str, len: u64) -> Result<[u8; 4], ArchiveError> {
     let mut header_values = Vec::with_capacity(1);
-    Range::bytes(len - 4..=len - 1)
+    Range::bytes(len - 4..len)
         .unwrap()
         .encode(&mut header_values);
+    // We just pushed a header in with the `.encode` above, so there will always
+    // be a value at `.first()`.
     let range_header = header_values.first().unwrap();
 
     let (status, headers, mut response) = attohttpc::get(url)
--- a/crates/archive/src/tarball.rs
+++ b/crates/archive/src/tarball.rs
@@ -9,9 +9,7 @@ use super::{Archive, ArchiveError, Origin};
 use attohttpc::header::HeaderMap;
 use flate2::read::GzDecoder;
 use fs_utils::ensure_containing_dir_exists;
-use hyperx::header::{
-    AcceptRanges, ByteRangeSpec, ContentLength, Header, Range, RangeUnit, TypedHeaders,
-};
+use headers::{AcceptRanges, ContentLength, Header, HeaderMapExt, Range};
 use progress_read::ProgressRead;
 use tee::TeeReader;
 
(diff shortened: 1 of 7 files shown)


crates/archive/src/tarball.rs
```diff
@@ -117,9 +114,14 @@ impl Archive for Tarball {
 /// downloading the entire gzip file. For very small files it's unlikely to be
 /// more efficient than simply downloading the entire file up front.
 fn fetch_isize(url: &str, len: u64) -> Result<[u8; 4], ArchiveError> {
-    let range_header = Range::Bytes(vec![ByteRangeSpec::FromTo(len - 4, len - 1)]);
+    let mut header_values = Vec::with_capacity(1);
+    Range::bytes(len - 4..=len - 1)
```
chriskrycho, line 118: I see why you kept this, as it's the same range as it was previously, but given it's doing `..=` now, I think it would be clearer to drop the `-1` and use `..` instead:

```suggestion
    Range::bytes(len - 4..len)
```

crates/archive/src/tarball.rs
```diff
@@ -117,9 +114,14 @@ impl Archive for Tarball {
...
+        .unwrap()
+        .encode(&mut header_values);
+    let range_header = header_values.first().unwrap();
```
chriskrycho, line 121:
```suggestion
    // We just pushed a header in with the `.encode` above, so there will always
    // be a value at `.first()`.
    let range_header = header_values.first().unwrap();
```

tests/acceptance/support/sandbox.rs
```diff
@@ -44,19 +44,24 @@ impl CacheBuilder {
...
-        let expiry_date = HttpDate::from(if self.expired {
+        let expiry_date = Expires::from(if self.expired {
             SystemTime::now() - one_day
         } else {
             SystemTime::now() + one_day
         });
+
+        let mut header_values = Vec::with_capacity(1);
+        expiry_date.encode(&mut header_values);
+        let encoded_expiry_date = header_values.first().unwrap();
```
chriskrycho, line 55: Out of curiosity, since this pattern is… kind of annoying 😂 is there a migration guide or anything so I could get my head around some of the reasoning behind what they chose API-wise? (I poked around but only a tiny bit! Definitely in “lazy maintainer” mode this Saturday morning.)

crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -253,7 +256,11 @@ fn resolve_node_versions(url: &str) -> Fallible<RawNodeIndex> {
             let expiry = create_staging_file()?;
             let mut expiry_file: &File = expiry.as_file();
 
-            write!(expiry_file, "{}", expires).with_context(|| {
+            let mut header_values = Vec::with_capacity(1);
+            expires.encode(&mut header_values);
+            let encoded_expires = header_values.first().unwrap().to_str().unwrap();
```
chriskrycho, line 261:
```suggestion
            // Since we just `.encode()`d into `header_values
            let encoded_expires = header_values.first().unwrap().to_str().unwrap();
```

tests/acceptance/support/sandbox.rs
chriskrycho, line 55:
```suggestion
        // Since we just `.encode()`d into `header_values, it is guaranteed to
        // have a `.first()`.
        let encoded_expiry_date = header_values.first().unwrap();
```

chriskrycho (approved):
Neat! I had a couple small tweaks, which I will apply, and one question, but this looks great. Thank you!

tests/acceptance/support/sandbox.rs
tottoto: I think the design would try to reduce allocations in the library.

PR #1660, opened 2024-01-22
chore: use once_cell crate uniformly

tottoto:
`once_cell`, `lazycell`, and `lazy_static` crates have been used in the `volta` project. The crates provides similar functionality. As parts of `once_cell`'s API is stabilized at the standard library and the other parts are going on, replaces `lazycell` and `lazy_static` with `once_cell` crate.


chriskrycho (approved):
Excellent. And if/when the rest of these are stabilized into `std`, we will have a straightforward migration path! Thank you!

chriskrycho:
I am re-running CI; that image endpoint fails randomly from time to time. :sigh:

chriskrycho:
Since the issue here is totally unrelated (just a problem with the CentOS mirror), we'll tackle that separately. Merging!

Issue #1391, opened 2022-11-30
Volta should create a shim for `yarnpkg`


yume-chan:
Yarn actually has two bin files:

![image](https://user-images.githubusercontent.com/1330321/204769114-e5abf7df-2bdd-4caa-80d5-2a9c38b8f01a.png)

And some tools are invoking `yarnpkg` instead: https://github.com/expo/expo-cli/issues/1455

So volta should also create a shim for `yarnpkg` to maintain compatibility.

charlespierce:
Thanks for reporting @yume-chan! Interestingly, it looks like that's been the case since 2016 🤯 I don't know how we missed that or never got a report about it until now! Will definitely look to include that shim as well.

migs35323:
whats the difference between the 2 binaries? (yarn vs yarnpkg)
i noticed that volta let me install both

migs35323:
whats the difference between the 2 binaries? (yarn vs yarnpkg)
i noticed that volta let me install both

aleqsio:
I've just now bumped into this issue and can confirm this is quite a problem when using volta together with expo – symlinks don't work with `Volta error: Could not find executable "yarnpkg"`, installing yarnpkg is pointing to a super old package, so the only way that I found is to remove volta entirely.

4lgn:
+1, installing packages with `npx expo install <package>` is not possible when using volta. Given the popularity of `expo` along with all their documentation using such installation of packages to ensure compatibility, this is quite a pain.

alitnk:
Expo [uses `yarnpkg`](https://github.com/expo/expo/blob/cf6faf4712520cd5260c563c01b73dec3fd68c1e/packages/%40expo/package-manager/src/node/YarnPackageManager.ts#L8C28-L8C28). I'm getting errors as well. The shim would be great!

elwayman02:
It looks like this is now also impacting the browserslist/update-db package, which switched to use `yarnpkg` a month ago: https://github.com/browserslist/update-db/issues/24

See also: https://github.com/browserslist/update-db/issues/27

elwayman02:
@charlespierce any updates on whether we'll be able to get a fix for this in Volta?

chriskrycho:
Should be reasonably straightforward, I think; we just need to finish clearing the deck for a lot of small infrastructural issues that have piled up and from which we are slowly digging our way out at present. (Also, if you wanted to learn some Rust… 😏)

PR #1670, opened 2024-01-28
feat: support yarnpkg as an alias for yarn

rytmis:
My attempt at implementing #1391. This is my first Rust code aside from a helloworld example, so you know, it mightn't be perfect yet.

I tested (and wrote) this on macOS Sonoma 14.3 by switching between the 1.1.1 release and this dev build, and it seems to work, but I'm not 100% sure on how to reliably verify it. I bumped the version to 1.1.2 so that the dev install script would let me switch, not sure if I was supposed to do that. :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "volta"
-version = "1.1.1"
+version = "1.1.2"
 authors = ["David Herman <david.herman@gmail.com>", "Charles Pierce <cpierce.grad@gmail.com>"]
 license = "BSD-2-Clause"
 repository = "https://github.com/volta-cli/volta"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1584,7 +1584,7 @@ checksum = "078775d0255232fb988e6fccf26ddc9d1ac274299aaedcedce21c6f72cc533ce"
 
 [[package]]
 name = "volta"
-version = "1.1.1"
+version = "1.1.2"
 dependencies = [
  "cfg-if",
  "ci_info",
--- a/crates/volta-core/src/run/mod.rs
+++ b/crates/volta-core/src/run/mod.rs
@@ -96,7 +96,8 @@ fn get_executor(
                     binary::command(exe, args, session)
                 }
             }
-            Some("yarn") => yarn::command(args, session),
+            Some("yarn") => yarn::command(args, session, "yarn"),
+            Some("yarnpkg") => yarn::command(args, session, "yarnpkg"),
             _ => binary::command(exe, args, session),
         }
     }
--- a/crates/volta-core/src/run/yarn.rs
+++ b/crates/volta-core/src/run/yarn.rs
@@ -16,7 +16,7 @@ use crate::session::{ActivityKind, Session};
 ///
 /// If the command is _not_ a global add / remove or we don't have a default platform, then
 /// we will allow Yarn to execute the command as usual.
-pub(super) fn command(args: &[OsString], session: &mut Session) -> Fallible<Executor> {
+pub(super) fn command(args: &[OsString], session: &mut Session, yarn_executable: &str) -> Fallible<Executor> {
     session.add_event_start(ActivityKind::Yarn);
     // Don't re-evaluate the context or global install interception if this is a recursive call
     let platform = match env::var_os(RECURSION_ENV_VAR) {
@@ -32,8 +32,8 @@ pub(super) fn command(args: &[OsString], session: &mut Session) -> Fallible<Exec
             Platform::current(session)?
         }
     };
-
-    Ok(ToolCommand::new("yarn", args, platform, ToolKind::Yarn).into())
+    
+    Ok(ToolCommand::new(yarn_executable, args, platform, ToolKind::Yarn).into())
 }
 
 /// Determine the execution context (PATH and failure error message) for Yarn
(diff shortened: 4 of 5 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "volta"
-version = "1.1.1"
+version = "1.1.2"
```
rwjblue, line 3: In general, we only bump the version when we do releases.

rwjblue (commented):
This seems totally on the right path! Thanks for working on it!!

I think there are a few things we'd still need to do:

- Ensure we create the `yarnpkg` shim (will require changes in the installer I think?)

https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/dev/unix/install.sh.in?plain=1#L94
https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/dev/unix/volta-install-legacy.sh?plain=1#L77
https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/wix/main.wxs?plain=1#L157-L172

- Add some tests so we don't accidentally break this new feature. I'm not 100% sure where to add the tests. Maybe @chriskrycho or @charlespierce have an idea...

rwjblue (commented):
Also need to make sure `volta run` knows how to handle this correctly.  I think that means adding another match arm here:

https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/crates/volta-core/src/run/mod.rs?plain=1#L99

Some tests are here (you could probably add a couple that call `volta run --yarn=1.7.71 yarnpkg --version` or something):

https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/tests/acceptance/volta_run.rs?plain=1#L357-L433

rytmis:
Interesting! Something gave me a yarnpkg symlink under .volta/bin--but I'll take a look at your links a bit later and see if I can figure what's what.

rwjblue:
> Something gave me a yarnpkg symlink under .volta/bin--but I'll take a look at your links a bit later and see if I can figure what's what.

Most likely that was done by `volta setup` (which your changes to `get_shim_list_deduped` do):

https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/crates/volta-core/src/shim.rs?plain=1#L26-L48

rytmis:
> Also need to make sure volta run knows how to handle this correctly. I think that means adding another match arm here:

Did I mess something up in the PR? I thought I handled it here:

https://github.com/volta-cli/volta/blob/994925bfa7f32cb05610ebd8521fceba137b24ef/crates/volta-core/src/run/mod.rs#L99-L100

And here:

https://github.com/volta-cli/volta/blob/994925bfa7f32cb05610ebd8521fceba137b24ef/crates/volta-core/src/run/yarn.rs#L22

https://github.com/volta-cli/volta/blob/994925bfa7f32cb05610ebd8521fceba137b24ef/crates/volta-core/src/run/yarn.rs#L40

As for the tests--do you prefer a separate test that duplicates the setup, or would a second assertion in the test do?

And finally: yarn 3 doesn't seem to ship the yarnpkg command. That results in a NoPlatform error, which is reported as a missing _node_ installation. The NoPlatform error should probably include the name of the missing component? Should I try to handle the yarn 3 case gracefully, or would the error suffice, if the message was correct?

rwjblue:
> Did I mess something up in the PR? I thought I handled it here:

Nope, you didn't miss anything. I did! 😸 I was jumping between main and your PR, and must have gotten confused about which branch I was on. 🤦 

> As for the tests--do you prefer a separate test that duplicates the setup, or would a second assertion in the test do?

I think additional assertions are fine (though I'll defer to @chriskrycho / @charlespierce if they have a strong preference). 

> And finally: yarn 3 doesn't seem to ship the yarnpkg command. 

Oof. That makes things a bit rough. I do worry about the situation where someone might check (e.g. in a script do `which yarnpkg` as a way to switch or something), but IMHO that's not a great way to test anyways so probably not a giant issue for us here.

> That results in a NoPlatform error, which is reported as a missing node installation. The NoPlatform error should probably include the name of the missing component? Should I try to handle the yarn 3 case gracefully, or would the error suffice, if the message was correct?

Yeah, I think we'll need to tweak this. I'd prefer to have a custom error specifically for this scenario (usage of `yarnpkg` with `yarn@3+`) (and we can throw another specific test for that ([here](https://github.com/volta-cli/volta/blob/889ff9792552bf372387bd02f424b25f7f48a1e6/tests/acceptance/volta_run.rs?plain=1#L357-L433)).

charlespierce:
Regarding the Yarn 3 / `yarnpkg` missing issues: It looks like the Yarn 1 version of `yarnpkg` was only ever a pure passthrough to `yarn`—would it make sense for Volta to treat it the same way? That is, we don't actually change the `yarn::command` function to take in the executable, but continue to always execute `yarn`, regardless of whether it was actually called using `yarn` or `yarnpkg`?

That would mean that `yarnpkg` continues to work on Yarn 3, even though it isn't actually shipped with it, but it would simplify the code a little and remove the issues with the specific version.

rytmis:
That makes more sense than any other alternative I can think of, yes. I reverted the yarn::command change and combined the yarn and yarnpkg arms of the match into one.

rytmis:
Anything else I should do to get this finalized? 

I just verified that this works on Windows--I went down this rabbit hole because I noticed that the lack of a `yarnpkg` shim to Volta was causing issues with caniuse-lite's update command, and lo and behold, that use case works now. :D

rwjblue (approved):
Seems good to me!

@charlespierce / @chriskrycho - Any other changes needed from y'alls perspective?

chriskrycho (approved):
Nope, seems excellent, let's do it! (And sorry for the delay!)

PR #1696, opened 2024-02-16
Create ARM Linux Build using cross-rs

charlespierce:
## Info
* We already support both ARM and x64 for macOS, we can do the same for Linux with `cross-rs` without needing too much extra work.
* Since this is a new build that we haven't previously supported, we don't need to adhere to the same binary compatibility guarantees that we are currently using for x64 Linux.

## Changes
* Updated the `release.yml` GitHub actions file to also generate an ARM build for Linux
    * This takes advantage of `cross-rs`, which makes the compilation process very straightforward and fast.
* Updated `volta-install.sh` to properly handle Linux ARM builds.

## Tested
* Downloaded the artifacts generated in CI and validated that they work in a Linux Docker container on an ARM MacBook.

## Notes
* Do we also need to update `COMPATIBILITY.md` to reflect the slightly more complicated compatibility that we support (since the ARM builds are made on a newer system)?


chriskrycho (approved):
This looks solid overall! I’ll mull on the question re: `COMPATIBILITY.md` but I definitely don’t think we need to block on it. 👍🏼

this-is-tobi:
Hello, thanks for the work, it's exciting to continue using volta on any type of system ! 

Just a question, is there a release planned to get the new `linux-arm64` binary ?

PR #1715, opened 2024-03-18
chore: apply small refactoring

tottoto:
Applies small refactoring.

### DIFF
--- a/crates/archive/src/lib.rs
+++ b/crates/archive/src/lib.rs
@@ -18,7 +18,7 @@ pub enum ArchiveError {
     HttpError(attohttpc::StatusCode),
 
     #[error("HTTP header '{0}' not found")]
-    MissingHeaderError(String),
+    MissingHeaderError(&'static attohttpc::header::HeaderName),
 
     #[error("unexpected content length in HTTP response: {0}")]
     UnexpectedContentLengthError(u64),
--- a/crates/volta-core/src/tool/node/resolve.rs
+++ b/crates/volta-core/src/tool/node/resolve.rs
@@ -157,15 +157,10 @@ fn read_cached_opt(url: &str) -> Fallible<Option<RawNodeIndex>> {
     })?;
 
     if let Some(date) = &expiry {
-        let expiry_date = Expires::decode(
-            &mut [date
-                .parse()
-                .with_context(|| ErrorKind::ParseNodeIndexExpiryError)?]
-            .iter(),
-        )
-        .with_context(|| ErrorKind::ParseNodeIndexExpiryError)?;
+        let expiry_date = httpdate::parse_http_date(date)
+            .with_context(|| ErrorKind::ParseNodeIndexExpiryError)?;
 
-        let current_date = Expires::from(SystemTime::now());
+        let current_date = SystemTime::now();
 
         if current_date < expiry_date {
             let index_file = volta_home()?.node_index_file();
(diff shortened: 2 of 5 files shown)


crates/archive/src/lib.rs
```diff
@@ -18,7 +18,7 @@ pub enum ArchiveError {
     HttpError(attohttpc::StatusCode),
 
     #[error("HTTP header '{0}' not found")]
-    MissingHeaderError(String),
+    MissingHeaderError(attohttpc::header::HeaderName),
```
charlespierce, line 21: It looks like the [`Header::name`](https://docs.rs/headers/latest/headers/trait.Header.html#tymethod.name) associated function returns a `&'static` reference to the `HeaderName` object. Given that, we can probably use the same here (`&'static HeaderName`) instead of needing to convert it to an owned value.

crates/volta-core/src/tool/node/resolve.rs
```diff
@@ -156,34 +156,28 @@ fn read_cached_opt(url: &str) -> Fallible<Option<RawNodeIndex>> {
...
-        .with_context(|| ErrorKind::ParseNodeIndexExpiryError)?;
-
-        let current_date = Expires::from(SystemTime::now());
+    if !expiry
+        .map(|date| httpdate::parse_http_date(&date))
+        .transpose()
+        .with_context(|| ErrorKind::ParseNodeIndexExpiryError)?
+        .is_some_and(|expiry_date| SystemTime::now() < expiry_date)
+    {
+        return Ok(None);
```
charlespierce, line 165: It's definitely possible I'm misreading, but I believe this might be inverting the condition. In the current code, when `SystemTime::now() < expiry_date`, then we read the cached file. However, in this one we're returning no cache with that same condition.

charlespierce (changes requested):
Nice, I really appreciate cleanups like this! A couple small comments to make sure the behavior is correct.

Separately, is there a way we can confirm the value that we're writing is parseable by `httpdate`? While the current approach is convoluted, it is making sure to compare apples to apples by using the `Expires` struct directly for the comparison.

It actually looks like the `Expires` header provides an implementation of `From<Expires> for SystemTime`, so we might be able access the time directly, rather than needing to parse the string value.

Another important piece is we want to make sure any comparison is backwards-compatible with the existing expiry file.

crates/archive/src/lib.rs
tottoto: Updated to use `&'static HeaderName`.

crates/volta-core/src/tool/node/resolve.rs
tottoto: I would think the behavior would be the same as the code before this change as the negation of this condition is evaluated.

tottoto:
> Separately, is there a way we can confirm the value that we're writing is parseable by httpdate? While the current approach is convoluted, it is making sure to compare apples to apples by using the Expires struct directly for the comparison.

As `headers::Expires` and `httpdate` both support HTTP-date format defined in rfc7231, they are compatible.

> It actually looks like the Expires header provides an implementation of From<Expires> for SystemTime, so we might be able access the time directly, rather than needing to parse the string value.

Yes. When using `Expires`, it is possible to evaluate like this `SystemTime::now() < expiry_date.into()`.

> Another important piece is we want to make sure any comparison is backwards-compatible with the existing expiry file.

Since `headers` crate actually uses `httpdate` to parse and format date and delegate the comparison to internal  `httpdate` implementation, and the `httpdate` uses `SystemTime` for its comparison, it behaves exactly the same.

charlespierce (approved):
LGTM, thanks!

PR #1751, opened 2024-05-13
Add test fixtures for Apple Silicon architecture.

charlespierce:
Info
-----
Our test fixtures do not currently include "versions" for Apple silicon machines. However, recently GitHub actions switched the default macOS runner to be an M1 machine. This caused our CI to fail when setting up the mocks because a fixture wasn't available.

Changes
-----
* Added Node fixtures for Apple ARM architectures

Tested
-----
* Need to test in CI 😄


chriskrycho (approved):
💙 thank you for fixing these. I especially like that the result is simpler and more in line with the internals.

PR #1753, opened 2024-05-13
Update cli_platform tests to use 'const' correctly

charlespierce:
Info
-----
The tests of the platform resolution for the various 'volta run' scenarios were using a lazily-loaded 'const' value. Clippy raised some warnings about those, as 'const' values are actually duplicated for each location they are used, so the interior mutability wasn't helpful.

It seems that the `Lazy` behavior was actually a way to work around the fact that `Version` doesn't have a 'const' constructor, however all of the fields are `pub`, so we can construct the values directly. This allows us to have a true `const` value and use it in each test without needing extra interior mutability or cloning.

Changes
-----
* Updated the `NODE_VERSION`, `NPM_VERSION`, and `YARN_VERSION` constants used in the `cli_platform` tests to be directly defined, rather than lazy loaded.
* Changed all call sites to use the constants directly, instead of cloning.

Tested
-----
* Ran the affected tests and verified that they still pass as expected.

### DIFF
--- a/crates/volta-core/src/platform/tests.rs
+++ b/crates/volta-core/src/platform/tests.rs
@@ -215,11 +215,28 @@ mod inherit_option {
 
 mod cli_platform {
     use node_semver::Version;
-    use once_cell::unsync::Lazy;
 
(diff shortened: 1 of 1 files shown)


chriskrycho (approved):
Love it!

PR #1755, opened 2024-05-16
Replace Windows symlinks with `volta run` scripts

### LINKED ISSUE #1397: Dismiss the use of symlinks on Windows

charlespierce:
Closes #1397 

Info
-----
* Huge kudos to @rcsilva83 and @Xstoudi for a bunch of iteration in #1552 on removing symlinks for Windows, to allow dropping the requirement for Developer Mode.
* Based on discussions in that PR, we had a new idea for replacing symlink shims on Windows by using scripts that call `volta run <package>`.
* This is a first-pass implementation of creating simple script-based shims for executing 3rd-party binaries on Windows.
* This should be a no-op for Unix.

Changes
-----
* Used `junction` to create NTFS junctions instead of directory symlinks (this part is directly copied from #1552)
* Updated the `shim` module to have different implementations of `create` for Unix and Windows.
    * The Unix implementation is the same as it was before.
    * The Windows implementation creates a very simple `.cmd` script that calls `volta run` for the shim. It also continues to create a custom Git Bash script, though that script is updated to now call `volta run` directly, instead of calling the existing shim via `cmd.exe`
* Create a new `v4` layout since the shim file on Windows is now `<binary>.cmd` instead of `<binary>.exe`.
* Created a migration to `v4` that removes and regenerates the shims, making sure that the old symlinks are removed. The migration also updates any existing directory symlinks to use junctions as well.

Tested
-----
* Installed packages using the current Volta version (i.e. with symlinks).
* Installed a new version of Volta based on these changes.
* Verified that running the existing packages still works and that the shims were properly migrated.
* Turned off developer mode.
* Verified that even without developer mode, I could install and run new packages 🎉 
* Did some very basic performance tests, on my local machine I wasn't seeing any statistically significant differences between the symlink-based approach and the `volta run` approach ⚡

### DIFF
--- a/crates/volta-layout/src/v4.rs
+++ b/crates/volta-layout/src/v4.rs
@@ -1,0 +1,125 @@
+use std::path::PathBuf;
+
+use volta_layout_macro::layout;
+
+pub use crate::v1::VoltaInstall;
+
+layout! {
+    pub struct VoltaHome {
+        "cache": cache_dir {
+            "node": node_cache_dir {
+                "index.json": node_index_file;
+                "index.json.expires": node_index_expiry_file;
+            }
+        }
+        "bin": shim_dir {}
+        "log": log_dir {}
+        "tools": tools_dir {
+            "inventory": inventory_dir {
+                "node": node_inventory_dir {}
+                "npm": npm_inventory_dir {}
+                "pnpm": pnpm_inventory_dir {}
+                "yarn": yarn_inventory_dir {}
+            }
+            "image": image_dir {
+                "node": node_image_root_dir {}
+                "npm": npm_image_root_dir {}
+                "pnpm": pnpm_image_root_dir {}
+                "yarn": yarn_image_root_dir {}
+                "packages": package_image_root_dir {}
+            }
+            "shared": shared_lib_root {}
+            "user": default_toolchain_dir {
+                "bins": default_bin_dir {}
+                "packages": default_package_dir {}
+                "platform.json": default_platform_file;
+            }
+        }
+        "tmp": tmp_dir {}
+        "hooks.json": default_hooks_file;
+        "layout.v4": layout_file;
+    }
+}
+
+impl VoltaHome {
+    pub fn node_image_dir(&self, node: &str) -> PathBuf {
+        path_buf!(self.node_image_root_dir.clone(), node)
+    }
+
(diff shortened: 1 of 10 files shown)


Xstoudi:
Hey @charlespierce, did you find time to check how it affects performance?
How much time do you think is needed to merge/release?

charlespierce:
Was able to get some testing done, and from what I can tell it doesn't have any major effect on performance, so this should be good to go 🎉

crates/volta-layout/src/v4.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    pub fn shim_file(&self, toolname: &str) -> PathBuf {
+        // On Windows, shims are created as `<name>.cmd` since they
+        // are thin scripts that use `volta run` to execute the command
+        #[cfg(windows)]
+        let toolname = format!("{}{}", toolname, ".cmd");
+
+        path_buf!(self.shim_dir.clone(), toolname)
+    }
```
chriskrycho, line 102: For posterity, this is the only difference between the v3 and v4 layouts, and this is the diff:

```diff
    pub fn shim_file(&self, toolname: &str) -> PathBuf {
-       path_buf!(self.shim_dir.clone(), executable(toolname))
+       // On Windows, shims are created as `<name>.cmd` since they
+       // are thin scripts that use `volta run` to execute the command
+       #[cfg(windows)]
+       let toolname = format!("{}{}", toolname, ".cmd");
+
+       path_buf!(self.shim_dir.clone(), toolname)
    }
```

chriskrycho (approved):

Issue #577, opened 2019-10-18
Running 'node' on Windows can show unhelpful error


charlespierce:
If you don't have a default Node selected and have no Node available elsewhere on the system, running `node` on Windows results in the error message:

```cmd
'node' is not recognized as an internal or external command, operable program or batch file
```

I believe this is the result of the passthrough calling `node` with the non-Volta path. On Unix, calling a non-existent command results in the call to `Command::status` returning an `Err`. On Windows, it results in an `Ok` with a status code of 1. If possible, we should detect this situation and show the Volta error.

One possibility would be to use `which` to see if there is something on the system path named `node` and only pass through if that succeeds.

charlespierce:
Thinking more, this may be a combination of the passthrough command and the fact that on Windows we currently always use `cmd.exe /C <command>` when executing a Command. For passthrough cases, it may be better to create a direct command.

charlespierce:
Further investigation reveals the above solution won't work. Apparently by default, Windows uses the `PATH` from the parent process (in this case the Volta shim) to look up an executable when using `std::process::Command` (see https://github.com/rust-lang/rust/issues/15149). In order to combat that, if you set a `PATH` environment variable, Rust internally does a lookup to try to find the executable on the _new_ `PATH`, and if it does find it then it uses it directly (see https://github.com/rust-lang/rust/blob/master/src/libstd/sys/windows/process.rs#L133-L149).

However, in the case of passing through to a `node` that doesn't exist, Rust looks for `node` and doesn't find it, so defaults to Windows' normal behavior, which uses the Volta shim's PATH to find `node`, meaning it finds the shim itself again, and again, and again, until the OS breaks the chain of nested calls. When we use `cmd.exe /C <command>`, we get an extra layer that makes the PATHs work out, but we run into the error message issues.

To properly support passthrough with correct PATH semantics _and_ good error messages, we may need to create our own windows Command wrapper that provides a similar layer as calling `cmd.exe`, but also provides a method for informing the top-level Volta process of it's inability to locate a command.

PR #1761, opened 2024-05-23
fix: Prevent fork bomb on Windows

### LINKED ISSUE #1741: Potential fork bomb
Executables shipped with volta may spawn tons of process and drain out memories.

Steps to reproduce:
1. Install volta.
2. Install node using `volta install node`.
3. Run node.exe from the Explorer.

Environment:
Operating System: Windows 11 23H2
Volta: v1.1.1
Node: 20.12.2 (latest)
npm: 10.5.0 (shipped with Node)

Screenshot:

Log:
```log
"node"
Volta v1.1.1

Node is not available.

To run any Node command, first set a default version using `volta install node`

Error cause: The paging file is too small for this operation to complete. (os error 1455)
```

chawyehsu:
Fix #1741

Before this patch, `volta-shim`(and its aliases) spawns itself in some cases causing recursive call (fork bomb). This is a critical issue but only occurs on Windows.


charlespierce:
Thanks for tackling this! Can you comment on the specific trigger condition and the mechanics of how it recurses? Without a good understanding of the overall root cause of the recursion, it's difficult to be confident that we're completely resolving the issue.

We're already working around some of the oddities of `Command` on Windows a bit by calling `cmd /C <tool>` instead of `<tool>` directly (see the behavior of the `create_command` function). I have a hypothesis that the underlying issue is when the Volta directory itself is the CWD for the command (which I think would be the case when you launch it from the file explorer). So even though we remove the Volta directory from the `Path`, it's still the "local" executable and is first place Windows looks to resolve the name.

If that's the case, we may be able to resolve this in a less intrusive way by detecting the CWD is one of the Volta dirs. I worry a little about the extra copying / allocating needed to hold onto the command, args, and env vars through the process. It's a relatively small impact, but this block is on the hot path for every call to any of the shims, so it's one of the most performance-sensitive areas of the app.

charlespierce:
One thought, if the above is accurate for how this recursion starts, would be to check on Windows right before executing if the CWD is one of the Volta directories. If it is, we can do the absolute path resolution and then copy the values (since we’ll need to create a new `Command` at that point regardless). Helpfully, `Command` itself has methods to get the root command, the arguments, and the environment variables, so we don’t need to copy ahead of time, we can read them from the already-built command to build a new one (see e.g. [`get_args`](https://doc.rust-lang.org/std/process/struct.Command.html#method.get_args))

chawyehsu:
> We're already working around some of the oddities of `Command` on Windows a bit by calling `cmd /C <tool>` instead of `<tool>` directly (see the behavior of the `create_command` function). I have a hypothesis that the underlying issue is when the Volta directory itself is the CWD for the command (which I think would be the case when you launch it from the file explorer). So even though we remove the Volta directory from the `Path`, it's still the "local" executable and is first place Windows looks to resolve the name.

I see that it had tried to remove _the Volta directory from the `Path`_ via `System::path()`, but that's not enough. **The udnerlying `CreateProcessW` searches hard-coded locations before the `PATH` env**:

> If the file name does not contain a directory path, the system searches for the executable file in the following sequence:
> 
> 1. The directory from which the application loaded.
> 2. The current directory for the parent process.
> 3. The 32-bit Windows system directory. Use the [GetSystemDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsystemdirectorya) function to get the path of this directory.
> 4. The 16-bit Windows system directory. There is no function that obtains the path of this directory, but it is searched. The name of this directory is System.
> 5. The Windows directory. Use the [GetWindowsDirectory](https://learn.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getwindowsdirectorya) function to get the path of this directory.
> 6. The directories that are listed in the PATH environment variable. Note that this function does not search the per-application path specified by the App Paths registry key. To include this per-application path in the search sequence, use the [ShellExecute](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/axe/shellexecute) function.

That's the root cause.

> we may be able to resolve this in a less intrusive way by detecting the CWD is one of the Volta dirs.

I get your thought here but it does not really help imo by just detecting CWD. It should only find the right executable by only checking from the PATH environment variable we give. _(and this is why `which` crate was introduced)_

chawyehsu:
Btw, I don't quite understand why the _infinite_ recursive call is allowed. I believe there should be a limit of max recursive depth, say 10, and once it exceeds it should halt. This should help to resolve from another perspective.

crates/volta-core/src/error/kind.rs
```diff
@@ -408,6 +408,9 @@ pub enum ErrorKind {
         file: PathBuf,
     },
 
+    /// Throw when recursion limit is reached
```
chriskrycho, line 411:
```suggestion
    /// Thrown when recursion limit is reached
```

tests/acceptance/volta_bypass.rs
```diff
@@ -15,19 +15,10 @@ fn shim_skips_platform_checks_on_bypass() {
...
             .with_stderr_contains("VOLTA_BYPASS is enabled[..]")
     );
-
-    #[cfg(windows)]
-    assert_that!(
-        s.process(shim_exe()),
-        execs()
-            .with_status(ExitCode::UnknownError as i32)
-            .with_stderr_contains("[..]is not recognized as an internal or external command[..]")
-    );
```
chriskrycho, line 32: Why is this test no longer valid? (To be extra clear, I am not saying you’re wrong about the change, it just is not at all obvious to me why it is being made!)

crates/volta-core/src/command.rs
```diff
@@ -28,3 +31,52 @@ cfg_if! {
...
+    {
+        let args = command.get_args().collect::<Vec<_>>();
+        //          cmd /c <name> [...other]
+        // args_idx     0  1      2..
+        let name = args.get(1).expect("should have the name");
+
+        if let Ok(mut paths) = which::which_in_global(name, Some(&path)) {
+            if let Some(exe) = paths.next() {
+                let mut new_command = create_command(exe);
+                let envs = command
```
chriskrycho, line 63: 1. Why does this need to filter the environment variables like this, rather than maintaining the same environment variables as the previous version of the command?
2. Can you add a comment explaining that here? That will help us remember (and other contributors understand)!

crates/volta-core/src/command.rs
```diff
@@ -28,3 +31,52 @@ cfg_if! {
...
+                    .get_envs()
+                    .filter(|(_, v)| v.is_some())
+                    .map(|(k, v)| (k, v.unwrap()))
+                    .collect::<Vec<_>>();
+
+                new_command.args(&args[2..]);
```
chriskrycho, line 69:
```suggestion
                // The args will be the command name and any additional args.
                new_command.args(&args[2..]);
```

crates/volta-core/src/command.rs
```diff
@@ -28,3 +31,52 @@ cfg_if! {
...
+
+    #[cfg(windows)]
+    {
+        let args = command.get_args().collect::<Vec<_>>();
+        //          cmd /c <name> [...other]
+        // args_idx     0  1      2..
+        let name = args.get(1).expect("should have the name");
+
+        if let Ok(mut paths) = which::which_in_global(name, Some(&path)) {
+            if let Some(exe) = paths.next() {
```
chriskrycho, line 61: As a stylistic thing (not a blocker!) since we're on a version of Rust which supports it, I prefer the `let … else` form, which flattens the code flow out a lot. It *does* require us to write the error out twice in that case, though, so it’s a small tradeoff around readability!

chriskrycho (changes requested):
Here’s a first pass review with some changes requested. @charlespierce may have some more insight on the Windows-specific bits!

crates/volta-core/src/command.rs
charlespierce: Agreed on adding a comment!

I believe this needs to filter because the output of [`Command::get_envs`](https://doc.rust-lang.org/std/process/struct.Command.html#method.get_envs) includes `None` values for variables that were explicitly removed. I don't think we do any explicit removing of environment variables currently, but possibly what we want to do is mirror the same behavior:

```rust
for (key, maybe_value) in command.get_envs() {
    match maybe_value {
        Some(value) => command.env(key, value),
        None => command.env_remove(key),
    }
}
```

crates/volta-core/src/command.rs
```diff
@@ -28,3 +31,52 @@ cfg_if! {
...
+    }
+
+    #[cfg(windows)]
+    {
+        let args = command.get_args().collect::<Vec<_>>();
+        //          cmd /c <name> [...other]
+        // args_idx     0  1      2..
+        let name = args.get(1).expect("should have the name");
+
+        if let Ok(mut paths) = which::which_in_global(name, Some(&path)) {
```
charlespierce, line 60: Given that the trigger for this (as I understand it) is the `cmd /c` implementation searching the CWD for the binary, even when it's not explicitly on the `Path`, would it be unreasonable to gate this reformulation of the `Command` on that, rather than doing it every time on Windows?

It's relatively minor, but the `which_in_global` is going to necessarily do a decent chunk of File I/O for every command run, which seems excessive for something on the hot path.

crates/volta-core/src/command.rs
```diff
@@ -28,3 +31,52 @@ cfg_if! {
...
+            if let Some(exe) = paths.next() {
+                let mut new_command = create_command(exe);
+                let envs = command
+                    .get_envs()
+                    .filter(|(_, v)| v.is_some())
+                    .map(|(k, v)| (k, v.unwrap()))
```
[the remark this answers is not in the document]
charlespierce: Or alternatively:

```rust
.filter_map(|(k, maybe_v)| Some(k).zip(maybe_v))
```

crates/volta-core/src/error/kind.rs
```diff
@@ -1246,6 +1249,10 @@ from {}
                 file.display(),
                 PERMISSIONS_CTA
             ),
+            ErrorKind::RecursionLimit => write!(
+                f,
+                "Recursive call limit reached."
+            ),
```
charlespierce, line 1255: I don't have a great suggestion on the top of my head, but is it impossible to create a call-to-action for this error? We have generally tried to always have a CTA included with any error message, so that we guide users to how to fix the problems they run into.

tests/acceptance/volta_bypass.rs
charlespierce: I actually think this isn't that the test is invalid, but rather that the Windows behavior now accurately matches the Unix behavior, so they don't need to have separate expectations.

charlespierce (changes requested):
Thanks for continuing on this! A few additional requests for cleanup.

Separately, and I haven't totally thought this through, but would it be unreasonable to directly error if one of the commands is run from either of the Volta directories? That might simplify the solution a lot, at the expense of making some specific uses errors when they otherwise wouldn't. However, those cases are currently fork bombs, so it would be an improvement even if it isn't a complete fix.

tests/acceptance/volta_bypass.rs
chawyehsu: @chriskrycho, @charlespierce was right on this.

chriskrycho: Excellent – thank you!

charlespierce:
Doing some spelunking in our issues (while investigating some other, ostensibly unrelated, Windows issues), and I've come around on this approach. I actually think we might be able to do even better and use the behavior of `which` to completely skip needing to add `cmd.exe /C` entirely. It's been ~5 years, but looking back at some of our issues (notably #577), I think the reason we use `cmd.exe /C` currently is twofold:

1. To work around issues with Rust's `Command` implementation around calling `.cmd` and `.bat` files without extensions.
2. To prevent fork bombs like this when calling a package that happens to not exist

Bypassing `cmd.exe` entirely, and using `which` directly to resolve the executable would solve both of those problems - `which` already handles detecting the correct file extension and returning it if it's not found.

Additionally, this would solve the linked issue around error messages, because if we can't locate the executable with `which`, then we can know (even without needing to try to spawn the process) that it won't work and can show a helpful error message instead if a generic `cmd.exe` `node is not a recognized command...` error.

It also (I _think_) would solve argument parsing issues like those mentioned in #1791, so this may be an altogether cleaner path forward.

Thanks for suggesting and pushing on this @chawyehsu!

chawyehsu:
Thanks for the informative response, would love to see improvement to this PR to make it a general solution to linked issues. @charlespierce

PR #1778, opened 2024-06-30
chore: update dependency to reduce transitive dependency

tottoto:
Updates some dependencies to reduce transitive dependencies.


chriskrycho (approved):
💙 love to see red in the lock file. Thank you!

Issue #1744, opened 2024-04-26
When trying to install new versions volta hangs indefinitely and blocks all other network requests from resolving


effervescentia:
Websites will not load and I have not waited long enough for the command to ever resolve or error (I've waited over 10 minutes)

I was able to solve this temporarily for `node` downloads by configuring the `hooks.json` file with a mirror

```json
{
    "node": {
        "index": {
            "template": "https://mirrors.dotsrc.org/nodejs/release/index.json"
        },
        "distro": {
            "template": "https://mirrors.dotsrc.org/nodejs/release/v{{version}}/{{filename}}"
        }
    }
}
```

this is also happening when trying to install new versions of `yarn`
I first encountered this close to 2 or 3 weeks ago and it has not yet been resolved
it doesn't looks like a new version of `volta` has been released, but perhaps some certificates have expired?

if I open Activity Monitor and quit `volta-shim` my website loads immediately

**OS**: MacOS 14.4.1

effervescentia:
as a follow up I wasn't able to find a relevant mirror for `yarn` but I followed [this comment](https://github.com/volta-cli/volta/issues/651#issuecomment-818927941) and used the releases from github as distro template

```json
{
    "node": {
        "index": {
            "template": "https://mirrors.dotsrc.org/nodejs/release/index.json"
        },
        "distro": {
            "template": "https://mirrors.dotsrc.org/nodejs/release/v{{version}}/{{filename}}"
        }
    },
    "yarn": {
        "distro": {
            "template": "https://raw.githubusercontent.com/yarnpkg/berry/%40yarnpkg/cli/{{version}}/packages/yarnpkg-cli/bin/yarn.js"
        }
    }
}
```

xavidop:
this happens to me as well!

edison-vflow:
I am also experiencing this issue
@Volta could you please help resolve

charlespierce:
This is extremely odd, thank you all for reporting! Is this still happening and consistent? And does it happen trying to access the paths directly via `curl` or the browser (e.g. `https://nodejs.org/dist/index.json`)?

gillyb:
Happening to me too..  
Trying to reach that url from the browser works for me.

effervescentia:
this is definitely still happening, and I don't have any issues curling that URL
It's also happening with `yarn` releases, so nothing specific to the NodeJS download servers

effervescentia:
@charlespierce anything else I can do to help find the source of the issue?

[...]

PR #1784, opened 2024-07-08
debugging: add a bunch of tracing around `archive`

chriskrycho:
## Background/Notes

This is not really intended to be merged as is, but it should allow folks to do some testing and see if we can get more info for #1744.

In the interest of *possibly* being able to build on this for merging it if we find that useful in the future, I used `trace!` rather than `debug!` for this extreme level of output, and added a `--very-verbose` flag for this mode.

**Note:** This is based on v1.1.1, *not* on the latest `main`, and I have intentionally marked it as a draft until we get more info from #1744.

## To Test

```sh
git clone https://github.com/volta-cli/volta.git
cd volta
git switch chriskrycho/debug-1744
cargo build --release
mv $HOME/.volta/bin/volta $HOME/.volta/bin/volta-backup
cp ./target/release/volta $HOME/.volta/bin/volta
volta --verbose --very-verbose
```

That final line should print the normal help output *preceded* by this line:

```
[very verbose] log level: VeryVerbose
```

(If you have a non-standard installation location, you will want to figure out where it is with `which volta` and do the same basic thing for the final two steps, but with the appropriate directory changes.)

Once you have that set up, you should be able to get much more info about where in the archive/unpacking process things are going wrong.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,6 +55,7 @@ dependencies = [
  "flate2",
  "fs-utils",
  "hyperx",
+ "log",
  "progress-read",
  "tar",
  "tee",
--- a/crates/archive/Cargo.toml
+++ b/crates/archive/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2018"
 
 [dependencies]
 flate2 = "1.0"
+log = { version = "0.4", features = ["std"] }
 tar = "0.4.13"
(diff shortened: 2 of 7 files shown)


chriskrycho:
Closing this as we have (at least theoretically!) resolved the issue it was intended to help debug: #1744. The useful-for-the-future parts have separately been extracted to #1793 and #1815.

Issue #1760, opened 2024-05-23
uninstall failed


lazy-coder97:
<img width="462" alt="image" src="https://github.com/volta-cli/volta/assets/134416734/e50ec580-ebfa-4688-a23c-72e3dba85633">
Why would it say the package is not found?

tang-nn:
![image](https://github.com/volta-cli/volta/assets/98725149/4749ed83-e05f-4f80-89a4-e89d09de606e)

chriskrycho:
Yarn is not a package, but rather a package manager, and we do not currently support uninstalling package managers – see #330 and especially #327 for further details. Closing this as a duplicate of #327 specifically! (That said, I think we could maybe help here with some better messaging!)

PR #1786, opened 2024-07-09
Better error messages for unsupported `uninstall` arguments

chriskrycho:
Instead of parsing whatever string the user supplied as the tool name and supplying a default `VersionSpec`, attempt to parse the value as a full specifier which may include a version. This means we can provide better errors when the user passes a version. Specifically we can report that Volta does not yet support uninstalling specific versions of tools. Previously, we would report something like this:

```
warning:  No package 'typescript@5.4' found to uninstall
```

Notice that the old message treated `'typescript@5.4'` as the name of the tool, when it should have been treating it as a tool and a version specifier. Now, we instead report that uninstalling specific versions of tools is unsupported.

The original motivation here was noticing that we printed errors like that if the user tried to uninstall a runtime or a package manager with a version specifier. This fixes that as well, since it no longer parses a string like `node@20.14.5` as a package, but rather as a runtime and version specifier, and can fall into the normal handling for runtimes.

## Background and Miscellanies.

- I noticed this when triaging #1760.
- I also got a bit lost trying to figure out what `volta_uninstall` vs. `direct_uninstall` meant, so I commented those.
- Likewise with some typo fixes!

### DIFF
--- a/crates/volta-core/src/error/kind.rs
+++ b/crates/volta-core/src/error/kind.rs
@@ -443,10 +443,10 @@ pub enum ErrorKind {
         name: String,
     },
 
-    /// Thrown when serializnig a bin config to JSON fails
+    /// Thrown when serializing a bin config to JSON fails
     StringifyBinConfigError,
 
-    /// Thrown when serializnig a package config to JSON fails
+    /// Thrown when serializing a package config to JSON fails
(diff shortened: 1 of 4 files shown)


charlespierce (approved):
LGTM! Love to have better / clearer error messages 💥

PR #1794, opened 2024-07-12
Simplify the Dependabot config

chriskrycho:
Since we added Dependabot a few years ago, they added support for this simpler way of configuring multiple directories for the same package ecosystem (see [Directories][docs] in the docs), including support for globbing.

[docs]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,61 +1,9 @@
 version: 2
 updates:
-  - directory: /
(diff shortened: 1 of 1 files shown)


charlespierce (approved):
🎉 Much cleaner!

PR #1801, opened 2024-07-16
Support Windows Arm64 Build

shibayan:
Relate #1270 

A build for Windows on Arm will be added to Volta.

The changes I made include creating an msi and zip for Arm64 in my repository, and I have confirmed that it works on Windows on Arm machines.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -148,27 +148,27 @@ jobs:
           TAG: ${{ github.ref }}
         run: echo "version=${TAG:11}" >> $GITHUB_OUTPUT
       - name: Fetch Linux artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v2
         with:
           name: linux
           path: release
       - name: Fetch Linux ARM artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v2
         with:
           name: linux-arm
           path: release
       - name: Fetch MacOS artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v2
         with:
           name: macos
           path: release
       - name: Fetch Windows installer
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v2
         with:
           name: windows-installer
           path: release
       - name: Fetch Windows zip
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v2
         with:
           name: windows-zip
           path: release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,13 +100,44 @@ jobs:
           name: windows-zip
           path: target/release/volta-windows.zip
 
+  windows-arm:
+    name: Build - Windows ARM
+    runs-on: windows-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Set up cargo
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          target: aarch64-pc-windows-msvc
+      - name: Add cargo-wix subcommand
+        run: cargo install cargo-wix
+      - name: Compile and package installer
+        run: |
+          cargo wix --nocapture --package volta --target aarch64-pc-windows-msvc --output target\wix\volta-windows-arm.msi
+      - name: Create zip of binaries
+        run: powershell Compress-Archive volta*.exe volta-windows-arm.zip
+        working-directory: ./target/aarch64-pc-windows-msvc/release
+      - name: Upload installer
+        uses: actions/upload-artifact@v4
+        with:
+          name: windows-installer-arm
+          path: target/wix/volta-windows-arm.msi
+      - name: Upload zip
+        uses: actions/upload-artifact@v4
+        with:
+          name: windows-zip-arm
+          path: target/aarch64-pc-windows-msvc/release/volta-windows-arm.zip
+
   release:
     name: Publish release
     runs-on: ubuntu-latest
     needs:
       - linux
+      - linux-arm
       - macos
       - windows
+      - windows-arm
     if: github.event_name == 'push'
     steps:
       - name: Check out code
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -117,30 +148,40 @@ jobs:
           TAG: ${{ github.ref }}
         run: echo "version=${TAG:11}" >> $GITHUB_OUTPUT
       - name: Fetch Linux artifact
-        uses: actions/download-artifact@v2
+        uses: actions/download-artifact@v4
```
chriskrycho, line 151: Let's do this separately! I am happy to take care of that, though it will be this evening or later this week; and you can merge those changes in or rebase on top of them once we have!

shibayan: Thanks, I understand. In this PR, I revert the version change of the download-artifact and create a separate PR to address it.

.github/workflows/release.yml
```diff
@@ -100,13 +100,44 @@ jobs:
...
+        with:
+          name: windows-zip-arm
+          path: target/aarch64-pc-windows-msvc/release/volta-windows-arm.zip
+
   release:
     name: Publish release
     runs-on: ubuntu-latest
     needs:
       - linux
+      - linux-arm
```
chriskrycho, line 137: I don’t think we should enable `linux-arm` in this PR. If we choose to enable it before shipping 2.0, we should do so as part of its own PR!

```suggestion
```

wix/main.wxs
```diff
@@ -21,11 +21,9 @@
   destination for 32-bit versus 64-bit installers. Removal of these lines will
   cause installation errors.
 -->
-<?if $(var.Platform) = x64 ?>
-    <?define Win64 = "yes" ?>
+<?if $(sys.BUILDARCH) = x64 or $(sys.BUILDARCH) = arm64 ?>
     <?define PlatformProgramFilesFolder = "ProgramFiles64Folder" ?>
 <?else ?>
-  <?define Win64 = "no" ?>
```
chriskrycho, line 28: Why the `Win64` changes here?

wix/main.wxs
```diff
@@ -192,9 +191,11 @@
             <ComponentGroupRef Id='Binaries'/>
         </Feature>
 
+        <?if $(sys.BUILDARCH) = x64 ?>
         <Feature Id='VCRedistributable' Title='Visual C++ Runtime' AllowAdvertise='no' Display='hidden' Level='1'>
             <MergeRef Id='VCRedist'/>
         </Feature>
+        <?endif ?>
```
chriskrycho, line 198: Likewise, why the addition of the `BUILDARCH` check here?

wix/main.wxs
```diff
@@ -44,12 +42,11 @@
             Keywords='Installer'
             Description='The JavaScript Launcher'
             Manufacturer='The Volta Maintainers'
-            InstallerVersion='450'
+            InstallerVersion='500'
             Languages='1033'
             Compressed='yes'
             InstallScope='perMachine'
-            SummaryCodepage='1252'
-            Platform='$(var.Platform)'/>
```
chriskrycho, line 52: Why are you dropping the `Platform` value here?

wix/main.wxs
```diff
@@ -62,16 +59,18 @@
                 <Directory Id='INSTALLDIR' Name='Volta'>
                 </Directory>
             </Directory>
+            <?if $(sys.BUILDARCH) = x64 ?>
             <Merge Id='VCRedist' SourceFile='wix\Microsoft_VC140_CRT_x64.msm' DiskId='1' Language='0'/>
+            <?endif ?>
```
chriskrycho, line 64: Why are you dropping this check?

wix/main.wxs
```diff
@@ -62,16 +59,18 @@
...
         </Directory>
 
         <ComponentGroup Id='Binaries' Directory='INSTALLDIR'>
-            <Component Id='voltaBinary' Guid='*' Win64='$(var.Win64)'>
+            <Component Id='voltaBinary' Guid='*'>
```
chriskrycho, line 68: Here and throughout, why are you removing `Win64=`?

chriskrycho (commented):
This looks reasonable to me overall! I had a bunch of questions, but they are questions where I literally just do not understand the reasoning, so do not take them as implying you’re *wrong* in making those changes—merely that I would like to understand them and make sure we have the reason for the change captured for posterity!

Also, @charlespierce can you take a gander here?

wix/main.wxs
shibayan: Added a condition that the VC++ redistribution runtime should not be included on arm64 since it was included for x64.

However, I do not know why the redistribution runtime is included. In my opinion, it seems unnecessary.

wix/main.wxs
shibayan: The `Win64` and `Platform` settings have been removed as they are now deprecated in wix. The details are described in the following pull request from cargo-wix, which is shared here.

https://github.com/volks73/cargo-wix/pull/113#issue-723263542

>I took the opportunity to remove use of Win64 and Platform in the wxs: those attributes are deprecated according to the wix documentation, passing -arch on the command line should be prefered (which this PR now does). var.Platform is replaced by sys.BUILDARCH.

wix/main.wxs
shibayan: I am deleting it for the same reason as here: https://github.com/volta-cli/volta/pull/1801#discussion_r1687443667

wix/main.wxs
shibayan: I am deleting it for the same reason as here: https://github.com/volta-cli/volta/pull/1801#discussion_r1687443667

wix/main.wxs
shibayan: Since the VC++ redistribution runtime was only available for x64, I have added a condition so that it is only included for x64 as before.

wix/main.wxs
charlespierce: It looks like that was added in #592, as it's needed to actually run Volta and not always available on the system.

charlespierce (approved):
LGTM! We've never actually supported / done builds for 32-bit Windows anyway, so the removal of the 64-bit checks is even more welcome!

wix/main.wxs
shibayan: I looked into it because I couldn't find a merge module for Arm64, and it seems that it is now deprecated.

>Merge modules (.msm files) for Visual C++ Redistributable files are deprecated. We don't recommend you use them for application deployment. Instead, we recommend central deployment of the Visual C++ Redistributable package.

https://learn.microsoft.com/en-us/cpp/windows/redistributing-visual-cpp-files?view=msvc-170#install-the-redistributable-merge-modules

chriskrycho:
Excellent – and @shibayan answered all my questions (thank you!).

PR #1802, opened 2024-07-17
Upgrade `download-artifact` action to v4

shibayan:
Since `download-artifact@v2` is no longer compatible with `upload-artifact@v4`, download-artifact will also be upgraded to v4.


chriskrycho (approved):
Excellent, thank you! Will merge once CI is ✅

PR #1812, opened 2024-07-30
Remove extra GET for tarball uncompressed size

### LINKED ISSUE #1744: When trying to install new versions volta hangs indefinitely and blocks all other network requests from resolving

charlespierce:
Fixes #1744 

Info
-----
Currently, we make two overlapping GET requests when fetching a tarball: First, the main request to get the headers and the data stream, and then a second request to attempt to fetch the ISIZE field which holds the uncompressed size. Getting the uncompressed size lets us use the stream data after uncompressing to populate our progress bar for tarballs.

However, the extra request causes issues for some users, where the second request for the ISIZE field hangs during the TLS handshake. By removing the uncompressed size entirely and connecting the progress bar to the compressed read data, we can avoid the extra request entirely.

Changes
-----
* Removed the `uncompressed_size` method from the `Archive` trait.
* Removed the extra request for the ISIZE field from the `Tarball` implementation of `Archive`
* Updated `Tarball` to emit progress off the compressed reader, rather than the output of the decoder (so that progress values will be based on the compressed size).

Tested
-----
* Ran locally and verified that fetching a Tarball still works as expected.
* Visual test didn't show any imperfections with the progress bar&mdash;it seemed to work the same as the previous approach using the uncompressed size. 

Notes
-----
* I didn't have a chance to test the progress bar on a Unix machine yet, I'd be interested to see if there is any noticeable delay or jitteriness from the change in how the progress bar is filled in those cases.

### DIFF
--- a/crates/archive/src/tarball.rs
+++ b/crates/archive/src/tarball.rs
@@ -2,25 +2,20 @@
 //! tarball in Unix operating systems.
 
 use std::fs::File;
-use std::io::{Read, Seek, SeekFrom};
+use std::io::Read;
 use std::path::Path;
 
 use super::{Archive, ArchiveError, Origin};
 use attohttpc::header::HeaderMap;
 use flate2::read::GzDecoder;
 use fs_utils::ensure_containing_dir_exists;
-use headers::{AcceptRanges, ContentLength, Header, HeaderMapExt, Range};
+use headers::{ContentLength, Header, HeaderMapExt};
 use progress_read::ProgressRead;
 use tee::TeeReader;
 
 /// A Node installation tarball.
 pub struct Tarball {
     compressed_size: u64,
-    // Some servers don't return the right data for byte range queries, so
-    // getting the uncompressed archive size for tarballs is an Option.
-    // If the uncompressed size is not available, the compressed size will be
-    // used for the download/unpack progress indicator, so that will be slightly off.
-    uncompressed_size: Option<u64>,
     data: Box<dyn Read>,
     origin: Origin,
 }
@@ -36,11 +31,9 @@ fn content_length(headers: &HeaderMap) -> Result<u64, ArchiveError> {
 
 impl Tarball {
     /// Loads a tarball from the specified file.
-    pub fn load(mut source: File) -> Result<Box<dyn Archive>, ArchiveError> {
-        let uncompressed_size = load_uncompressed_size(&mut source);
+    pub fn load(source: File) -> Result<Box<dyn Archive>, ArchiveError> {
         let compressed_size = source.metadata()?.len();
         Ok(Box::new(Tarball {
-            uncompressed_size,
             compressed_size,
             data: Box::new(source),
             origin: Origin::Local,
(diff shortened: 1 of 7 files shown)


crates/archive/src/tarball.rs
```diff
@@ -81,16 +68,13 @@ impl Archive for Tarball {
...
-    }
     fn unpack(
         self: Box<Self>,
         dest: &Path,
         progress: &mut dyn FnMut(&(), usize),
     ) -> Result<(), ArchiveError> {
-        let decoded = GzDecoder::new(self.data);
-        let mut tarball = tar::Archive::new(ProgressRead::new(decoded, (), progress));
+        let decoded = GzDecoder::new(ProgressRead::new(self.data, (), progress));
+        let mut tarball = tar::Archive::new(decoded);
```
chriskrycho, line 77: Just for my own clarity: what motivated changing up where we put the `ProgressRead`? It seems like it *should* be the same behavior either way, since ultimately in both cases it’s just wrapping the `Read` stream?

chriskrycho (approved):
This looks eminently reasonable, and local testing on my Mac seems to be working well with no visual hiccups.

(Intuitively, this makes a lot of sense to me: decoding the file should more or less always be dwarfed by the download time, and the download time is going to be a function of the compressed, not the uncompressed, size.)

I had one question, but it’s absolutely not blocking. :shipit:

crates/archive/src/tarball.rs
charlespierce: Yeah, I can definitely explain! For some background:

1. `ProgressRead` is a thin wrapper around a reader that calls the provided `FnMut` every time data is read, passing in the amount of data that was read from the wrapped reader.
2. `GzDecoder` is a `Read` wrapper around a stream of _compressed_ bytes, which produces a stream of _uncompressed_ bytes.

Given that, the motivation behind the change is that previously we were wrapping the `decoded` reader with `ProgressRead`, meaning the number of bytes it would report would be from the stream of _uncompressed_ bytes—they would add up to the total uncompressed size. Since we're no longer fetching / relying on the uncompressed size, we instead move the `ProgressRead` to wrap the stream of _compressed_ bytes, which allows the total amount of data recorded to match the compressed size.

PR #1815, opened 2024-07-30
infra: enable tracing-level logging

chriskrycho:
The previous work on this in #1793 made it *possible* to use more log levels; this actually *uses* that new possibility. There is only a single tracing message at present, used to indicate that tracing itself is turned on, but we can use it for any future debugging work. Extracted from #1784, where we used it to help diagnose the issues underlying #1744.

### DIFF
--- a/crates/archive/Cargo.toml
+++ b/crates/archive/Cargo.toml
@@ -19,3 +19,4 @@ cfg-if = "1.0"
 headers = "0.4"
 thiserror = "1.0.16"
 attohttpc = { version = "0.28", default-features = false, features = ["json", "compress", "tls-rustls-native-roots"] }
+log = { version = "0.4", features = ["std"] }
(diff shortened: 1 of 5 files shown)


crates/archive/Cargo.toml
```diff
@@ -19,3 +19,4 @@ cfg-if = "1.0"
 headers = "0.4"
 thiserror = "1.0.16"
 attohttpc = { version = "0.28", default-features = false, features = ["json", "compress", "tls-rustls-native-roots"] }
+log = { version = "0.4", features = ["std"] }
```
chriskrycho, line 22: I included this here originally as part of the debugging work in #1784, but it seems reasonable to just have it here regardless. It would also be good to update these kinds of dependencies to be `log.workspace = true` instead of specifying their own dependencies, but that’s for a different change!

charlespierce: Seems reasonable! I agree it's worth updating our use of Cargo workspaces—I think this repo pre-dates a lot of the improvements in that space that Cargo has shipped—but also agree it's for a separate PR and not a blocker for anything.

charlespierce (approved):
LGTM!

PR #1829, opened 2024-08-10
Fix two typo

karthest:
Hi I'm using volta a lot and reading its source code and try to contribute . I think the code here is to init Project from a certain directory ( by resolve closest package.json ), so here should be from rather than for right ?


chriskrycho (commented):
Thanks for the suggested change! In this case, `for_*` is actually preferable. Either could work, but “for” rather than “from” is really more accurate for what we’re trying to say in normal English idioms: the project *for* that directory, rather than the project *from* that directory. Closing this accordingly!

karthest:
Thanks for your kind response

PR #1833, opened 2024-08-12
Update 'zip' to the latest version

charlespierce:
Info
-----
* With the bump to our MSRV, we can now take advantage of the latest version of `zip`
* Additionally, we can take advantage of `ZipStreamReader`, which is an implementation of a streamed Zip extractor. It's marked as `unstable` so the crate authors recommend that we pin to a specific `zip` version when using it.
* The `ZipStreamReader` lets us greatly simplify the code for fetching & extracting a Zip file, making it much more analogous to the `tarball` code.

Changes
-----
* Updated the version of `zip`, keeping the same features enabled.
* Altered the implementation of `Zip` to take advantage of `ZipStreamReader`

Tested
-----
* Ran Volta locally on Windows and verified that fetching a Node version still works as expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,6 +81,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arbitrary"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


chriskrycho (approved):
🤩👏🏼

PR #1857, opened 2024-08-16
Fix argument parsing for `volta run` to properly handle flags

### LINKED ISSUE #1852: Volta doesn't pass certain arguments to the globally installed npm package
## Reproduction Steps
1. Create a simple npm package.

    <p><details open>
    <summary><code>package.json</code></summary>

    ```json
    {
      "name": "my-pkg",
      "version": "0.0.1",
      "bin": {
        "my-cmd": "./index.mjs"
      },
      "main": "./index.mjs"
    }
    ```

    </details></p>

    <p><details open>
    <summary><code>index.mjs</code></summary>

    ```js
    #!/usr/bin/env node
    console.log(process.argv);
    ```

    </details></p>

2. Install this package globally with `npm`.

    ```sh
    npm i -g ./my-pkg/
    ```

3. Run `my-cmd` with `-h` or `--help` arguments (e.g. `my-cmd -h`).

    ⇒ Then, Volta outputs the `volta run` help message.

    ```console
    $ my-cmd -h
    Run a command with custom Node, npm, pnpm, and/or Yarn versions
    
    Usage: volta.exe run [OPTIONS] <COMMAND> [ARGS]...
    
    Arguments:
      <COMMAND>  The command to run
      [ARGS]...  Arguments to pass to the command
    
    Options:
          --node <version>    Set the custom Node version
          --npm <version>     Set the custom npm version
          --bundled-npm       Forces npm to be the version bundled with Node
          --pnpm <version>    Set the custon pnpm version
          --no-pnpm           Disables pnpm
          --yarn <version>    Set the custom Yarn version
          --no-yarn           Disables Yarn
          --env <NAME=value>  Set an environment variable (can be used multiple times)
          --verbose           Enables verbose diagnostics
          --very-verbose      Enables trace-level diagnostics
          --quiet             Prevents unnecessary output
      -h, --help              Print help
    ```

    Note that I have confirmed that the behavior is as expected when `my-cmd` is run without any arguments.

## Consideration
The following script exists in `${LOCALAPPDATA}/Volta/bin/my-cmd`.
```bash
#!/bin/bash
volta run "$(basename $0)" "$@"
```
This is my guess, but I suspect that the `"$@"` arguments that should be passed to the package are being interpreted as `volta run` arguments.

In fact, after modifying the script as follows (simply adding `--`), this issue no longer occurs.

```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-volta run "$(basename $0)" "$@"
+volta run -- "$(basename $0)" "$@"
```
I am not sure if this workaround is universally correct, so I would appreciate your consideration of a safe fix for this issue.

charlespierce:
Fixes #1852 

Info
-----
Clap's parsing for variable-length argument lists has precedence rules for what happens when a potential vararg overlaps with a known flag value (see https://docs.rs/clap/latest/clap/struct.Arg.html#method.allow_hyphen_values):

1. If we are already in a list of variable args, the args take precedence and any potential flags are ignored (flags are treated as part of the vararg)
2. If we are _not_ already in a list of variable args, then known flags take precedence and are parsed as such, rather than starting the vararg list

As a result of this precedence, with `volta run` set up to have `args` be separate from `command`, we could run into parsing issues with commands that have flags that overlap with Volta's. For example:

```
volta run node -h
```

When parsing this, Clap would see `node` as the `command`, and then it would parse `-h` as _Volta's_ `-h` (help) flag and print Volta's help. This is because the variable-length `args` hadn't started yet (we hadn't run into any unknown positional arguments), so Clap treated `-h` as a known flag with higher precedence. This is exactly the opposite of what we actually want for `volta run`

Instead, what we want is that everything after the `command` argument should be treated as part of the arguments to the command; regardless of whether the flag overlaps with Volta's flags.

To achieve that, we combine `command` and `args` into a _single_ vararg, so that by virtue of having a command in the first place, the vararg has started parsing and Clap will treat any flags as more argument values, rather than as flags.

Changes
-----
* Updated the command definition of `Run` to have `command_and_args` as a single list instead of a value and a list.
* Pulled the command and args out inside of `Command::run` so that they can still be passed to `execute_tool` in the same manner as before

Tested
-----
* Ran `volta run node -h` and verified that it correctly runs `node -h`, rather than showing the help output for `volta run`
* Ran `volta run node` to make sure the arg processing is handled correctly when there are no additional arguments.
* Ran `volta run` to validate that Clap shows a "missing required argument" message rather than creating a `Run` instance with an empty `command_and_args` vector.

Note
-----
* AFAICT, this parsing idiosyncrasy has existed since we implemented `volta run`, however it is a higher priority now because we switched the 3rd-party shims on Windows to use `volta run` internally in #1755

### DIFF
--- a/src/command/run.rs
+++ b/src/command/run.rs
@@ -44,12 +44,14 @@ pub(crate) struct Run {
     #[arg(long = "env", value_name = "NAME=value", num_args = 1)]
     envs: Vec<String>,
 
-    /// The command to run
-    command: OsString,
-
-    /// Arguments to pass to the command
-    #[arg(allow_hyphen_values = true, trailing_var_arg = true)]
-    args: Vec<OsString>,
+    /// The command to run, along with any arguments
+    #[arg(
+        allow_hyphen_values = true,
+        trailing_var_arg = true,
+        value_name = "COMMAND",
+        required = true
+    )]
+    command_and_args: Vec<OsString>,
 }
 
 impl Command for Run {
@@ -59,7 +61,13 @@ impl Command for Run {
         let envs = self.parse_envs();
         let platform = self.parse_platform(session)?;
 
-        match execute_tool(&self.command, &self.args, &envs, platform, session).into_result() {
+        // Safety: At least one value is required for `command_and_args`, so there must be at
+        // least one value in the list. If no value is provided, Clap will show a "required
+        // argument missing" message and this function won't be called.
+        let command = &self.command_and_args[0];
+        let args = &self.command_and_args[1..];
+
+        match execute_tool(command, args, &envs, platform, session).into_result() {
             Ok(()) => {
                 session.add_event_end(ActivityKind::Run, ExitCode::Success);
                 Ok(ExitCode::Success)


rwjblue (approved):
Good sleuthing!!

PR #1861, opened 2024-08-19
fix: update checksum for unix install script

jsejcksn:
It looks like the checksum for the unix install script hasn't been updated along with updates to the script itself. This PR updates the contents of the checksum file to align it with the current install script content.

Before this PR:

```
./dev/unix % sha256sum -c SHASUMS256.txt
volta-install.sh: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
```

After:

```
./dev/unix % sha256sum -c SHASUMS256.txt
volta-install.sh: OK
```

---

This can be automated in the future (e.g. using a GitHub Action) — whether that's actually generating the checksum, or simply verifying correctness prior to a merge.


chriskrycho (approved):
Ooof! Thanks for catching this! I’ll have @charlespierce do a double-check as well before merging, and we’ll definitely want it to automate it going forward.

jsejcksn:
> we’ll definitely want it to automate it going forward

@chriskrycho I'm not familiar with the build pipeline for this repo or I would have included that change in this PR. I suggest also adding an integration test as part of the deployment process for the Volta website: so that the install script on the site is checked against the hash in this repo.

charlespierce (approved):
Verified the hash locally, LGTM! Thanks for catching this.

Agreed we should automate, I think validating is probably an easier action to write than generating, if I have some time this weekend I'll take a look at adding a check for that.

PR #1877, opened 2024-08-27
chore: use httpdate to format expires

tottoto:
Uses `httpdate` crate to format expires. This simplify the implementation a little, and uses the same `httpdate` crate directly for encoding and decoding not depending on the specification or the internal implementation.


chriskrycho (approved):
This is a nice win! Thanks!

tghvsh:
[](url)

Botyy1 (approved):

PR #1878, opened 2024-08-29
[IMP] Read the download URL from environment variables.

MrTian257:
在特定环境下会无法访问node.org等网络的，可以从环境变量中获取新的node下载地址，或加速

--
In certain environments where access to node.org and similar networks is restricted, you can obtain a new Node download URL from environment variables or use a mirror for faster downloads.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,8 @@
+{
+  "name": "volta",
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "features": {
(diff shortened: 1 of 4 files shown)


chriskrycho (commented):
We already have a system in place for specifying custom mirrors: [hooks](https://docs.volta.sh/advanced/hooks)—and we definitely don't want to hard code other URLs. If we wanted to support using environment variables, that would need [an RFC](https://github.com/volta-cli/rfcs). Closing this accordingly. Thanks.

PR #1977, opened 2024-12-21
chore: remove legacy numeric constant import

tottoto:
Removes the legacy numeric constant import.


chriskrycho (approved):
Whoa, that’s ancient!

PR #1981, opened 2024-12-24
chore: simplify conditional expression

tottoto:
Simplifies the conditional expressions.


charlespierce (approved):
LGTM! Thanks for catching those 🎉

PR #1994, opened 2025-01-11
chore: fix Cargo.lock file

tottoto:
Fixes the `Cargo.lock` file.

```
error: failed to parse lock file at: /Users/runner/work/volta/volta/Cargo.lock

Caused by:
  package `winreg` is specified twice in the lockfile
```


rwjblue (approved):
Strange how we got in that state (folks are typically not modifying the lockfile directly). 

Thanks for fixing!

PR #2027, opened 2025-03-27
Bump `zip` crate to `2.5.0`

MarkusPettersson98:
All versions of `zip < 2.3.0` has been yanked: https://github.com/zip-rs/zip2/issues/315. This PR bumps `zip` to the latest version as there seems to be no breaking changes made since `2.1.6` which is the current version that `volta` depend on.


Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
chriskrycho, line 3: This is why the build is failing. Best guess: you used Rust 1.85 while updating, rather than the Rust version from the Rust toolchain file. Which is odd: if you’re using rustup, that *should* have used the local version automatically? In any case, should be a quick fix.

Issue #2053, opened 2025-06-24
Calling `node` shouldn't add `yarn` to the PATH


arcanis:
Generally speaking, only tools that matter to the current process should be injected to the PATH. For example, running `node ...` should only add `.volta/tools/image/node/22.13.1/bin` to the PATH; `.volta/tools/image/yarn/1.22.19/bin` should only be added to the PATH once the `yarn` shim is called.

The reason for that is that until the Yarn shim is called, Volta doesn't know whether it *would ever be called*. On my system the `yarn` binary comes from another source, let's say `/path/to/my/yarn`, but when calling `execFile` from within Node the Volta Yarn is unexpectedly called because Volta force-injected the shim.

ZJRui:
No one's maintaining it, the project should be shut down.

PR #2057, opened 2025-07-07
Implements `VOLTA_FEATURE_YARN=0`

arcanis:
I'm implementing a version manager for Yarn, and Volta gets in the way: every call to `node` corrupts the PATH by shadowing the existing `yarn` binary.

In theory Volta should be smarter and only update the PATH for binaries it's sure to own (see https://github.com/volta-cli/volta/issues/2053), but in practice this is likely to be a significant change and I doubt we have the bandwidth to land this kind of change medium term.

Instead, this PR implements a `VOLTA_FEATURE_YARN=0` environment variable that, when set, will prevent Volta from shadowing the Yarn binary.  Any wrapper meant to own the management of Yarn releases can then set this variable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1657,7 +1657,7 @@ dependencies = [
  "volta-core",
  "volta-migrate",
  "which",
- "winreg",
+ "winreg 0.53.0",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


marion1407-art (approved):
B-2PC37711PS920213V

PR #2083, opened 2025-11-24
fix(install script)!: exit with non-zero exit code on error

oscar-vitecenova:
The install script should exit with a non-zero exit code when an error occurs during the installation. Otherwise users invoking the script can't handle the error on their end. Currently the script exits with zero exit code even when an error occurs.


Ascalad:
Pull request and add workflows to fix workflow non-implementation issues

